Issue #6, opened 2026-02-16
Only half of system RAM is detected


Kevazy:
First of all, nice work! Super clean interface.

I noticed that it detected "45.8 GB total" system RAM while the actual amount should be closer to 96 GB, as that's what's installed and shown in the Windows task manager. I ran the command in WSL. Everything else is shown and functioning correctly.

Any idea why?

AlexsJones:
@Kevazy thanks! I have done zero testing on WSL or Windows. So that it's functioning is pretty awesome - I will endevour to get a VM up to test this tomorrow and figure it out.

MedievalSp:
Also interested in getting it working in Windows! 🙏 this app is appealing to people with gaming rigs.

N4zroth:
Same for me, it also isn't detecting my GPU in WSL for some reason.

Kevazy:
This 45.8 GB total was a false alarm - it's apparently WSL's own memory allocation, so it has nothing to do with the code. Added a tag for this in PR #8. I got it running in Windows natively (by compiling with cargo) and everything is working perfectly on my end.

@N4zroth @MedievalSp I recommend creating specific issues regarding your requests - closing this here.

AlexsJones:
This has been merged

PR #8, opened 2026-02-16
TUI: append (WSL) tag to RAM when running inside WSL

Kevazy:
## Summary
  This PR adds a small clarification in the TUI system bar:
  - RAM text now appends ` (WSL)` when llmfit is running inside Windows Subsystem for Linux.

  Example:  
image

  ## Why
  I reported (#6) that RAM looked incorrect in WSL as it did not align with my installed size.
  I later found out that this is expected WSL behavior due to its set memory allocation, not a bug in llmfit’s RAM detection.
  This change makes that context explicit and avoids confusion for Windows+WSL users.

  ## Changes
  - Added internal WSL environment detection helper in `src/hardware.rs`.
  - Cached detection with `OnceLock` to avoid repeated checks in the render path.
  - Updated TUI RAM label formatting in `src/tui_ui.rs` to append ` (WSL)` when applicable.

  ## Validation
  - Verified behavior on native Windows (no suffix).
  - Verified behavior in WSL (suffix shown).

### DIFF
--- a/src/hardware.rs
+++ b/src/hardware.rs
@@ -359,6 +359,27 @@ impl SystemSpecs {
     }
 }
 
+pub(crate) fn is_running_in_wsl() -> bool {
+    static IS_WSL: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *IS_WSL.get_or_init(detect_running_in_wsl)
+}
+
+fn detect_running_in_wsl() -> bool {
+    if !cfg!(target_os = "linux") {
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


Kevazy:
Makes sense to show the true WSL memory as well (instead of some PowerShell bypass to read the actual system RAM) considering LLMs in WSL relies on that configuration. But - this makes it more important to provide a reliable Windows/PowerShell install script so that you don't have to go down the WSL path if not needed (like I did).

PR #26, opened 2026-02-20
feat: MLX-native inference support for Apple Silicon

counterposition:
## Background

llmfit already knows you're on an Apple Silicon Mac — it detects the Metal GPU backend and unified memory. But until now, it assumed you'd always run models through **Ollama/llama.cpp**, which uses a model format called GGUF.

On Apple Silicon, there's a second option: **MLX**, Apple's own machine-learning framework. MLX models run 20–80% faster than GGUF models on the same hardware because they're optimized for the Metal GPU and unified memory architecture at a lower level. They're also distributed in a different format (stored in the Hugging Face `mlx-community` namespace rather than as `.gguf` files).

This PR teaches llmfit about MLX so that, on Apple Silicon, it can:

1. **Estimate speed correctly** — previously, speed estimates for Apple Silicon used llama.cpp throughput numbers. Now the fit engine knows about MLX and picks the faster runtime when available.
2. **Recommend the right quantization format** — GGUF models use formats like `Q4_K_M`; MLX models use `mlx-4bit` / `mlx-8bit`. The quantization selection logic now handles both hierarchies.
3. **Detect MLX models already on disk** — by scanning the Hugging Face model cache for `mlx-community` models, no Python dependency required.
4. **Download MLX models** — via `huggingface-cli` (or the `huggingface_hub` Python package as a fallback).
5. **Surface all of this in the TUI and CLI** — the system bar shows MLX availability alongside Ollama, the detail view shows the recommended runtime with a speed comparison, and `--json` output includes a `runtime` field.

Non-Apple users are unaffected: the runtime defaults to `LlamaCpp` on CUDA/CPU systems, and no MLX UI elements appear.

## What changed

| Commit | What it does |
|--------|--------------|
| `b6d4f52` | Add `InferenceRuntime` enum and MLX quantization support in the fit engine |
| `f581f41` | Add `MlxProvider` for installed-model detection and downloads |
| `de76b5d` | Remove Python fallback for Hugging Face downloads (simplification) |
| `c003e57` | Surface runtime info in TUI, CLI, and JSON output |
| `bf360e4` | Version bump in Cargo.lock |

## Reviewer note

This is a fairly large PR. For the clearest reading, **please go commit-by-commit** — each commit is a self-contained increment and the ordering follows the natural dependency graph (fit engine → provider → display). Reading the diff as a whole makes the changes harder to follow than reading them in the order they were built.

## Testing

- `cargo test` — all existing and new tests pass
- On Apple Silicon: TUI shows MLX status in system bar, models show MLX runtime with speed comparison in detail view
- On non-Apple systems: behavior is identical to pre-MLX (no visible changes)
- `llmfit fit --json` — output includes `"runtime"` field
- `llmfit recommend --runtime mlx` — filters to MLX-recommended models

### DIFF
--- a/src/fit.rs
+++ b/src/fit.rs
@@ -63,11 +63,11 @@ pub struct ModelFit {
     pub moe_offloaded_gb: Option<f64>, // GB of inactive experts offloaded to RAM
     pub score: f64,                    // weighted composite score 0-100
     pub score_components: ScoreComponents,
-    pub estimated_tps: f64, // estimated tokens per second
-    pub best_quant: String, // best quantization for this hardware
-    pub use_case: UseCase,          // inferred use case category
-    pub runtime: InferenceRuntime,   // inference runtime (MLX or llama.cpp)
-    pub installed: bool,             // model found in a local runtime provider
+    pub estimated_tps: f64,        // estimated tokens per second
+    pub best_quant: String,        // best quantization for this hardware
+    pub use_case: UseCase,         // inferred use case category
+    pub runtime: InferenceRuntime, // inference runtime (MLX or llama.cpp)
+    pub installed: bool,           // model found in a local runtime provider
 }
 
 impl ModelFit {
@@ -207,8 +207,13 @@ impl ModelFit {
 
         // Add runtime comparison note on Apple Silicon
         if runtime == InferenceRuntime::Mlx {
-            let llamacpp_tps =
-                estimate_tps(model, &best_quant_str, system, run_mode, InferenceRuntime::LlamaCpp);
+            let llamacpp_tps = estimate_tps(
+                model,
+                &best_quant_str,
+                system,
+                run_mode,
+                InferenceRuntime::LlamaCpp,
+            );
             if llamacpp_tps > 0.1 {
                 let speedup = ((estimated_tps / llamacpp_tps - 1.0) * 100.0).round();
                 if speedup > 0.0 {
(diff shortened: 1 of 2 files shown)


counterposition:
This was originally submitted as a trio of stacked PRs, but stacked PRs don't work well with forked repos on GitHub, which is why I made this one PR instead.

counterposition:
> **One question**: The Python fallback for HF downloads was removed in `de76b5d` — is `huggingface-cli` reliably available on most macOS MLX setups? If someone has MLX via pip but not the CLI, would they hit a wall?

I think it's rather uncommon for `huggingface-cli` to be installed on most people's Macs unless they are ML researchers or already deep in the local models rabbit hole. While newer versions of macOS ship with Python 3 and pip by default, we shouldn't automatically install it for them, because everyone has a different way of managing Python versions libraries on their machine. Some people install everything globally (which doesn't scale), others use Conda, others use poetry, etc. I personally think [astral-sh/uv](https://docs.astral.sh/uv/) is far and away the most user-friendly and least brittle way to manage Python toolchains. But I don't want to foist it upon users, which is why I just left it as an actionable suggestion in the [error message](https://github.com/AlexsJones/llmfit/pull/26/changes#diff-9a81d2527f08eb7e941b2e38a1097c19a55f5bc3376f8c7ab2bb3e0334135d76R335).

There is a similar issue with `mlx_lm`; it's not installed by default but is essential for running MLX models locally. I was looking for a good place to surface a hint to the user. Since it's a UX concern, I'll leave this to @AlexsJones .

AlexsJones:
Please format

PR #49, opened 2026-02-22
fix: use quantization in path selection for both dense and MoE models

### LINKED ISSUE #50: Fix: path selection uses default quantization before checking if lower quants fit
## Problem

The current code makes path selection decisions (GPU vs CPU offload vs MoE offload) based on the model's default quantization or static `min_vram` values, before checking what quantizations could actually fit in the available memory.

This leads to situations where:
- A dense model with high default quant (e.g., F16) might be marked as not fitting in VRAM, but a lower quant (e.g., Q4_K_M) would fit fine
- A MoE model with high default quant might not fit for MoE offload, but a lower quant would work

## Solution

This is fixed by PR #49, which:
1. For dense models: Uses `choose_quant()` before deciding GPU path
2. For MoE models: Tries the full quantization hierarchy in `moe_offload_path()` before falling back to CPU
3. Adds comprehensive test coverage

## Related
PR: #49

luojiyin1987:
This PR fixes the logic where path selection was done before checking what quantizations could actually fit.

## Changes:

1. **Dense models**: Use `choose_quant()` before deciding GPU path instead of just checking `min_vram`
2. **MoE models**: Try quantization hierarchy in `moe_offload_path()` before falling back to CPU
3. Add `moe_memory_for_quant()` helper to compute MoE memory at specific quantization levels
4. Add comprehensive test coverage

This fixes the issue where models that could run with lower quantization were incorrectly marked as non-runnable.

## Test coverage:
- All existing tests pass
- New test for dense model quantization in path selection
- New test for MoE model quantization in path selection

Close #50

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -108,6 +108,17 @@ impl ModelFit {
 
         let min_vram = model.min_vram_gb.unwrap_or(model.min_ram_gb);
         let use_case = UseCase::from_model(model);
+        let default_mem_required =
+            model.estimate_memory_gb(model.quantization.as_str(), model.context_length);
+
+        // Determine inference runtime up front so path selection can use
+        // the correct quantization hierarchy.
+        let runtime = if system.backend == GpuBackend::Metal && system.unified_memory {
+            InferenceRuntime::Mlx
+        } else {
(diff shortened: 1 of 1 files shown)


AlexsJones:
@three-foxes-in-a-trenchcoat please review

PR #53, opened 2026-02-23
ci: enable windows build targets

akarsh16reddy:
Adds Windows (x86_64 and aarch64) as release targets

- Produce .zip for Windows targets, .tar.gz for Linux/macOS
- Use RUNNER_TEMP instead of /tmp for Windows runner compatibility
- Update release glob to include *.zip artifacts

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


akarsh16reddy:
Hey @AlexsJones - adding Windows binaries to the release pipeline so users can grab a pre-built executable directly rather than needing Cargo installed. Covers both x86_64 and aarch64. Happy to adjust if this isn't a direction you want to take the project!

Release test: https://github.com/akarsh16reddy/llmfit/releases
Actions test: https://github.com/akarsh16reddy/llmfit/actions

PR #60, opened 2026-02-23
fix: add --local flag to install script and improve fallback logic

### LINKED ISSUE #56: Does it really need to be installed system-wide?
```
$ curl -fsSL https://llmfit.axjns.dev/install.sh | sh
  > llmfit installer
  > Downloading llmfit v0.3.8 for x86_64-unknown-linux-musl...
  > Extracting...
  > Installing to /usr/local/bin (requires sudo)...
[sudo] password for user:

sudo: a password is required
```

Does it really need to be installed system-wide without using packages?

three-foxes-in-a-trenchcoat:
## Summary

Fixes #56 — the install script now supports local installation without requiring sudo.

## Changes

1. **Add `--local` / `-l` flag** — forces installation to `~/.local/bin`
2. **Check for interactive terminal** — only attempt sudo if stdin is a TTY (`[ -t 0 ]`)
3. **Graceful fallback** — if sudo fails, automatically fall back to `~/.local/bin`
4. **Better PATH guidance** — shows copy-paste command when install dir isn't in PATH

## Usage

```bash
# Default behavior (tries /usr/local/bin with sudo, falls back to ~/.local/bin)
curl -fsSL https://llmfit.axjns.dev/install.sh | sh

# Force local install (no sudo prompt)
curl -fsSL https://llmfit.axjns.dev/install.sh | sh -s -- --local
```

## Testing

- Verified argument parsing works with `sh -s -- --local`
- Verified TTY check prevents sudo prompt in non-interactive environments

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,13 @@ brew install llmfit
 curl -fsSL https://llmfit.axjns.dev/install.sh | sh
 ```
 
-Downloads the latest release binary from GitHub and installs it to `/usr/local/bin` (or `~/.local/bin`).
+Downloads the latest release binary from GitHub and installs it to `/usr/local/bin` (or `~/.local/bin` if no sudo).
+
+**Install to `~/.local/bin` without sudo:**
+
+```sh
+curl -fsSL https://llmfit.axjns.dev/install.sh | sh -s -- --local
+```
 
 ### From source
 


akarsh16reddy:
It would also help to update the README to reflect the support for --local flag

three-foxes-in-a-trenchcoat:
Good call! Just pushed an update to document the `--local` flag in the README. 🦊

PR #67, opened 2026-02-23
Support light console themes

AndiDog:
I did simple replacements to accomplish this:

- Text color: `Color::White` removed to use the theme default (as is the case for background color)
- Table `row_highlight_style` changed to `Color::LightBlue` which normally adapts to theme
- Cleanup: `add_modifier(Modifier::BOLD)` → `bold()`

### DIFF
--- a/llmfit-tui/src/tui_ui.rs
+++ b/llmfit-tui/src/tui_ui.rs
@@ -1,7 +1,7 @@
 use ratatui::{
     Frame,
     layout::{Constraint, Direction, Layout, Rect},
-    style::{Color, Modifier, Style},
+    style::{Color, Style},
     text::{Line, Span},
     widgets::{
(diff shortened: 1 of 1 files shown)


AlexsJones:
I like the idea, let's build this out to allow people to have a theme picker. What do you think?

AndiDog:
I think TUIs should automatically work with the default ANSI color palette and not unnecessarily add theming support and such. By not using `White` or `Black` explicitly, this is easily done. If I were the author of this project, I wouldn't like the code to be blown up to much more than _showing LLM fit_. Anyway, please adapt the PR to however you see fit 😉. I don't have a strong opinion.

AlexsJones:
Both approaches can exist, and I agree that default ANSI colours are the way to go.

PR #70, opened 2026-02-23
fix: use aggregated VRAM for multi-GPU fit scoring

### LINKED ISSUE #68: 2 GPUs detected but only 1 factored in calculation
Hi - Super cool project. It looks like both my 3090s are detected, but when looking at the memory utilization only VRAM from one card is used. This should be based on 48gb right?

Image

three-foxes-in-a-trenchcoat:
## Summary

Fixes #68 — systems with multiple identical GPUs (e.g., 2x RTX 3090) now correctly use total VRAM for model fitting.

## Problem

Previously, llmfit only used per-card VRAM from the primary GPU:
- User has: 2x RTX 3090 (24GB each = **48GB total**)
- llmfit used: **24GB** for fit scoring
- Result: Models that could run on 48GB were marked as "too tight"

## Solution

1. Add `total_gpu_vram_gb` field to `SystemSpecs` that calculates `per_card_vram * count`
2. Use `total_gpu_vram_gb` for fit scoring (non-unified-memory systems)
3. Update display to show total VRAM: `RTX 3090 x2 (24 GB each = 48 GB total)`

## Why This Is Correct

Multi-GPU tensor splitting is supported by:
- **llama.cpp**: `--tensor-split` flag
- **vLLM**: Native multi-GPU support
- **Ollama**: Uses llama.cpp under the hood

## Testing

- `cargo check` ✅
- All 59 tests pass ✅

## Before/After

| System | Before | After |
|--------|--------|-------|
| 2x RTX 3090 | "24 GB" | "24 GB each = 48 GB total" |
| Model fit scoring | Uses 24GB | Uses 48GB |

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -88,9 +88,7 @@ impl SystemSpecs {
         let has_gpu = !gpus.is_empty();
         let gpu_vram_gb = primary.and_then(|g| g.vram_gb);
         // Total VRAM = per-card VRAM * count (for multi-GPU tensor splitting)
-        let total_gpu_vram_gb = primary.and_then(|g| {
-            g.vram_gb.map(|vram| vram * g.count as f64)
-        });
+        let total_gpu_vram_gb = primary.and_then(|g| g.vram_gb.map(|vram| vram * g.count as f64));
         let gpu_name = primary.map(|g| g.name.clone());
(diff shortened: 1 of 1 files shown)


AlexsJones:
@three-foxes-in-a-trenchcoat please cargo fmt

three-foxes-in-a-trenchcoat:
Done! cargo fmt applied. 🦊

PR #85, opened 2026-02-25
Add half‑page scrolling support

### LINKED ISSUE #84: Add half‑page scrolling support
Can we have support for `Ctrl+U` and  `Ctrl+D` support which does half-pages (maybe 5 rows?)

I found it useful to check what models are less "perfect" just for exploratory reasons.

I am happy to work on this (I already have something working) shouldn't be a lot at all after looking at the codebase.

iZarrios:
Closes #84

### DIFF
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -266,6 +266,16 @@ impl App {
         }
     }
 
+    pub fn half_page_up(&mut self) {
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat (approved):
Clean implementation! Bounds checking looks correct, consistent with the existing page_up/page_down pattern. LGTM 🦊

PR #97, opened 2026-02-26
feat: add 18 new models to database (Feb 2026)

lichengzhe:
## Summary

Add 18 recently released LLM models to the scraper's curated list, along with updated MoE configurations and active parameter mappings.

### Flagship / High Priority
| Model | Parameters | Type | Context | Release |
|-------|-----------|------|---------|---------|
| DeepSeek-V3.2 | 685B (37B active) | MoE | 128k | Dec 2025 |
| DeepSeek-V3.2-Speciale | 685B (37B active) | MoE | 128k | Dec 2025 |
| GLM-5 (zai-org) | 744B (40B active) | MoE | 200k | Feb 2026 |
| Kimi-K2.5 | 171B | Multimodal | 256k | Jan 2026 |
| MiniMax-M2.5 | 230B (10B active) | MoE | 128k | Feb 2026 |
| MiMo-V2-Flash | 309B (15B active) | MoE | 128k | Dec 2025 |

### Mid-tier / Specialized
| Model | Parameters | Type | Context | Release |
|-------|-----------|------|---------|---------|
| Nemotron-3-Nano-30B-A3B | 30B (3B active) | Mamba2+MoE | 1M | 2025 |
| Nemotron-Nano-9B-v2 | 9B | Mamba2 hybrid | 128k | 2025 |
| Phi-4-reasoning | 14B | Dense | 32k | Apr 2025 |
| Phi-4-mini-reasoning | 3.8B | Dense | 16k | Apr 2025 |
| Phi-4-multimodal-instruct | 14B | Multimodal | 128k | Apr 2025 |
| EXAONE-4.0-32B | 32B | Hybrid attn | 128k | Jul 2025 |
| EXAONE-4.0-1.2B | 1.2B | Dense | 32k | Jul 2025 |

### Lightweight / Edge
| Model | Parameters | Type | Context | Release |
|-------|-----------|------|---------|---------|
| SmolLM3-3B | 3B | Dense | 128k | Jul 2025 |
| MiMo-7B-RL | 7B | Dense | 32k | May 2025 |
| Gemma-3n-E4B-it | 8B (eff. 4B) | MatFormer | 128k | Jun 2025 |
| Gemma-3n-E2B-it | 4B (eff. 2B) | MatFormer | 128k | Jun 2025 |

## Changes

- `scripts/scrape_hf_models.py`:
  - Added 18 new entries to `TARGET_MODELS`
[...]

PR #99, opened 2026-02-26
fix: improve scoring algorithm for MoE models, recency, and fit curve

lichengzhe:
## Summary

Three targeted improvements to the scoring algorithm that address known inaccuracies.

### 1. MoE models: use active parameters for quality scoring

**Problem:** `quality_score()` uses total parameter count (`params_b()`) to determine base quality. For MoE models like Qwen3-Coder-Next (80B total / 3B active), this gives an inflated score of 95 (the 40B+ tier), when inference quality is closer to a 3-7B dense model.

**Fix:** When `active_parameters` is available, use it instead of total params for the quality tier lookup.

### 2. Recency bonus

**Problem:** A 7B model from 2024 and a 7B model from 2026 get the same quality score, despite newer models being significantly better at the same parameter count (better training data, architecture improvements).

**Fix:** Add a small recency bonus using the existing `release_date` field: +3 for models < 3 months old, +1.5 for < 9 months.

### 3. Smooth fit_score curve

**Problem:** The fit score uses hard step thresholds — jumping from 100 to 70 at 80% utilization, and from 70 to 50 at 90%. A model using 79% memory scores 100, but 81% scores 70.

**Fix:** Replace with a Gaussian curve centered at 65% utilization (`width=0.25`), providing smooth transitions with no jarring jumps.

## Tests

All 60 existing tests pass. Three `fit_score` tests updated to match the new smooth curve behavior.

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -1,6 +1,22 @@
 use crate::hardware::{GpuBackend, SystemSpecs};
 use crate::models::{self, LlmModel, UseCase};
 
+/// Calculate how many months ago a date string (YYYY-MM-DD) is from today.
+fn months_ago(date_str: &str) -> Option<u32> {
+    let parts: Vec<&str> = date_str.split('-').collect();
+    if parts.len() < 2 { return None; }
+    let year: i32 = parts[0].parse().ok()?;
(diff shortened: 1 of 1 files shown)


lichengzhe:
Replacing with a clean PR that's independent from #97

PR #101, opened 2026-02-26
Add basic support for Ascend NPUs

rauletorresc:
**Description of the feature**

We have added basic support for Ascend NPUs by querying via `npu-smi` the number of devices and their individual memory.

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -900,7 +900,7 @@ impl SystemSpecs {
             .filter_map(|line| line.split(':').last())
             .map(|s| s.trim().to_string())
             .collect();
-        
+
         if ids.is_empty() {
             return Vec::new();
         }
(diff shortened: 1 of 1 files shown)


rauletorresc:
> Good idea adding Ascend NPU support — nice to see niche hardware covered. However, there's a critical issue that needs fixing before merge:
> 
> ### 🔴 Blocker: `expect()` will crash llmfit on systems without `npu-smi`
> ```rust
> let list_output = std::process::Command::new("npu-smi")
>     .args(["info", "-l"])
>     .output()
>     .expect("Failed to run npu-smi");  // ← panics on 99.9% of systems
> ```
> 
> This needs to be `.ok()?` like every other detection function in `hardware.rs`. For reference, here's how NVIDIA detection handles it:
> 
> ```rust
> let output = match std::process::Command::new("nvidia-smi")
>     .arg("--query-gpu=memory.total,name")
>     .arg("--format=csv,noheader,nounits")
>     .output()
> {
>     Ok(o) if o.status.success() => o,
>     _ => return Vec::new(),
> };
> ```
> 
> ### Other issues:
> 1. **Returns `Some` even with 0 NPUs** — should return `None` when `ids.is_empty()`:
>    ```rust
>    if ids.is_empty() {
>        return None;
>    }
>    ```
> 2. **Minor formatting:** missing space in `detect_ascend_npus(){` → `detect_ascend_npus() {`
> 3. **Bandwidth estimate of 70.0** — any justification for this number? Ascend 910B has ~400 GB/s HBM bandwidth which is closer to CUDA-class. 70.0 seems conservative.
> 
> Happy to help iterate on this — the core detection logic is sound, just needs the safety fixes. 🦊

- I have addressed all issues.
- I have adjusted the memory bandwidth to a more realistic value.
- I have added each NPU as an individual device.

rauletorresc:
> Great work addressing all the feedback! 🦊
> 
> All my previous concerns have been resolved:
> 
> ✅ **Error handling fixed** - Uses `match` pattern instead of `expect()`, won't crash on systems without npu-smi ✅ **Empty check** - Returns `Vec::new()` when no NPUs found ✅ **Individual NPUs** - Creates separate `GpuInfo` for each device ✅ **Bandwidth updated** - 390.0 GB/s is realistic for Ascend 910B (~400 GB/s actual) ✅ **Proper spacing** - Function declaration formatted correctly
> 
> ## ⚠️ One blocker remaining:
> **Rustfmt is failing** - you need to run:
> 
> ```shell
> cargo fmt
> ```
> 
> Then commit and push the formatting changes. This is just automatic formatting, nothing to fix manually.
> 
> Once CI is green, this is ready to merge! The core logic is solid. 👍

Done

PR #110, opened 2026-02-27
security: add SHA256 checksum verification to install.sh

Lalatenduswain:
## Summary

- Generate a per-asset `.sha256` file in the release workflow alongside each `.tar.gz`/`.zip`
- Publish `.sha256` files as GitHub Release assets
- Verify the checksum in `install.sh` before extracting, aborting on mismatch
- Supports `sha256sum` (Linux) and `shasum` (macOS); gracefully skips for older releases that predate this change

## Test plan

- [ ] Trigger a release build and confirm `.sha256` files appear in the GitHub Release assets
- [ ] Run `install.sh` on Linux and verify "Verifying checksum..." step passes
- [ ] Run `install.sh` on macOS and verify the `shasum` fallback works
- [ ] Corrupt the downloaded asset manually and confirm the installer aborts with the expected error

### DIFF
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -209,6 +209,12 @@ impl LlmModel {
         Some((size_gb * 1.1).max(0.5))
     }
 
+    /// Returns true if this model is MLX-specific (Apple Silicon only).
+    /// MLX models are identified by having "-MLX" in their name.
+    pub fn is_mlx_only(&self) -> bool {
+        self.name.to_uppercase().contains("-MLX")
+    }
+
     /// For MoE models, compute RAM needed for offloaded (inactive) experts.
     /// Returns None for dense models.
     pub fn moe_offloaded_ram_gb(&self) -> Option<f64> {
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -221,6 +221,13 @@ fn run_recommend(
         _ => true,
     });
 
+    // Hide MLX-only models on non-Apple Silicon systems
+    let is_apple_silicon =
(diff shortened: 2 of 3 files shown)


AlexsJones:
Please format and review advice from @three-foxes-in-a-trenchcoat

Lalatenduswain:
Thanks for the thorough review @three-foxes-in-a-trenchcoat! I've addressed all your feedback in the latest commit. Here's a summary of what was done:

**1. Checksum generation format**
Updated `release.yml` to use the `awk`-based approach you suggested, ensuring a consistent output format regardless of whether `sha256sum` or `shasum` is used.

**2. `--locked` flag on `cross` install**
Removed `--locked` — you're right that it's redundant when installing a specific pinned version from crates.io. Changed to:
```bash
cargo install cross --version 0.2.5
```

**3. URL scheme validation logging**
Updated both `OLLAMA_HOST` and `MLX_LM_HOST` validation in `providers.rs` to use the `.ok().and_then(...)` pattern you suggested, so the invalid URL value is now included in the warning message for easier debugging.

**4. Checksum download timeout**
Added `--max-time 10` to the `curl` call in `verify_checksum()` in `install.sh`.

**Regarding the Tauri version bump (0.3.8 → 0.4.8):** That was part of a broader dependency update in the same branch and wasn't intentionally bundled here — noted for future PRs, I'll keep unrelated changes in separate commits.

Thanks again for the detailed review!

Lalatenduswain:
@AlexsJones could you please approve the CI run on the latest commit? The workflow is currently pending manual approval. All reviewer feedback from @three-foxes-in-a-trenchcoat has been addressed and the formatting is fixed.

Issue #127, opened 2026-03-01
Security: Path Traversal in GGUF Download Destination Handling


maciejmaczko:
### Summary
`llmfit-core` currently accepts a GGUF `filename` and uses it directly to build the destination path during download.

Because `filename` is joined into the cache directory without path-component validation, values containing traversal sequences (for example `../`) can escape the intended model cache directory and write files elsewhere on disk.

### Impact
An attacker who can influence the `filename` passed to GGUF download logic may cause arbitrary file writes relative to current user permissions.

Potential impact includes:
- Overwriting files outside `~/.cache/llmfit/models`
- Planting files in unexpected locations
- Corrupting user data/config under the same OS account

### Affected Code Paths
Main sink:
- `llmfit-core/src/providers.rs` in `LlamaCppProvider::download_gguf()`
  - `let dest_path = models_dir.join(filename);`
  - followed by `File::create(...)` and `rename(...)`

Input flow:
- `filename` can come from:
  - `list_repo_gguf_files()` (remote HF tree API output)
  - direct API usage of `download_gguf(repo_id, filename)` by downstream callers

### Root Cause
No canonical path safety checks are applied before writing:
- no rejection of absolute paths
- no rejection of `..` path traversal components
- no strict validation that destination remains within model cache root
- no early API-level rejection in `download_gguf()` before network/download logic starts

#### PoC A: Path Resolution Demonstration
```python
from pathlib import Path
models = Path('/Users/alice/.cache/llmfit/models')
print((models / '../../outside.gguf').resolve(strict=False))
# -> /Users/alice/.cache/outside.gguf
```
This demonstrates that joining `../../outside.gguf` escapes the intended cache directory.

#### PoC B: API-Level Trigger
```rust
use llmfit_core::providers::LlamaCppProvider;

fn main() {
    let provider = LlamaCppProvider::new();
    // Traversal payload in filename:
    let _ = provider.download_gguf("some/repo", "../../outside.gguf");
}
```
This value is accepted and used to construct the write path.

### Expected vs Actual Behavior
Expected:
- GGUF filename is a safe relative `.gguf` path inside the model cache root.
- Any traversal (`..`), absolute path, or invalid extension is rejected.

Actual:
- Unsafe path components are accepted.
- Download write target can escape the intended cache directory.

three-foxes-in-a-trenchcoat:
[...]

PR #128, opened 2026-03-01
Fix GGUF path traversal and add validation tests

maciejmaczko:
## Fix: Prevent GGUF Path Traversal in Download Handling

Closes: https://github.com/AlexsJones/llmfit/issues/127

### What This PR Changes
This PR hardens GGUF download path handling in `llmfit-core`.

1. Adds strict GGUF filename sanitization:
- rejects empty values
- rejects non-`.gguf` extension
- rejects absolute paths
- rejects non-normal path components (including parent traversal)

2. Enforces sanitization in `download_gguf()` before destination path construction.

3. Filters unsafe file entries in `list_repo_gguf_files()` so invalid remote paths are dropped early.

### Why This Fix Works
The vulnerability exists because untrusted path strings are joined directly into the cache directory.

The new sanitization gate ensures only safe relative `.gguf` paths are allowed. Any input containing traversal semantics is rejected before file creation or rename is attempted, preventing writes outside the intended cache root.

### Code References
- `llmfit-core/src/providers.rs`
  - sanitization helper added
  - `download_gguf()` now validates input before path join
  - `list_repo_gguf_files()` now ignores unsafe paths

### Tests Added
Unit tests added in `llmfit-core/src/providers.rs`:
- `test_sanitize_gguf_relative_path_accepts_regular_filename`
- `test_sanitize_gguf_relative_path_accepts_nested_relative_path`
- `test_sanitize_gguf_relative_path_rejects_parent_dir_traversal`
- `test_sanitize_gguf_relative_path_rejects_absolute_path`
- `test_sanitize_gguf_relative_path_rejects_non_gguf_extension`
- `test_download_gguf_rejects_traversal_before_network`

### Validation
- `cargo test -p llmfit-core` passes
- full workspace `cargo test` passes

### Notes
This PR keeps support for nested relative subpaths (for example `org/model.gguf`) while blocking traversal and absolute paths.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -563,7 +563,7 @@ impl LlamaCppProvider {
             .into_iter()
             .filter_map(|e| {
                 let path = e.get("path")?.as_str()?.to_string();
-                if !path.ends_with(".gguf") {
+                if sanitize_gguf_relative_path(&path).is_err() {
                     return None;
                 }
                 let size = e.get("size").and_then(|v| v.as_u64()).unwrap_or(0);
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat:
Hey @maciejmaczko — thanks for both the excellent issue report (#127) and this fix! Your write-up was thorough and made the vulnerability very clear.

We ended up merging a parallel fix in #129 that covers the same ground — input validation (empty, absolute, traversal, extension) plus a canonicalization safety check as a second layer.

I do like that your PR also filters unsafe entries in `list_repo_gguf_files()`, which is a nice defense-in-depth touch we didn't include. If you'd like to open a smaller follow-up PR adding just that part, we'd happily merge it.

Thanks again for the responsible disclosure and the quick turnaround! 🦊

maciejmaczko:
@three-foxes-in-a-trenchcoat Thanks! I set up [new PR](https://github.com/AlexsJones/llmfit/pull/144)

let me know if you need further amendments.

PR #130, opened 2026-03-01
Add new Qwen3.5 models to database

SirLorrence:
### All Multimodal vision and text models:
- Qwen3.5-27B (27.8B params, 256k context)
- Qwen3.5-35B-A3B (36.0B MoE, 256k context)
- Qwen3.5-122B-A10B (125.1B MoE, 256k context)
- Qwen3.5-397B-A17B (403.4B MoE, 256k context)

### Updated context length detection in `scrape_hf_models.py`                                                                                                                                                  

**Problem:** The scraper was returning `context_length: 4096` instead of the correct `262144` for Qwen3.5 models.                                                                                                                                                                                                   
**Cause:** Qwen3.5 models have a nested config structure. The `infer_context_length()`function only checked top-level config keys, missing the nested text_config.max_position_embeddings.
**Fix**: Updated infer_context_length() to also check text_config for multimodal models after checking top-level keys.

### DIFF
--- a/scripts/scrape_hf_models.py
+++ b/scripts/scrape_hf_models.py
@@ -78,6 +78,10 @@ TARGET_MODELS = [
     "Qwen/Qwen3-30B-A3B",
     "Qwen/Qwen3-235B-A22B",
     "Qwen/Qwen3-Coder-480B-A35B-Instruct",
+    "Qwen/Qwen3.5-27B",
+    "Qwen/Qwen3.5-35B-A3B",
+    "Qwen/Qwen3.5-122B-A10B",
+    "Qwen/Qwen3.5-397B-A17B",
     # Microsoft Phi
     "microsoft/phi-3-mini-4k-instruct",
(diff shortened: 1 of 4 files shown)


scripts/scrape_hf_models.py
```diff
@@ -337,18 +344,31 @@ def infer_context_length(config: dict | None) -> int:
...
+
+    # Check top-level config
+    for key in keys_to_check:
         if key in config:
             val = config[key]
             if isinstance(val, int) and val > 0:
                 return val
+
+    # For multimodal models (e.g., Qwen3.5), check text_config
+    if "text_config" in config and isinstance(config["text_config"], dict):
```
reneleonhardt, line 365: Default value `None` would have sufficed 🙂
```suggestion
    if isinstance(config.get("text_config"), dict):
```

SirLorrence: Well the small qwen3.5 models have been dropped (0.8b - 9b) so i can fix that.

PR #131, opened 2026-03-01
feat: add runtime model database updates from HuggingFace

saiteja007-mv:
## Summary

- **New `llmfit update` command** — fetches the latest trending and top-downloaded text-generation models from the HuggingFace API and caches them locally. The cache is auto-merged at startup so the model list stays current without a binary rebuild.
- **`Added to HF` date column** — all CLI table outputs now show when each model was added to HuggingFace.
- **`--sort` flag on `list` and `fit`** — allows sorting by `date`, `params`, `ctx`, `mem`, or `score`.

## New command: `llmfit update`

```sh
llmfit update                      # fetch 100 trending + 50 top-downloaded
llmfit update --trending 200       # custom limit
llmfit update --token hf_xxx       # with HF auth token (also via HF_TOKEN env)
llmfit update --status             # show cache info
llmfit update --clear              # reset to embedded-only list
```

Models are saved to `~/.llmfit/hf_models_cache.json` (Linux/macOS) or `%APPDATA%\llmfit\hf_models_cache.json` (Windows). `ModelDatabase::new()` automatically merges cached models on every invocation — **no rebuild required**.

## Date column & sorting

```sh
llmfit list --sort date            # newest models first (default)
llmfit list --sort params          # largest models first
llmfit fit  --sort date -n 10      # 10 newest compatible models
llmfit fit  --sort score           # best fit score (default)
```

The TUI already had the `Date` column and sort; this PR aligns the CLI table output.

## Implementation

| File | Change |
|------|--------|
| `llmfit-core/src/update.rs` | **New** — HF API fetch, cache I/O, param/use-case/context/RAM heuristics, 13 unit tests |
| `llmfit-core/src/models.rs` | `ModelDatabase::embedded()` helper; `::new()` auto-merges cache |
| `llmfit-core/src/lib.rs` | Re-exports for new update module |
| `llmfit-tui/src/display.rs` | `Added to HF` column in all CLI tables |
| `llmfit-tui/src/main.rs` | `Update` subcommand + `--sort` flags on `list`/`fit` |

## Test plan

- [ ] `cargo test -p llmfit-core` — all 78 tests pass (13 new in `update.rs`)
- [ ] `llmfit update` — fetches models and reports added count
- [ ] `llmfit update --status` — shows cache path and model count
- [ ] `llmfit update --clear` — resets cache
- [ ] `llmfit list --sort date` — newest models appear at top
- [ ] `llmfit fit --sort date` — newest compatible models appear first
- [ ] `llmfit list` — `Added to HF` column is visible in table

### DIFF
--- a/llmfit-core/src/update.rs
+++ b/llmfit-core/src/update.rs
@@ -72,20 +72,17 @@ pub fn load_cache() -> Vec<LlmModel> {
 
 /// Persist a model list to the cache file, creating the directory if needed.
 pub fn save_cache(models: &[LlmModel]) -> Result<(), String> {
-    let path = cache_file()
-        .ok_or_else(|| "Cannot determine cache directory".to_string())?;
+    let path = cache_file().ok_or_else(|| "Cannot determine cache directory".to_string())?;
     if let Some(dir) = path.parent() {
-        std::fs::create_dir_all(dir)
-            .map_err(|e| format!("Failed to create cache dir: {e}"))?;
+        std::fs::create_dir_all(dir).map_err(|e| format!("Failed to create cache dir: {e}"))?;
     }
     let envelope = CacheEnvelope {
         version: CACHE_VERSION,
         models: models.to_vec(),
     };
-    let json = serde_json::to_string_pretty(&envelope)
-        .map_err(|e| format!("Serialize error: {e}"))?;
-    std::fs::write(&path, json)
-        .map_err(|e| format!("Failed to write cache: {e}"))?;
+    let json =
+        serde_json::to_string_pretty(&envelope).map_err(|e| format!("Serialize error: {e}"))?;
+    std::fs::write(&path, json).map_err(|e| format!("Failed to write cache: {e}"))?;
     Ok(())
 }
 
@@ -96,8 +93,7 @@ pub fn clear_cache() -> Result<usize, String> {
         _ => return Ok(0),
     };
     let count = load_cache().len();
-    std::fs::remove_file(&path)
-        .map_err(|e| format!("Failed to delete cache: {e}"))?;
+    std::fs::remove_file(&path).map_err(|e| format!("Failed to delete cache: {e}"))?;
     Ok(count)
 }
 
@@ -153,19 +149,26 @@ fn parse_expert_suffix(s: &str) -> Option<u32> {
 /// from a model identifier string.
 fn extract_model_params(
     model_id: &str,
-) -> (String, Option<u64>, bool, Option<u32>, Option<u32>, Option<u64>) {
+) -> (
+    String,
+    Option<u64>,
+    bool,
+    Option<u32>,
+    Option<u32>,
+    Option<u64>,
+) {
     let up = model_id.to_uppercase();
     // Split on typical separators but NOT on '.' so that "1.5B" stays intact.
-    let tokens: Vec<&str> = up.split(|c: char| matches!(c, '-' | '/' | '_' | ' ' | ':')).collect();
+    let tokens: Vec<&str> = up
+        .split(|c: char| matches!(c, '-' | '/' | '_' | ' ' | ':'))
+        .collect();
 
     // MoE pattern "8X7B": N experts × M params each.
     for tok in &tokens {
         if let Some(x_pos) = tok.find('X') {
             let (left, right) = tok.split_at(x_pos);
             let right = &right[1..]; // skip 'X'
-            if let (Ok(n_exp), Some(per_exp)) =
-                (left.parse::<u32>(), parse_param_str(right))
-            {
+            if let (Ok(n_exp), Some(per_exp)) = (left.parse::<u32>(), parse_param_str(right)) {
                 if (2..=512).contains(&n_exp) {
                     let total = per_exp.saturating_mul(n_exp as u64);
                     let active_exp = 2u32.min(n_exp);
@@ -176,7 +179,14 @@ fn extract_model_params(
                     } else {
                         format!("{}x{}M", n_exp, per_exp / 1_000_000)
                     };
-                    return (s, Some(total), true, Some(n_exp), Some(active_exp), Some(active));
+                    return (
+                        s,
+                        Some(total),
+                        true,
+                        Some(n_exp),
+                        Some(active_exp),
+                        Some(active),
+                    );
                 }
             }
         }
@@ -184,9 +194,7 @@ fn extract_model_params(
 
     // MoE pattern "17B-16E": per-expert params + expert count.
     for window in tokens.windows(2) {
-        if let (Some(pb), Some(ne)) =
-            (parse_param_str(window[0]), parse_expert_suffix(window[1]))
-        {
+        if let (Some(pb), Some(ne)) = (parse_param_str(window[0]), parse_expert_suffix(window[1])) {
             if (2..=512).contains(&ne) {
                 let total = pb.saturating_mul(ne as u64);
                 let ae = 2u32.min(ne);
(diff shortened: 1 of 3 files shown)


NichyX:
any update on this?

saiteja007-mv:
> any update on this?

Not yet

llmfit-core/src/models.rs
```diff
@@ -293,32 +293,59 @@ impl Default for ModelDatabase {
...
+            active_parameters: e.active_parameters,
+            release_date: e.release_date,
+            gguf_sources: e.gguf_sources,
+        })
+        .collect()
+}
+
 impl ModelDatabase {
+    /// Load only the compile-time embedded model list (no cache).
+    /// Used internally by the updater to determine which models are already known.
```
AlexsJones, line 504: Name-only dedup is fragile. If HuggingFace renames a model or the embedded list uses a slightly different ID format (e.g. `meta-llama/Llama-3.1-8B` vs `meta-llama/llama-3.1-8b`), duplicates could slip trough. case-insensitive helps but doesn't cover all variations. Consider normalizing further or matching on a canonical slug.

llmfit-core/src/update.rs
```diff
@@ -0,0 +1,573 @@
...
+                    format!("{}B", b)
+                }
+            } else {
+                format!("{}M", raw / 1_000_000)
+            };
+            return (s, Some(raw), false, None, None, None);
+        }
+    }
+
+    ("7B".to_string(), None, false, None, None, None)
```
AlexsJones: Defaulting to `("7B", None, ...)` when param parsing fails means some models will silently appear as 7B with fabricated RAM esitmates. consider returning an `Option` or marking these as "unknown" so they don't mislead users.

llmfit-core/src/update.rs
```diff
@@ -0,0 +1,573 @@
...
+                if frac > 0 {
+                    format!("{}.{}B", b, frac)
+                } else {
+                    format!("{}B", b)
+                }
+            } else {
+                format!("{}M", total / 1_000_000)
+            };
+            (s, Some(total), false, None, None, None)
+        } else {
```
AlexsJones, line 395: When `safetensors.total` is available, `is_moe` is hardcoded to `false`. This means MoE models with safetensors metadata (e.g. Mixtral) loose their MoE classification and get inaccurate VRAM estimates. consider falling back to name-based MoE detection even when safetensors provides the param count — use safetensors for total param accuracy but still parse the name for architecture hints.

llmfit-core/src/update.rs
```diff
@@ -0,0 +1,573 @@
...
+        ureq::get(&url)
+            .config()
+            .timeout_global(Some(std::time::Duration::from_secs(30)))
+            .build()
+            .call()
+    };
+    match resp {
+        Ok(r) => r.into_body().read_json::<Vec<HfApiModel>>().unwrap_or_default(),
+        Err(_) => vec![],
+    }
```
AlexsJones, line 356: All errors here are silently swallowed — HTTP failures, rate limiting, auth errors, malformed JSON all return an empty vec. the user only sees "No models fetched" which dosn't help diagnose the issue. Consider propagating the error (or at least logging it) so users know if it's a 401 (bad token), 429 (rate limited), DNS failure, etc.

llmfit-core/src/update.rs
```diff
@@ -0,0 +1,573 @@
...
+    let path = match cache_file() {
+        Some(p) if p.exists() => p,
+        _ => return vec![],
+    };
+    let Ok(content) = std::fs::read_to_string(&path) else {
+        return vec![];
+    };
+    serde_json::from_str::<Vec<LlmModel>>(&content).unwrap_or_default()
+}
+
```
AlexsJones, line 72: No cache versioning or invalidation. if a new binary ships with corrected specs for a model (or removes one), stale cached entrys persist indefinitely. Consider adding a schema version or embedding a hash of the embedded list so the cache can be auto-cleared on upgrades.

llmfit-core/src/update.rs
```diff
@@ -0,0 +1,573 @@
...
+    Some(LlmModel {
+        name: hf.id,
+        provider,
+        parameter_count: param_str,
+        parameters_raw: params_raw,
+        min_ram_gb: min_ram,
+        recommended_ram_gb: rec_ram,
+        min_vram_gb: min_vram,
+        quantization: "Q4_K_M".to_string(),
+        context_length,
```
AlexsJones, line 428: Hardcoding quantization to `Q4_K_M` for all fetched models is a rough aproximation. the RAM/VRAM estimates downstream depend on this. Worth noting this limitation in a comment, or consider making it configurable.

llmfit-core/src/models.rs
saiteja007-mv: Good catch. Added `canonical_slug()` in `models.rs` that strips the `org/` prefix, lowercases, and collapses `-`/`_`/`.` before comparing — so `meta-llama/Llama-3.1-8B` and `meta-llama/llama-3.1-8b` now deduplicate correctly. Applied consistently in both `ModelDatabase::new()` and `update_model_cache()`.

llmfit-core/src/update.rs
saiteja007-mv: Fixed. The fallback now returns `"Unknown"` as the param label instead of `"7B"`, so models with unparseable names no longer appear to be 7B in the table. The `estimate_ram` call still uses 7B as a conservative RAM heuristic for these models (better than crashing or showing 0), which is noted in the surrounding comment.

llmfit-core/src/update.rs
saiteja007-mv: Good point — fixed. When safetensors provides the exact total param count, we now still run \ on the model name to obtain the MoE flags (\, \, \, \), and use safetensors only for the numeric precision. Mixtral-style models will correctly keep their MoE classification and accurate VRAM estimates.

llmfit-core/src/update.rs
saiteja007-mv: Fixed. \ now returns \ with actionable messages — 401 tells the user to check \, 429 tells them to wait, and other errors include the raw message for diagnosis. Each fetch (trending, downloads) reports warnings via the progress callback on failure but doesn't abort — if both fail the existing 'No models fetched' error is still returned.

llmfit-core/src/update.rs
saiteja007-mv: Added cache versioning via a CACHE_VERSION constant and a CacheEnvelope wrapper struct. The cache is now saved as {"version": 1, "models": [...]}. On load, if the version doesn't match (or the file is in the old unversioned format), it is silently discarded and a fresh fetch is triggered on next update. Bumping CACHE_VERSION in future releases is enough to force a cache refresh.

llmfit-core/src/update.rs
saiteja007-mv: Agreed — added a comment directly above the field explaining that Q4_K_M is a conservative approximation and that actual available quantizations depend on the GGUF files published for each model. Making it configurable per-model would require the HF API to return quant metadata, which it currently does not expose for most models.

AlexsJones:
@saiteja007-mv please resolve the conflict - ping me @ diretly after so I can prioritise this
Also formatting needs applying again

saiteja007-mv:
@AlexsJones Conflicts resolved and `cargo fmt` applied — branch is ready to merge.

saiteja007-mv:
@AlexsJones Rebased onto latest main — conflict in `models.rs` resolved (added `ModelFormat::default()` to new `format` field). Branch is clean and ready to merge.

NichyX:
Seems good! @saiteja007-mv @AlexsJones

Issue #88, opened 2026-02-25
Observation regarding low tok/s numbers:


kursad-k:
Hi,

I can achieve 270 tok/s with Nemotron-3-Nano-30B-A3B-GGUF on a locally running Lemonade server. However, llmfit lists a similar model at only 8.8 tok/s for my system.

`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8 nvidia 31.6B 54 8.8 Q4_K_M GPU`

This number seems significantly lower than expected, it isn't even close to the performance I'm seeing.

three-foxes-in-a-trenchcoat:
Thanks for reporting this @kursad-k! 🦊

This is a really interesting observation. The tok/s estimates in llmfit are based on theoretical bandwidth calculations and conservative heuristics rather than real-world benchmarks from specific inference engines. A few things that could explain the 270 vs 8.8 gap:

1. **MoE architecture**: Nemotron-3-Nano-30B-A3B is a Mixture-of-Experts model — only ~3B parameters are active per forward pass despite the 30B total. llmfit likely uses the full parameter count (31.6B) for its throughput estimate, which would massively underestimate actual MoE performance.

2. **Inference engine optimizations**: Lemonade (and engines like llama.cpp, vLLM, etc.) apply speculative decoding, KV-cache optimizations, and engine-specific tricks that aren't captured in llmfit's static model.

3. **Quantization differences**: You're seeing Q4_K_M in llmfit vs FP8 in practice — the quantization format and how the engine handles it matters a lot.

This relates to #80 (ideal inference engine detection) as well — real-world performance varies significantly by engine. We might want to:
- Flag MoE models so the estimator can use active parameter count instead of total
- Add a disclaimer that estimates are theoretical/conservative
- Eventually allow user-submitted benchmarks (ties into #81)

Would you be able to share your GPU model and VRAM? That'd help calibrate the estimates.

three-foxes-in-a-trenchcoat:
Hey @kursad-k 👋

Good catch — the tok/s estimates in llmfit are based on conservative benchmarks and generic throughput curves, not runtime-specific optimizations. Lemonade Server likely uses speculative decoding, continuous batching, or other inference optimizations that dramatically improve throughput beyond what a naive "load model, run inference" estimate gives.

A few things that could help close the gap:

**1. The estimation model doesn't account for inference engine optimizations**

[...]

PR #137, opened 2026-03-02
feat: bandwidth-based tok/s estimation for known GPUs

kira-ariaki:
## Summary

Replace the fixed per-backend speed constant (`K=220` for all CUDA GPUs) with a physics-grounded estimation model based on actual GPU memory bandwidth. This significantly improves tok/s accuracy for known GPUs while maintaining backward compatibility for unrecognized hardware.

## Problem

The current `estimate_tps()` uses a single constant `K=220` for all CUDA GPUs, regardless of whether it's a GTX 1650 (128 GB/s) or an RTX 4090 (1008 GB/s). This produces estimates that can be **4–30x lower than actual performance** on high-end cards.

Example (Qwen3.5-27B Q4_K_M on RTX 4090):
| | Old estimate | New estimate | Measured |
|---|---|---|---|
| tok/s | 9.1 | 39.9 | ~40 |

This was reported in #88, where a user observed 270 tok/s on their system while llmfit estimated 8.8.

## Approach

LLM token generation is **memory-bandwidth-bound** — each generated token requires reading the full model weights once from VRAM. The theoretical throughput ceiling is:

```
max_tps = memory_bandwidth_GB/s ÷ model_size_GB
```

Real-world efficiency is ~50–70% of this ceiling due to kernel overhead, KV-cache reads, and memory controller effects.

This PR:
1. Adds a `gpu_memory_bandwidth_gbps()` lookup table (~80 GPUs: NVIDIA consumer/datacenter, AMD RDNA/CDNA, Apple Silicon)
2. Adds `quant_bytes_per_param()` to compute model size from quantization format
3. Updates `estimate_tps()` in both `fit.rs` and `plan.rs` to use bandwidth-based estimation when the GPU is recognized
4. **Falls back to the original fixed-constant approach** for unrecognized GPUs — no regressions for unknown hardware

## Theoretical basis

- **kipply, "[Transformer Inference Arithmetic](https://kipp.ly/transformer-inference-arithmetic/)" (2022)** — First-principles derivation showing token generation is memory-bandwidth-bound at batch size 1
- **Google, "[Efficiently Scaling Transformer Inference](https://arxiv.org/abs/2211.05102)" (arXiv:2211.05102)** — Confirms bandwidth-bound behavior on TPU v4 for low-batch inference
- **ggerganov, [Apple Silicon benchmarks](https://github.com/ggml-org/llama.cpp/discussions/4167) (Discussion #4167)** — Official llama.cpp data showing near-linear tok/s scaling with memory bandwidth across M1/M1 Pro/M1 Max/M1 Ultra
- **ggerganov, [T4 benchmarks](https://github.com/ggml-org/llama.cpp/discussions/4225) (Discussion #4225)** — T4 (320 GB/s) running 7B F16 at ~16 tok/s, consistent with bandwidth model

## Validation

| GPU | Model | Quant | Old | New | Measured | Source |
|-----|-------|-------|-----|-----|----------|--------|
| RTX 4090 (1008 GB/s) | 27B | Q4_K_M | 9.1 | 39.9 | ~40 | Our hardware |
| T4 (320 GB/s) | 7B | F16 | ~10 | ~14 | ~16 | ggerganov #4225 |
| M1 Max (400 GB/s) | 7B | Q4_0 | — | — | 61 | ggerganov #4167 |

## Changes

- `hardware.rs`: Add `gpu_memory_bandwidth_gbps()` — lookup table for ~80 GPUs
- `models.rs`: Add `quant_bytes_per_param()` — bytes per parameter for each quantization format
- `fit.rs`: Update `estimate_tps()` to prefer bandwidth-based estimation
- `plan.rs`: Update `estimate_tps_with_gpu()` and callers to thread GPU name through plan mode

## Tests

9 new tests added (84 total, all passing):
- Bandwidth lookup correctness (NVIDIA, AMD, Apple Silicon, unknown GPUs)
- RTX 4090 vs RTX 3060 speed comparison
- RTX 4090 27B Q4 realistic range check
- T4 7B F16 realistic range check
- Unknown GPU fallback behavior
- CPU-only mode ignores GPU bandwidth

Partially addresses #88.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,15 @@ llmfit plan "Qwen/Qwen2.5-Coder-0.5B-Instruct" --context 8192 --json
 
    Dimensions are combined into a weighted composite score. Weights vary by use-case category (General, Coding, Reasoning, Chat, Multimodal, Embedding). For example, Chat weights Speed higher (0.35) while Reasoning weights Quality higher (0.55). Models are ranked by composite score, with unrunnable models (Too Tight) always at the bottom.
 
-5. **Speed estimation** -- Estimated tokens per second using backend-specific constants:
+5. **Speed estimation** -- Token generation in LLM inference is memory-bandwidth-bound: each token requires reading the full model weights once from VRAM. When the GPU model is recognized, llmfit uses its actual memory bandwidth to estimate throughput:
+
+   Formula: `(bandwidth_GB_s / model_size_GB) × efficiency_factor`
+
+   The efficiency factor (0.55) accounts for kernel overhead, KV-cache reads, and memory controller effects. This approach is validated against published benchmarks from llama.cpp ([Apple Silicon](https://github.com/ggml-org/llama.cpp/discussions/4167), [NVIDIA T4](https://github.com/ggml-org/llama.cpp/discussions/4225)) and real-world measurements.
+
+   The bandwidth lookup table covers ~80 GPUs across NVIDIA (consumer + datacenter), AMD (RDNA + CDNA), and Apple Silicon families.
+
+   For unrecognized GPUs, llmfit falls back to per-backend speed constants:
 
    | Backend | Speed constant |
    |---|---|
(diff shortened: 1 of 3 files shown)


AlexsJones:
@kira-ariaki thanks this is a good step in the right direction.
Please 1: cargo fmt & update the readme https://github.com/AlexsJones/llmfit?tab=readme-ov-file#how-it-works

AlexsJones (changes requested):
cargo fmt + update readme

kira-ariaki:
Thanks for the review! Applied `cargo fmt` and updated the README's "Speed estimation" section to document the bandwidth-based approach + fallback behavior. CI should be green now.

PR #144, opened 2026-03-02
Harden GGUF filename validation in repo listing

maciejmaczko:
- filter unsafe HF tree entries in list_repo_gguf_files via shared validation
- keep traversal/absolute/subdirectory protections
- allow explicit current-dir form ./model.gguf
- add focused tests for filtering and filename validation edge cases

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -27,7 +27,7 @@ jobs:
         uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@5f0a5eb7e47090af52db6c4e7f4c30fb76fc17ee # v3.9.0
+        uses: docker/setup-buildx-action@v3
 
       - name: Log in to GitHub Container Registry
         uses: docker/login-action@830126ccf9677aa6e98b8f5a8b7ea177ab1f8e63 # v4.0.0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2097,7 +2097,7 @@ checksum = "11d3d7f243d5c5a8b9bb5d6dd2b1602c0cb0b9db1621bafc7ed66e35ff9fe092"
 
 [[package]]
 name = "llmfit"
-version = "0.5.5"
+version = "0.5.7"
 dependencies = [
  "clap",
  "colored",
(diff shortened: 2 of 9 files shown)


AlexsJones:
I like the hardening direction here.
One requested change before merge:

Please disallow [./model.gguf](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (and any non-basename form) in [validate_gguf_filename](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
For cache writes, safest invariant is: exactly one plain filename ending in .gguf, with no path components at all.
Rationale:

and asaccepting ./... expands allowed input without real benefit.
A strict basename-only rule is easier to reason about and prevents future path-handling regressions.
Suggested test additions:

ensure these are rejected: [./model.gguf](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [model.gguf/](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [.\\model.gguf](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and any absolute/prefixed path form.
Keep existing positive cases like [model.gguf](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and Llama-3.1-8B-Q4_K_M.gguf.

after that thiis looks good to approve.ty

AlexsJones (changes requested):
See PR comments please

maciejmaczko:
@AlexsJones Totally fair point. I intentionally allowed ./ at first because it refers to the current directory, not parent traversal like ../. In practice, model.gguf and ./././model.gguf resolve to the same target.

That said, I’ve now updated the PR exactly as requested: validate_gguf_filename enforces strict basename-only input (no ./, no path components), and I added tests for the rejected forms you listed (./model.gguf, model.gguf/, .\\model.gguf, and absolute/prefixed paths), while keeping valid basename cases.

The PR is updated. After this is merged, I’ll look for another area in the project where I can help.

I’m new to Rust, so thanks for the patience and guidance.

Issue #147, opened 2026-03-02
The "--json" flag of the "list" command appears not to work.


sammwyy:
When using the command "llmfit list --help", the help message tells me that there is a "--json" flag to be able to display the result in JSON, the program ignores it and shows me the result in an ASCII table.

OS: Windows 10.
Bin version: `llmfit --version\nllmfit 0.5.5` (0.5.5)
Downloaded from cargo (`cargo install llmfit`)

```bash
~> llmfit list --help
List all available LLM models

Usage: llmfit.exe list [OPTIONS]

Options:
      --json  Output results as JSON (for tool integration)
  -h, --help  Print help
~> llmfit list --json

=== Available LLM Models ===
Total models: 514

╭────────┬───────────────────────────────────────────────────────────────┬────────────────────────────────┬─────────┬───────┬────────────┬────────┬─────────┬──────┬───────┬─────────╮
│ Status │ Model                                                         │ Provider                       │ Size    │ Score │ tok/s est. │ Quant  │ Runtime │ Mode │ Mem % │ Context │
├────────┼───────────────────────────────────────────────────────────────┼────────────────────────────────┼─────────┼───────┼────────────┼────────┼─────────┼──────┼───────┼─────────┤
│ --     │ echarlaix/tiny-random-PhiForCausalLM                          │ echarlaix                      │ 80K     │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 0k      │
│ --     │ peft-internal-testing/tiny-random-GPT2LMHeadModel             │ peft-internal-testing          │ 83K     │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 0k      │
│ --     │ peft-internal-testing/tiny-random-gpt2                        │ peft-internal-testing          │ 112K    │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 0k      │
│ --     │ peft-internal-testing/tiny-random-GPTJForCausalLM             │ peft-internal-testing          │ 129K    │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 0k      │
│ --     │ allenai/Olmo-3-7B-Instruct                                    │ allenai                        │ 528K    │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 65k     │
│ --     │ allenai/Olmo-3-7B-Think                                       │ allenai                        │ 528K    │ -     │ -          │ Q4_K_M │ -       │ -    │ -     │ 65k     │
[...]

PR #149, opened 2026-03-02
feat: multiple display mode and json flag for list command

sammwyy:
I moved the display.rs file from the TUI (CLI mode) to its own module, src/display/mod.rs, as a generic trait.

I created two implementations: json_mode.rs and table_mode.rs, with the possibility of adding more in the future.

This way, each file handles its own implementation. In the CLI, I simply initialize the display mode based on the global flag "--json" instead of using an if statement in each subcommand.

I also added support for the "--json" flag to the "list" command.


three-foxes-in-a-trenchcoat (approved):
Clean refactor! The trait-based `DisplayMode` approach is well-structured — makes it easy to add more output formats down the road (CSV, YAML, etc.). The `mod.rs` factory function is a nice touch.

The JSON mode fixes #147 properly, and moving the display logic out of the monolithic `display.rs` into separate modules keeps things organized.

A couple of minor thoughts (non-blocking):
- The shared `round1`/`round2` helpers are a nice detail for consistent formatting
- Future: could consider a `--format` flag that takes `json|table|csv` instead of just `--json`, but that's a follow-up

LGTM 🦊

AlexsJones:
Please rebase and we can merge @sammwyy

sammwyy:
already

AlexsJones:
Closing as superseded, with apologies for the long silence. `llmfit list --json` works on current main (verified on 0.9.34, and another user confirmed on 0.9.25), so the core fix here has been covered. The branch is now ~80 files behind main including regenerated catalog data, so a rebase isn't practical. If the additional display modes are still something you'd like, a fresh PR against current main would be very welcome — the list rendering now lives in llmfit-tui/src/display.rs. Thanks for the contribution and for reporting #147 in the first place.

PR #162, opened 2026-03-04
feat: add model comparison workflow (llmfit diff) and TUI compare mode

AissamDjahnine:
## Summary
Adds a first-class model comparison workflow in both CLI and TUI.
- CLI: new diff subcommand for side-by-side comparison.
- TUI: new mark-and-compare flow to compare two selected models interactively.

This helps users choose between close candidates (for example two Perfect fits) with direct metric deltas.

### 1) New CLI command: llmfit diff
Added a new subcommand:

```bash
llmfit diff [model_a] [model_b] [--fit <...>] [--sort <...>] [-n <...>]
```

#### Behavior:
-   If model_a and model_b are provided: compares those two models.
-   If model names are omitted: auto-selects top -n models after applying --fit + --sort.
-   Requires at least 2 models after filtering.
-   Supports JSON output via existing global --json flag.

#### Supported options:
    --fit: all | perfect | good | marginal | tight | runnable
    --sort: score | tps | params | mem | ctx | date | use
    -n, --limit: number of models in auto-compare mode (default 2, minimum 2)

### 2) New CLI diff renderer
Added a dedicated side-by-side comparison output with:
- Score (+/- delta vs baseline), Baseline tok/s (+/- delta), Fit level, Run mode, Runtime, Memory utilization (+/- delta), Params, Context, Best quant, Provider

### 3) New TUI compare workflow

- Compare pane now also shows Provider, Use, Released, Quant (best + default), Badges (MoE/Offload/Notes), and delta hints (↑/↓/=).
- Added compare state and keybindings in TUI:
    -   m: mark current model for compare
    -   c: compare marked model vs current selected model (toggle on/off)
    -   x: clear marked model

#### UI behavior:

-   Marked model is shown with * in the first table column.
-   Compare pane shows two side-by-side panels:
      - left: marked baseline model
      - right: selected model with metric deltas vs baseline
-   Status bar help updated with compare shortcuts.

### 4) Tests
Added unit tests for new CLI helper logic:

- fit filter behavior (Runnable excludes TooTight)
- model selector resolution prefers exact matches
- ambiguous partial selector returns a clear error

## Examples
CLI auto-compare top perfect models
```bash
llmfit diff --fit perfect --sort score -n 2
```

CLI compare top runnable models by throughput
```bash
llmfit diff --fit runnable --sort tps -n 3
```

CLI explicit compare by model names
```bash
llmfit diff "moonshotai/Moonlight-16B-A3B-Instruct" "Qwen/Qwen2.5-14B-Instruct"
```

#### TUI compare flow

- Move to model A and press m
image

- Move to model B and press c
image

- Press c again to return to table
- Press x to clear mark

## Validation

- cargo fmt passes
- Manual CLI verification: llmfit diff --fit perfect --sort score -n 2 side-by-side output and deltas confirmed
- Manual TUI workflow verified with m/c/x interactions

## Notes

- Comparison deltas are relative to the first (baseline) model in the selected comparison set.
- Existing fit/ranking logic is reused; no scoring algorithm changes were introduced
- JSON output was simplified per review by deriving Serialize on ModelFit and using direct serde_json::to_string_pretty(...) paths (including list --json).

### DIFF
--- a/llmfit-tui/src/tui_ui.rs
+++ b/llmfit-tui/src/tui_ui.rs
@@ -13,8 +13,7 @@ use crate::theme::ThemeColors;
 use crate::tui_app::{
     App, AvailabilityFilter, DownloadCapability, DownloadProvider, FitFilter, InputMode, PlanField,
 };
-use llmfit_core::fit::FitLevel;
-use llmfit_core::fit::SortColumn;
+use llmfit_core::fit::{FitLevel, ModelFit, SortColumn};
 use llmfit_core::hardware::is_running_in_wsl;
 use llmfit_core::providers;
 
@@ -557,9 +556,9 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect, tc: &ThemeColors) {
             };
 
             let marker = if app.compare_mark_model.as_deref() == Some(fit.model.name.as_str()) {
-                "*"
+                format!("{}*", fit_indicator(fit.fit_level))
             } else {
-                fit_indicator(fit.fit_level)
+                fit_indicator(fit.fit_level).to_string()
             };
 
             Row::new(vec![
@@ -727,9 +726,9 @@ fn draw_compare(frame: &mut Frame, app: &App, area: Rect, tc: &ThemeColors) {
     } else {
         " ="
     };
-    let mem_hint = if mem_delta < -0.05 {
+    let mem_hint = if mem_delta > 0.05 {
         " ↑"
-    } else if mem_delta > 0.05 {
+    } else if mem_delta < -0.05 {
         " ↓"
     } else {
         " ="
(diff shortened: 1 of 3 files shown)


AlexsJones (changes requested):
Thanks for thinking about this.

I believe this is a premature abstraction we don't necessarily need. Having almost 300 lines of hand crafted json macros doesn't feel right. 


My proposal for you to implement:
```
#[derive(serde::Serialize)] 
```
to ModelFit in llmfit-core/src/fit.rs and `println!("{}", serde_json::to_string_pretty(&fits).unwrap());`

e.g. 
```
 if json {
      println!("{}", serde_json::to_string_pretty(&models).unwrap());
  } else {
      display_all_models(&models);
  }


```

The useful part of this PR is adding --json to the list command. Deriving Serialize on
ModelFit would get JSON output for free everywhere with near-zero code.

llmfit-tui/src/tui_ui.rs
```diff
@@ -644,6 +652,382 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect, tc: &ThemeColors) {
     }
 }
 
+fn draw_compare(frame: &mut Frame, app: &App, area: Rect, tc: &ThemeColors) {
```
AlexsJones, line 654: These two panels are almost identical and like 100 lines each, can we move into a helper perhaps?

llmfit-core/src/fit.rs
```diff
@@ -89,7 +89,7 @@ pub struct ScoreComponents {
     pub context: f64,
 }
 
-#[derive(Clone)]
+#[derive(Clone, serde::Serialize)]
```
AlexsJones, line 92: Field names like "category" (from use_case.label()) and "fit_level" (from fit_text()) will now serialize as raw Rust enum variants instead of human-readable strings... Probably the best way forward here is to use Serialize for the new diff --json path, but leave the existing            
  system_json()/fit_to_json() functions in place for list --json and fit --json so the output doesn't change.

llmfit-tui/src/tui_ui.rs
```diff
@@ -554,8 +556,14 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect, tc: &ThemeColors) {
                 Style::default()
             };
 
+            let marker = if app.compare_mark_model.as_deref() == Some(fit.model.name.as_str()) {
+                "*"
```
AlexsJones: When a model is marked for compare, * fully replaces the fit indicator (the colored dot) in that row.

llmfit-tui/src/tui_ui.rs
```diff
@@ -644,6 +652,382 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect, tc: &ThemeColors) {
...
+        " ="
+    };
+    let tps_hint = if tps_delta > 0.05 {
+        " ↑"
+    } else if tps_delta < -0.05 {
+        " ↓"
+    } else {
+        " ="
+    };
+    let mem_hint = if mem_delta < -0.05 {
```
AlexsJones: For every other metric, ↑ means the value increased. Here ↑ means "better".. which is the opposite direction. The
  code works, it's just unintuitive to read without context.

llmfit-tui/src/tui_ui.rs
AissamDjahnine: Refactored `draw_compare` by extracting shared rendering into:
- `render_compare_panel(...)`
- `compare_badges(...)`
- `CompareMetrics`

This removes duplicated panel code while preserving behavior.

llmfit-core/src/fit.rs
AissamDjahnine: Agreed. I restored the existing handcrafted JSON mapping (`system_json` / `fit_to_json`) for existing JSON outputs to keep backward-compatible, human-readable fields.

Serde-derived output is now used only for the new `diff --json` path (`display_json_diff_fits(...)`).

llmfit-tui/src/tui_ui.rs
AissamDjahnine: Fixed indicator rendering so mark state is additive: the row now shows `fit_indicator + "*"`, preserving the colored fit dot while still indicating marked state.

llmfit-tui/src/tui_ui.rs
AissamDjahnine: Updated delta arrows to consistently represent direction of change:
- `↑` value increased
- `↓` value decreased
- `=` unchanged

Also added an in-view legend clarifying that for Mem%, lower is better.

AlexsJones:
I like this, thanks - I am not sure how intuitive the key bindings are though.

PR #168, opened 2026-03-05
fix: correct PCI ID reading on Linux and improve AMD GPU identification

Edythator:
This PR addresses the following issues within `llmfit_core::hardware`:
- `get_amd_gpu_name_lspci` did not take PCI IDs into account for matching, unlike its NVIDIA counterpart `get_nvidia_gpu_name_lspci`. This PR updates it to use slot hints.
- `/sys/class/drm/card*/device/uevent` returns "full" PCI IDs, which include their domain numbers. The existing functions did not account for this. To ensure accurate matching against the full PCI ID, the `-D` flag is now passed to `lspci` before matching the card's PCI IDs against the output.


AlexsJones:
Nice and clean, thanks

PR #169, opened 2026-03-05
Add copy function to TUI

commandlinetips:
I use this tool often. Sometimes I find a model that doesn't have a download source available, and I need to search for it manually. Typing the full name each time was tedious.

This adds a `y` keybinding that copies the selected model's full name to the clipboard.

Thanks for the tool.


README.md
```diff
@@ -1,5 +1,5 @@
 # llmfit
-
+this is a copy of the main repo just add a copy function
```
AlexsJones: Seems un-unecessary

rust.rule.md
```diff
@@ -0,0 +1,205 @@
+# Rust Coding Rules for llmfit
```
AlexsJones: Let's remove for now - I agree with the spirit but its outside of the PR scope

llmfit-tui/src/tui_events.rs
```diff
@@ -121,12 +124,15 @@ fn handle_search_mode(app: &mut App, key: KeyEvent) {
             app.clear_search();
         }
 
-        KeyCode::Char(c) => app.search_input(c),
+        // Allow copy while searching
+        KeyCode::Char('y') => app.copy_selected_model_name(),
```
AlexsJones: Remove the y keybinding from search mode (or move it after search_input so it doesn't break typing 'y')

.gitignore
```diff
@@ -1,4 +1,20 @@
+# Rust
 /target
```
AlexsJones, line 2: Please revert the gitignore changes

commandlinetips:
sorry for the late respond  i have make all the change you asked for , thank you for this cool tool

AlexsJones:
@commandlinetips sorry to ask but can you cherry pick your change onto a new check out of the repo? I had to force push the tree due to large objects in the history (myfault) which has caused a large cascade of commits piled onto your changes

commandlinetips:
i will open a new PR

Issue #74, opened 2026-02-24
llmfit doesn't detect CPU and GPU on Android


Nikitf777:
When I'm running llmfit on Android in Termux, it reports that I have an unknown CPU and no GPU:

![Screenshot_20260224-132118_Termux.png](https://github.com/user-attachments/assets/203fc90d-0151-4869-b63c-5fb268e53d57)

But my GPU is actually accessible in Termux via the Turnip driver (Vulkan works):

![Screenshot_20260224-161730_Termux.png](https://github.com/user-attachments/assets/dd886478-0d17-4c64-ac6a-43118f206204)

CPU info is also available in `/proc/cpuinfo` and other sources.

TheRealMaN:
Doesnt detect "AMD Radeon RX 570 4 GB" on macOS Tahoe

PR #177, opened 2026-03-08
Fix Android CPU and Vulkan GPU detection fallback

MahmoudAdelbghany:
## Summary
- add CPU detection fallbacks for Android and Termux when sysinfo returns empty or unknown CPU
- add Vulkan GPU detection fallback and skip software renderers like SwiftShader and llvmpipe
- add tests for CPU and Vulkan parsing paths

## Validation
- run Android emulator and collect hardware outputs via adb
- cargo test -p llmfit-core -q

Fixes https://github.com/AlexsJones/llmfit/issues/74

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -231,11 +231,9 @@ impl SystemSpecs {
 
         // Vulkan fallback (e.g. Android/Termux with Turnip)
         for vulkan_gpu in Self::detect_vulkan_gpu_info() {
-            let dominated = gpus.iter().any(|existing| {
-                let existing_lower = existing.name.to_lowercase();
-                let vulkan_lower = vulkan_gpu.name.to_lowercase();
-                existing_lower.contains(&vulkan_lower) || vulkan_lower.contains(&existing_lower)
-            });
+            let dominated = gpus
+                .iter()
+                .any(|existing| Self::is_same_gpu_name(&existing.name, &vulkan_gpu.name));
             if !dominated {
                 gpus.push(vulkan_gpu);
             }
@@ -1031,9 +1029,36 @@ impl SystemSpecs {
         }
     }
 
+    fn has_command(command: &str) -> bool {
+        let Some(path_var) = std::env::var_os("PATH") else {
+            return false;
+        };
+
+        for path in std::env::split_paths(&path_var) {
+            let candidate = path.join(command);
+            if candidate.is_file() {
+                return true;
+            }
+
+            #[cfg(target_os = "windows")]
+            for ext in [".exe", ".cmd", ".bat", ".com"] {
+                let candidate = path.join(format!("{command}{ext}"));
+                if candidate.is_file() {
+                    return true;
+                }
+            }
+        }
+
+        false
+    }
+
     /// Detect GPUs via Vulkan. This is especially useful on Android/Termux,
     /// where vendor-specific Linux utilities may be unavailable.
     fn detect_vulkan_gpu_info() -> Vec<GpuInfo> {
+        if !Self::has_command("vulkaninfo") {
+            return Vec::new();
+        }
+
         let output = match std::process::Command::new("vulkaninfo")
             .arg("--summary")
             .output()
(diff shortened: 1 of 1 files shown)


llmfit-core/src/hardware.rs
```diff
@@ -233,6 +229,18 @@ impl SystemSpecs {
...
+        // Vulkan fallback (e.g. Android/Termux with Turnip)
+        for vulkan_gpu in Self::detect_vulkan_gpu_info() {
+            let dominated = gpus.iter().any(|existing| {
+                let existing_lower = existing.name.to_lowercase();
+                let vulkan_lower = vulkan_gpu.name.to_lowercase();
+                existing_lower.contains(&vulkan_lower) || vulkan_lower.contains(&existing_lower)
+            });
+            if !dominated {
+                gpus.push(vulkan_gpu);
+            }
```
AlexsJones, line 239: the substring containment check both ways could produce false positves. e.g. a GPU named "RTX" would match "RTX 4090" even if they're diffrent entries. might want to normalize names or compare a canonical id instead.

llmfit-core/src/hardware.rs
```diff
@@ -1023,6 +1031,84 @@ impl SystemSpecs {
...
+            .arg("--summary")
+            .output()
+        {
+            Ok(o) if o.status.success() => o,
+            _ => match std::process::Command::new("vulkaninfo").output() {
+                Ok(o) if o.status.success() => o,
+                _ => return Vec::new(),
+            },
+        };
+
```
AlexsJones, line 1072: this spawns `vulkaninfo` twice on systems without vulkan. could check if the binery exists first rather than running two failing processes.

llmfit-core/src/hardware.rs
```diff
@@ -1023,6 +1031,84 @@ impl SystemSpecs {
...
+        for line in text.lines() {
+            let trimmed = line.trim();
+            if trimmed.is_empty() {
+                continue;
+            }
+
+            if let Some((key, value)) = trimmed.split_once('=')
+                && key.trim().eq_ignore_ascii_case("deviceName")
+            {
+                let name = value.trim();
```
AlexsJones: the `if let ... && let ...` chains are a relativley new rust feature (stabilized in 1.87.0). worth checking this doesn't break the project's MSRV.

llmfit-core/src/hardware.rs
```diff
@@ -1023,6 +1031,84 @@ impl SystemSpecs {
...
+        for name in Self::parse_vulkan_device_names(&text) {
+            if Self::is_software_vulkan_device(&name) {
+                continue;
+            }
+            *grouped.entry(name).or_insert(0) += 1;
+        }
+
+        grouped
+            .into_iter()
+            .map(|(name, count)| GpuInfo {
```
AlexsJones, line 1085: `infer_gpu_backend` probably won't match mobile GPUs like "Adreno (TM) 740" or "Mali-G78" since they don't hit any existing paterns. would be good to make sure the fallback backend is sensible here, Vulkan seems like the right chioce since that's how they were detected.

llmfit-core/src/hardware.rs
```diff
@@ -1159,6 +1245,68 @@ impl SystemSpecs {
...
+        None
+    }
+
+    fn read_android_soc_name() -> Option<String> {
+        let output = std::process::Command::new("getprop")
+            .arg("ro.soc.model")
+            .output()
+            .ok()?;
+        if !output.status.success() {
+            return None;
```
AlexsJones: minor but this and the `/proc/cpuinfo` reads could be gated behind `#[cfg(target_os = "linux")]` to make the platfrom intent explicit and skip unnecessary process spawns on macos/windows.

llmfit-core/src/hardware.rs
MahmoudAdelbghany: Fixed in bee8482. I replaced substring containment with normalized exact-name matching so short names like "RTX" no longer collapse into "RTX 4090".

llmfit-core/src/hardware.rs
MahmoudAdelbghany: Fixed in bee8482. I now check PATH for `vulkaninfo` first and return early when it is not installed, so we do not run the fallback probe on missing binary setups.

llmfit-core/src/hardware.rs
MahmoudAdelbghany: Good point. I rewrote the new Vulkan parser logic in this PR to avoid `if let ... && let ...` chains and keep compatibility conservative. I also noticed there are older let-chains already in `hardware.rs`; if you want a strict MSRV floor, I can follow up by adding an explicit `rust-version` in Cargo.toml.

llmfit-core/src/hardware.rs
MahmoudAdelbghany: Fixed in bee8482. Vulkan-discovered GPUs are now tagged with `GpuBackend::Vulkan` directly instead of going through `infer_gpu_backend`.

llmfit-core/src/hardware.rs
MahmoudAdelbghany: Fixed in bee8482. I gated `/proc/cpuinfo` and `getprop` CPU fallback paths behind `#[cfg(target_os = "linux")]`.

llmfit-core/src/hardware.rs
```diff
@@ -1023,6 +1029,135 @@ impl SystemSpecs {
...
+            }
+        }
+
+        false
+    }
+
+    /// Detect GPUs via Vulkan. This is especially useful on Android/Termux,
+    /// where vendor-specific Linux utilities may be unavailable.
+    fn detect_vulkan_gpu_info() -> Vec<GpuInfo> {
+        if !Self::has_command("vulkaninfo") {
```
Nikitf777, line 1058: Why not use a direct Vulkan API call (e.g. via the `vulkano` crate)? This would make the algorithm much simpler, faster and more portable (because it wouldn't depend on the `vulkaninfo` utility).

PR #203, opened 2026-03-11
fix: correct memory estimates for hybrid attention+SSM architectures

Pipboyguy:
Hybrid architectures like Qwen3.5 only use attention in a fraction of layers (8/32), with the rest being Mamba2 SSM needing no KV cache. Both the memory and speed estimators assumed pure attention, overshooting memory by up to 4x and undershooting speed by ~40%.

Adds `LlmModel::kv_cache_ratio()` method that detects hybrid architectures by name and adjusts both `estimate_memory_gb()` and `estimate_tps()`.

| Model | Before | After | Measured |
|-------|--------|-------|----------|
| Qwen3.5-9B Q8_0 / RTX 4080 memory | Doesn't fit | 15.7 GB | 8.5 GB (8k ctx) |
| Qwen3.5-9B Q8_0 / RTX 4080 speed | 40.9 t/s | 57.6 t/s | 67 t/s |
| Qwen3.5-9B best quant / 16GB | Q3_K_M | Q8_0 | Q8_0 |


llmfit-core/src/models.rs
```diff
@@ -278,6 +278,20 @@ impl LlmModel {
         quant_bpp(&self.quantization)
     }
 
+    /// Fraction of layers that use attention (and thus need KV cache).
+    /// Hybrid architectures only use attention in a subset of layers;
+    /// the rest are SSM (Mamba) with constant-size state, no KV cache.
+    pub fn kv_cache_ratio(&self) -> f64 {
+        let name = self.name.to_lowercase();
```
AlexsJones: name matching feels fragile — could the scraper pull this from the model config instead (e.g. architecture type or num_attention_layers)? would future-proof it for new hybrid models.

llmfit-core/src/models.rs
```diff
@@ -278,6 +278,20 @@ impl LlmModel {
...
+        if name.contains("qwen3.5") {
+            0.25 // 8/32 layers are attention, rest Mamba2
```
AlexsJones: is 0.125 validated? qwen3.5 has measured data in the PR description but jamba doesn't — worth a quick sanity check.

llmfit-core/src/models.rs
```diff
@@ -278,6 +278,20 @@ impl LlmModel {
...
+        } else if name.contains("jamba") {
+            0.125
+        } else {
+            1.0
+        }
+    }
+
     /// Parameter count in billions, extracted from parameters_raw or parameter_count.
```
AlexsJones, line 302: estimate is 15.7 GB vs 8.5 GB measured — still ~2x off. ssm layers probably have way less per-layer state than attention, so model weight memory might need adjusting too, not just kv cache.

llmfit-core/src/fit.rs
```diff
@@ -751,9 +751,11 @@ fn estimate_tps(
             let bytes_per_param = models::quant_bytes_per_param(quant);
             let model_gb = params * bytes_per_param;
 
-            // Efficiency factor — captures overhead not in the simple
-            // bandwidth / model-size formula.
-            let efficiency = 0.55;
+            // Efficiency: 0.55 baseline for pure-attention models.
```
AlexsJones: nice formula, makes sense physically. the 0.3 constant could use a brief comment on where it comes from.

llmfit-core/src/models.rs
Pipboyguy: Good call. Replaced the name matching with a data-driven `attention_ratio` field in the JSON. the scraper now computes it from config.json (`layer_types` array, `attn_layer_period`, or `full_attention_interval`). Rust side just reads `self.attention_ratio.unwrap_or(1.0)`. New hybrid models get correct ratios automatically on next scrape, no code changes needed.

llmfit-core/src/models.rs
Pipboyguy: Jamba-v0.1 config has `attn_layer_period: 8`, `num_hidden_layers: 32` → 4/32 = 0.125, checks out. Also tested with Jamba-Reasoning-3B which has `attn_layer_period: 14` → 1/14 ≈ 0.071 loaded it in llama.cpp on an RTX 4080 and the KV cache was 2.1 MiB (vs ~30+ MiB for a pure-attention 3B), confirming the ratio is correct. Since the scraper now computes this from config.json, each Jamba variant gets its own accurate ratio.

llmfit-core/src/models.rs
Pipboyguy: You were right — the gap was from embeddings, not SSM layer weights. llama.cpp keeps `token_embd` on CPU for CUDA builds, so our GPU estimate was including ~1 GB+ of weights that never hit VRAM.

Fixed by pulling `vocab_size` and `hidden_size` from HF config into the JSON (435 models patched), then subtracting `vocab_size * hidden_size * bpp` from the GPU weight estimate in `estimate_memory_gb()`.

Empirical results after the fix:
- Qwen3.5-9B Q8_0: predicted 8.96 GB, measured 8.31 GB (+7.8%)
- Jamba-Reasoning-3B Q4_K_M: predicted 2.27 GB, measured 2.11 GB (+7.4%)

The remaining ~7% is from some tensors (norms, biases, Mamba state matrices) stored at f32 regardless of quant level — not worth modelling IMO.

llmfit-core/src/fit.rs
Pipboyguy: Added a comment explaining the derivation :)

PR #204, opened 2026-03-11
fix: subtract CPU-resident embeddings from VRAM estimate

Pipboyguy:
In llama.cpp, `token_embd` is hardcoded to CPU even with `-ngl 99`. For large-vocabulary models this is significant — Qwen3.5-9B loses ~1 GB to CPU (11% of model weights).

Adds `cpu_embedding_gb()` that estimates the CPU-resident portion from known vocabulary sizes. Skips models <3B which use tied weights.

Validated against llama-server buffer reports on RTX 4080:

| Model | Predicted CPU | Measured CPU | Error |
|-------|--------------|-------------|-------|
| Qwen3.5-9B Q8_0 | 1.07 GB | 1.03 GiB | 6% |
| Qwen3.5-9B Q6_K | 0.81 GB | 0.78 GiB | 5% |

1 file changed, 37 insertions, 1 deletion.

### DIFF
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -278,6 +278,42 @@ impl LlmModel {
         quant_bpp(&self.quantization)
     }
 
+    /// Estimated size (GB) of CPU-resident token embeddings.
+    /// In llama.cpp, token_embd always stays on CPU even with -ngl 99.
+    /// Small models (<3B) typically use tied weights where the embedding
+    /// doubles as the output head on GPU, so we don't subtract those.
+    fn cpu_embedding_gb(&self, quant: &str) -> f64 {
+        if self.params_b() < 3.0 {
+            return 0.0;
+        }
+        let bpp = quant_bpp(quant);
+        let name = self.name.to_lowercase();
+        let vocab: f64 = if name.contains("qwen3.5") {
+            248320.0
+        } else if name.contains("gemma") {
+            256000.0
+        } else if name.contains("qwen") {
+            152064.0
+        } else if name.contains("llama-3") || name.contains("llama3") || name.contains("llama-4") {
+            128256.0
+        } else if name.contains("deepseek") {
+            102400.0
+        } else {
+            32000.0
+        };
+        let params = self.params_b();
+        let embed_dim: f64 = if params > 50.0 {
+            8192.0
+        } else if params > 14.0 {
+            5120.0
+        } else if params > 3.0 {
+            4096.0
+        } else {
+            2048.0
+        };
+        (vocab * embed_dim * bpp) / 1e9
+    }
+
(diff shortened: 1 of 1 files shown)


llmfit-core/src/models.rs
```diff
@@ -278,6 +278,42 @@ impl LlmModel {
...
 
+    /// Estimated size (GB) of CPU-resident token embeddings.
+    /// In llama.cpp, token_embd always stays on CPU even with -ngl 99.
+    /// Small models (<3B) typically use tied weights where the embedding
+    /// doubles as the output head on GPU, so we don't subtract those.
+    fn cpu_embedding_gb(&self, quant: &str) -> f64 {
+        if self.params_b() < 3.0 {
+            return 0.0;
+        }
+        let bpp = quant_bpp(quant);
```
AlexsJones, line 289: same concern as #203 — hardcoded vocab sizes by name will need updating for every new model family. the scraper already hits the HF config endpoint, could it pull `vocab_size` and `hidden_size` into hf_models.json? would eliminate both this lookup and the embed_dim heuristic below.

llmfit-core/src/models.rs
```diff
@@ -278,6 +278,42 @@ impl LlmModel {
...
+        };
+        let params = self.params_b();
+        let embed_dim: f64 = if params > 50.0 {
+            8192.0
+        } else if params > 14.0 {
+            5120.0
+        } else if params > 3.0 {
+            4096.0
+        } else {
+            2048.0
```
AlexsJones, line 312: embed_dim guessed from param count feels loose — a 14B model with 5120 vs 4096 embed_dim changes the result by 25%. if vocab_size comes from the scraper, hidden_size should too.

llmfit-core/src/models.rs
```diff
@@ -278,6 +278,42 @@ impl LlmModel {
...
+        };
+        (vocab * embed_dim * bpp) / 1e9
+    }
+
```
AlexsJones, line 316: the  floor is a nice safety net but it means for some models the embedding subtraction silently gets clamped. might be worth logging or flagging when the floor kicks in — could indicate the estimate is way off.

llmfit-core/src/models.rs
```diff
@@ -278,6 +278,42 @@ impl LlmModel {
         quant_bpp(&self.quantization)
     }
 
+    /// Estimated size (GB) of CPU-resident token embeddings.
+    /// In llama.cpp, token_embd always stays on CPU even with -ngl 99.
+    /// Small models (<3B) typically use tied weights where the embedding
```
AlexsJones, line 283: this only matters for llama.cpp right? ollama and mlx handle embeddings differently. might want to gate this on the runtime provider, or at least note it in the doc comment.

Issue #172, opened 2026-03-06
'llmfit info' does not return with exact match


davemedvitz:
For `llmfit info`, I expect that an exact match would return the info for the matched model.  Instead, it always does a search. 

In v0.6.3, I have the following 

```
# llmfit info Qwen/Qwen3-Coder-Next

Multiple models found. Please be more specific:
  - Qwen/Qwen3-Coder-Next-FP8
  - Qwen/Qwen3-Coder-Next
```
where I was wanting the info for `Qwen/Qwen3-Coder-Next`

three-foxes-in-a-trenchcoat:
Good catch! The `info` command should prioritize an exact match when one exists, rather than always falling back to fuzzy search. Right now it seems like the lookup just does a substring/contains check and returns all matches without checking for an exact hit first.

A simple fix would be: check if any result matches the query exactly (case-sensitive), and if so, return just that one. Only fall back to the "Multiple models found" disambiguation when there's no exact match.

I can put together a PR for this if you'd like — should be a straightforward change in the info subcommand's model resolution logic.

PR #206, opened 2026-03-12
fix: prefer exact matches in info selection

haosenwang1018:
$## Summary\n- make `llmfit info <model>` reuse exact-match-first selector logic instead of raw fuzzy search results\n- preserve substring fallback when there is no exact match\n- add unit coverage for exact-match preference on plain model lists\n\n## Testing\n- `cargo fmt --check`\n- `cargo test -p llmfit selector_prefers_exact_match -- --nocapture`\n- `cargo test -p llmfit generic_selector_prefers_exact_match_for_models -- --nocapture`\n\nCloses #172

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -659,26 +659,31 @@ fn fit_matches_filter(fit: &ModelFit, filter: FitArg) -> bool {
     }
 }
 
-fn find_fit_index_by_selector(fits: &[ModelFit], selector: &str) -> Result<usize, String> {
+fn find_name_index_by_selector<T>(
+    items: &[T],
+    selector: &str,
+    get_name: impl Fn(&T) -> &str,
+) -> Result<usize, String> {
     let needle = selector.trim().to_lowercase();
     if needle.is_empty() {
         return Err("Model selector cannot be empty".to_string());
     }
 
-    if let Some((idx, _)) = fits
+    if let Some((idx, _)) = items
         .iter()
         .enumerate()
-        .find(|(_, f)| f.model.name.to_lowercase() == needle)
+        .find(|(_, item)| get_name(item).to_lowercase() == needle)
     {
         return Ok(idx);
     }
 
-    let matches: Vec<(usize, &str)> = fits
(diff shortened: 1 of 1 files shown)


llmfit-tui/src/main.rs
```diff
@@ -659,26 +659,31 @@ fn fit_matches_filter(fit: &ModelFit, filter: FitArg) -> bool {
...
+    if let Some((idx, _)) = items
         .iter()
         .enumerate()
-        .find(|(_, f)| f.model.name.to_lowercase() == needle)
+        .find(|(_, item)| get_name(item).to_lowercase() == needle)
     {
         return Ok(idx);
     }
 
-    let matches: Vec<(usize, &str)> = fits
```
AlexsJones, line 676: the switch from `&str` to `String` here is a litle unfortunate allocation-wise but totally fine for this list size

llmfit-tui/src/main.rs
```diff
@@ -1552,4 +1556,55 @@ mod tests {
...
+                provider: "Qwen".to_string(),
+                parameter_count: "7B".to_string(),
+                parameters_raw: None,
+                min_ram_gb: 4.0,
+                recommended_ram_gb: 8.0,
+                min_vram_gb: Some(4.0),
+                quantization: "Q4_K_M".to_string(),
+                context_length: 8192,
+                use_case: "general".to_string(),
+                is_moe: false,
```
AlexsJones, line 1574: maybe worth a little test helper to build these model structs so future tests dont need all the boilerplate, but no biggie for now

llmfit-tui/src/main.rs
```diff
@@ -1346,22 +1355,17 @@ fn main() {
...
+                        println!("\n{}", err);
+                        return;
                     }
-                    return;
-                }
+                };
 
-                let fit = ModelFit::analyze_with_context_limit(results[0], &specs, context_limit);
+                let fit = ModelFit::analyze_with_context_limit(&models[idx], &specs, context_limit);
                 if cli.json {
```
AlexsJones, line 1369: solid approach using get_all_models here, keeps info consistant with the rest of the selector behaviour

PR #219, opened 2026-03-12
Use a direct API call for Vulkan GPU detection

Nikitf777:
This PR makes llmfit use the `vulkano` crate for Vulkan GPU detection instead of the `vulkaninfo` utility. This makes the code much simpler, faster and more portable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "as-raw-xcb-connection"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "175571dd1d178ced59193a6fc02dde1b972eb0bc56c892cde9beeceac5bf0f6b"
+
+[[package]]
+name = "ash"
+version = "0.38.0+1.3.281"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bb44936d800fea8f016d7f2311c6a4f97aebd5dc86f09906139ec848cf3a46f"
+dependencies = [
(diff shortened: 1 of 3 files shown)


Nikitf777:
But I think most of these dependencies aren't even included in the final binary. It only affects the build process.

Nikitf777:
Ash also requires to use unsafe blocks multiple times for this task. I could wrap them, but at this point I'm just reimplementing the vulkano library.

Here's the simpliest ash solution:
```Rust
fn detect_vulkan_gpu_info() -> Vec<GpuInfo> {
    let entry = unsafe {
        match ash::Entry::load() {
            Ok(entry) => entry,
            Err(_) => return Vec::new(),
        }
    };

    let app_info = vk::ApplicationInfo::default();
    let create_info = vk::InstanceCreateInfo::default().application_info(&app_info);

    let instance = unsafe {
        match entry.create_instance(&create_info, None) {
            Ok(instance) => instance,
            Err(_) => return Vec::new(),
        }
    };

    let physical_devices = unsafe {
        match instance.enumerate_physical_devices() {
            Ok(devices) => devices,
            Err(_) => return Vec::new(),
        }
    };

    if physical_devices.is_empty() {
        return Vec::new();
    }

    let mut grouped: BTreeMap<String, u32> = BTreeMap::new();
    for (_, device) in physical_devices.iter().enumerate() {
        let properties = unsafe { instance.get_physical_device_properties(*device) };

        if matches!(properties.device_type, vk::PhysicalDeviceType::CPU) {
            continue;
        }

        let name = unsafe { CStr::from_ptr(properties.device_name.as_ptr()) }
            .to_string_lossy()
            .to_string();

        *grouped.entry(name).or_insert(0) += 1;
    }

    grouped
        .into_iter()
        .map(|(name, count)| GpuInfo {
            backend: GpuBackend::Vulkan,
            count,
            name,
            unified_memory: false,
            vram_gb: None,
        })
        .collect()
}
```

PR #220, opened 2026-03-12
feat: add TUI runtime filter popup

### LINKED ISSUE #197: [Feature request] Filter by backend, or suggestion about new UI mode
llmfit now show lots of models by default, while we can filter out with Providers, Use Case, Caps, Fit and Avail, it is easier for pulling new models if we can filter by backends/runtime:
- List models supporting Ollama, e.g., to pull the corresponding ollama flavor model, instead of try pulling and find out "No compatible provider available for this model".

In this case, even if llmfit supports more backends/runtime, the logic is still very clear. And also, #113 can have more simple solution, with different default filter options on different platform.

(BTW, why the message is "no compatible **provider**", while the Provider filter showed the company name?)

-------------
Also, perhaps the Mode column should also have a filter.

Consider that the model list is a kind of structured table, would it be more intuitive if the UI is a kind of "interactive table" as in excel, that we can filter, sort at each (or a subset of) column? For example:
- Provider
- Mode
- Use Case
- Caps
- Fit
- Backend

Can be used for filtering and sorting; and
- Params
- Score
- tok/s
- Ctx
- Mem
- Date

Can be used for sorting and thresholding.

Then one can do this in the interactive table:
List all models in date order, that they have:
- scores > 80,
- Ctx > 128k,
- Fit good or perfect,
- params < 70B,
- uses Ollama & llama.cpp as backend,
- for general purpose,
- and is not installed.

haosenwang1018:
## Summary
- add a TUI runtime/backend filter popup alongside the existing Provider / Use Case / Capability filters
- wire the new filter into `apply_filters()` using each fit's existing `runtime` field
- expose the filter in the search/filter bar and status/help text with the `B` key

## Why
Closes #197. The API already supported runtime filtering, but the interactive TUI did not expose a corresponding backend/runtime filter.

## Testing
- `cargo fmt --check`
- `cargo test -p llmfit --no-run`

### DIFF
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -1,4 +1,4 @@
-use llmfit_core::fit::{FitLevel, ModelFit, SortColumn, backend_compatible};
+use llmfit_core::fit::{FitLevel, InferenceRuntime, ModelFit, SortColumn, backend_compatible};
 use llmfit_core::hardware::SystemSpecs;
 use llmfit_core::models::{Capability, ModelDatabase, UseCase};
 use llmfit_core::plan::{PlanEstimate, PlanRequest, estimate_model_plan};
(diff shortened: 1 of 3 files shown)


AlexsJones:
Do you have a screenshot of it?

PR #225, opened 2026-03-13
docs: clarify containerized llama.cpp detection

### LINKED ISSUE #207: Docker version of llama.cpp is not detected
Hi, i searched for it but could not find any issue describing exactly my problem.

When running ollama with docker compose -> llmfit detects ollama
When running llama.cpp with docker compose -> llmfit cannot detect llama.cpp

Ports are default:
- Ollama -> 11434
- llama.cpp -> 8080 

If you need anything else, just let me know.
Thanks!

haosenwang1018:
## Summary
- document that `llmfit` detects `llama.cpp` via local host binaries on `PATH`
- clarify that, unlike Ollama, containerized `llama.cpp` runtimes are not auto-detected over an HTTP endpoint today
- add practical workarounds for Docker/Podman users

## Why
Fixes #207. Users can run Ollama in Docker and still get detection because Ollama has an API endpoint that `llmfit` checks, but `llama.cpp` detection currently relies on local `llama-cli` / `llama-server` binaries. Making that distinction explicit avoids the impression that a Dockerized `llama.cpp` server on port 8080 should already be auto-detected.

## Validation
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,20 @@ How it works:
 - downloads GGUF files into the local llama.cpp model cache
 - marks models installed when matching GGUF files are present locally
 
+### Containerized llama.cpp
+
+`llmfit` currently detects `llama.cpp` by looking for local `llama-cli` / `llama-server` binaries in the same host environment where `llmfit` is running. Unlike Ollama, it does **not** probe a remote or containerized `llama.cpp` HTTP endpoint for runtime detection.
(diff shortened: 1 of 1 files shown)


segganeadel:
On windows installting `which` via scoop enabled me to detect the `llama.cpp` runtime

PR #226, opened 2026-03-13
docs: clarify runtime vs model provider terminology

### LINKED ISSUE #198: No compatible provider available for this model
I am confused by terminology. `llmfit` provider is a model author/site (<kbd>P</kbd> gives a list of 140 providers like Google and HuggingFace), and it also seems that provider is the inference engine (https://github.com/AlexsJones/llmfit/blob/main/llmfit-core/src/providers.rs).

So which provider is not available? Is it download site that is down? Or is it `llama.cpp` that can not run this model?

Image

What is wrong here?

haosenwang1018:
## Summary
- document the distinction between model providers (catalog/authors) and runtime/download providers (Ollama, MLX, llama.cpp)
- update the TUI pull status text to say `runtime/download source` instead of the overloaded `provider`

## Why
Fixes #198. The current UI and docs use "provider" for two different concepts, which makes errors like "No compatible provider available for this model" sound like a registry/model-author problem when the real issue is usually runtime compatibility on the local machine.

## Validation
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,6 +493,15 @@ If Ollama is not running, Ollama-specific operations are skipped; the TUI still 
 
 llmfit integrates with [llama.cpp](https://github.com/ggml-org/llama.cpp) as a runtime/download provider in both TUI and CLI.
 
+### Terminology: model provider vs runtime provider
+
+llmfit uses the word "provider" in two different places:
+
+- **Model provider**: the model author or catalog source shown in filters (for example Hugging Face, Google, Microsoft)
+- **Runtime / download provider**: the engine that can actually run or fetch the selected model on your machine (currently Ollama, MLX, or llama.cpp)
+
+So when the TUI says a model has no compatible runtime/download source available, it means llmfit could not find a supported engine for that model on your current machine — not that the model author or registry is missing.
+
 Requirements:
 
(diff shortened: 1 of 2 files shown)


llmfit-tui/src/tui_app.rs
abitrolly: The message is still not actionable. Is it possible to explicitly say where is the problem - no runtime or no model? And what to do?

README.md
```diff
@@ -538,6 +493,15 @@ If Ollama is not running, Ollama-specific operations are skipped; the TUI still
...
 
+### Terminology: model provider vs runtime provider
+
+llmfit uses the word "provider" in two different places:
+
+- **Model provider**: the model author or catalog source shown in filters (for example Hugging Face, Google, Microsoft)
+- **Runtime / download provider**: the engine that can actually run or fetch the selected model on your machine (currently Ollama, MLX, or llama.cpp)
+
+So when the TUI says a model has no compatible runtime/download source available, it means llmfit could not find a supported engine for that model on your current machine — not that the model author or registry is missing.
+
```
abitrolly, line 504: As mentioned in https://github.com/AlexsJones/llmfit/issues/198#issuecomment-4046083650 the better name for "model provider" is "vendor".

PR #232, opened 2026-03-13
Use ash for Vulkan GPU detection

Nikitf777:
Same as #219 but using `ash` instead of `vulkano`.
Since ash provides low-level Vulkan bindings, this new implementation now uses unsafe blocks multiple times. I'm not sure if they need to be wrapped, because it can make the implementation too complex for such a simple task.


Nikitf777:
Wtf happened here? This PR definitely did not look like this at the beginning.

Issue #175, opened 2026-03-07
GPU not detected on Android (Termux/PRoot) — e.g. Samsung S23 Ultra, Adreno


actuallyrizzn:
## Environment

- **Device:** Samsung S23 Ultra (Qualcomm Snapdragon 8 Gen 2, Adreno 740 GPU)
- **Stack:** Android → Termux → PRoot Ubuntu → running llmfit
- **llmfit:** v0.6.3 (prebuilt `aarch64-unknown-linux-musl` binary)

## Issue

llmfit reports **no GPU detected** on this device. The system has a capable GPU (Adreno 740), but the tool only sees:

- CPU: Cortex-A510 (8 cores)
- Total RAM: 6.91 GB
- Backend: CPU (ARM)
- **GPU: Not detected**

So all model recommendations are CPU-only (“Marginal” fit, no GPU-based suggestions).

## Why this happens

llmfit’s GPU detection is built for **desktop/server** environments:

- **NVIDIA:** `nvidia-smi` (and Linux sysfs fallback)
- **AMD:** `rocm-smi` or Linux DRM/sysfs
- **Intel:** sysfs / `lspci`
- **Apple:** `system_profiler` (macOS)
- **Ascend:** `npu-smi`

On **Android (Termux + PRoot)**:

- None of these interfaces are available. There is no `nvidia-smi`, no `rocm-smi`, and no standard Linux DRM/sysfs GPU nodes visible to the proot environment.
- The GPU is driven by the Android graphics stack (Vulkan/OpenGL ES via the Android HAL), not by the usual Linux GPU APIs that llmfit (or any generic Linux binary) can query.
- From inside PRoot, the system therefore looks like a headless Linux box with no GPU, even though the hardware has one.

So the “no GPU” result is expected with the **current** detection logic; the gap is that **Android / SoC GPUs with unified memory** are not considered.

## Workaround (current behavior)

Using the **`--memory`** override lets users get GPU-style recommendations without detection:

```bash
llmfit --memory=8G fit -n 20
llmfit --memory=8G recommend --json --limit 10
```

That’s a manual workaround, not automatic detection.

## Possible directions (not proposed as full solutions)

Ideas that might help, without claiming they are complete or correct:

1. **Android / unified-memory heuristic**  
[...]

PR #233, opened 2026-03-13
docs: clarify android memory override caveat

haosenwang1018:
## Summary
- clarify the Android / Termux `--memory` workaround in the README
- add an `llmfit info ... --memory=8G` example alongside `fit` and `recommend`
- explicitly state that `--memory` does not mean a mobile GPU runtime has been detected

## Why
Issue #175 is mostly a documentation / expectation-management gap today. The README already mentions manual memory override, but it does not clearly distinguish between "use this for rough sizing" and "llmfit has actually detected a usable Android GPU backend". This PR makes that caveat explicit.

## Validation
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,9 +622,12 @@ If you still want GPU-style recommendations on a unified-memory phone or tablet,
 ```sh
 llmfit --memory=8G fit -n 20
 llmfit recommend --json --memory=8G --limit 10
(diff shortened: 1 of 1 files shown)


AlexsJones:
Please hold off more Prs
I can’t review them this quickly

PR #251, opened 2026-03-16
Fix exact model matches for llmfit info

### LINKED ISSUE #172: 'llmfit info' does not return with exact match

jasperan:
Closes #172

## Summary
- route `llmfit info` through the exact-first selector used by planning
- merge duplicate exact-name catalog entries on load so collisions do not break exact lookup
- add regression coverage for exact-vs-partial selection and duplicate catalog merges

## Verification
- `cargo test`
- `cargo run --quiet -- info Qwen/Qwen3-Coder-Next`

### DIFF
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
 
 /// Quantization levels ordered from best quality to most compressed.
@@ -378,7 +380,7 @@ impl LlmModel {
 
 /// Intermediate struct matching the JSON schema from the scraper.
 /// Extra fields are ignored when mapping to LlmModel.
-#[derive(Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
 struct HfModelEntry {
     name: String,
     provider: String,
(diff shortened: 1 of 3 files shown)


jasperan:
just noticed there's already another PR that should fix it and Alex already reviewed, you can close this PR if needed

jasperan:
#172

AlexsJones:
Hey, thanks for putting this together! Just a heads up — we ended up merging #206 which covers the exact-match fix for `info`, so there'll be some overlap there.

That said, the duplicate catalog entry dedup logic you've got here is really interesting and solves a real problem. Would you be up for splitting that part out into its own PR? It deserves its own discussion and review rather than being bundled with the info fix. The merge logic and test coverage you wrote for it look solid.

jasperan:
of course @AlexsJones ! I will do that right away.

jasperan:
Split the duplicate catalog entry work into #254, which is based on current main and only contains the catalog dedupe + regression coverage.

PR #259, opened 2026-03-17
Improve HF API data extraction

julien-c:
## Summary
- Fix parameter count: use `sum` instead of `max` for safetensors per-dtype params
- Fall back to `gguf.total` when safetensors metadata is missing
- Add `license` field from `cardData`
- Replace `text2text-generation` with `image-text-to-text` in discovery pipelines
- Detect `tool_use` capability from chat_template instead of hardcoded model family heuristics

## Test plan
- [ ] Run scraper against a set of known models and verify parameter counts
- [ ] Verify license field is populated correctly for models with string and array licenses
- [ ] Check tool_use detection against models with known chat_template support

### DIFF
--- a/scripts/scrape_hf_models.py
+++ b/scripts/scrape_hf_models.py
@@ -805,7 +805,7 @@ def enrich_gguf_sources(models: list[dict]) -> int:
 # ---------------------------------------------------------------------------
 
 # Pipeline tags to search for discoverable models
-DISCOVER_PIPELINES = ["text-generation", "image-text-to-text"]
+DISCOVER_PIPELINES = ["text-generation", "text2text-generation", "image-text-to-text"]
 
 # Orgs to skip — these publish many fine-tunes that clutter the list
 SKIP_ORGS = {


AlexsJones (commented):
Thanks for the PR! A few questions inline.

scripts/scrape_hf_models.py
```diff
@@ -619,7 +617,11 @@ def scrape_model(repo_id: str) -> dict | None:
     if not total_params:
         params_by_dtype = safetensors.get("parameters", {})
         if params_by_dtype:
-            total_params = max(params_by_dtype.values())
+            total_params = sum(params_by_dtype.values())
```
AlexsJones, line 620: Switching from `max` to `sum` here — wouldn't this double-count params for models that store the same weights in multiple dtypes? Would love to see some example outputs if you've tested this.

scripts/scrape_hf_models.py
```diff
@@ -619,7 +617,11 @@ def scrape_model(repo_id: str) -> dict | None:
...
+
+    if not total_params:
+        gguf = info.get("gguf", {})
+        total_params = gguf.get("total")
```
AlexsJones, line 624: Nice, the GGUF fallback makes total sense.

scripts/scrape_hf_models.py
```diff
@@ -657,12 +659,13 @@ def scrape_model(repo_id: str) -> dict | None:
...
+        "capabilities": infer_capabilities(repo_id, pipeline_tag, use_case_str, config),
         "pipeline_tag": pipeline_tag or "unknown",
         "architecture": architecture,
         "hf_downloads": info.get("downloads", 0),
         "hf_likes": info.get("likes", 0),
         "release_date": (info.get("createdAt") or "")[:10] or None,
+        "license": l[0] if isinstance(l := (info.get("cardData") or {}).get("license"), list) else l,
     }
 
     # Add MoE fields if detected
```
AlexsJones, line 674: This walrus one-liner is pretty dense — would you be open to pulling the license lookup into a couple lines above?

scripts/scrape_hf_models.py
```diff
@@ -799,7 +802,7 @@ def enrich_gguf_sources(models: list[dict]) -> int:
 # ---------------------------------------------------------------------------
 
 # Pipeline tags to search for discoverable models
-DISCOVER_PIPELINES = ["text-generation", "text2text-generation"]
+DISCOVER_PIPELINES = ["text-generation", "image-text-to-text"]
```
AlexsJones: This replaces `text2text-generation` rather than adding alongside it — does that mean we're intentionally dropping seq2seq models like T5 from disocvery?

scripts/scrape_hf_models.py
```diff
@@ -524,17 +524,15 @@ def infer_capabilities(repo_id: str, pipeline_tag: str | None, use_case: str) ->
...
-        "tool" in uc
-        or "function call" in uc
-        or "qwen3" in rid
-        or "qwen2.5" in rid
-        or "command-r" in rid
-        or ("llama-3" in rid and "instruct" in rid)
-        or ("mistral" in rid and "instruct" in rid)
-        or "hermes" in rid
-    ):
+    # Tool use — check chat_template for "tools" mention
```
AlexsJones, line 527: Love the idea of checking `chat_template` instead of hardcoding model families. Quick question though — does the inline `info["config"]` reliably have `chat_template`, or would we need the full config from `fetch_config_json()` to catch everything?

scripts/scrape_hf_models.py
julien-c: - that's what we do on our side to compute total parameters (note, the total should always be present on our side whenever there's a safetensors property anyways): 

```ts
this.safetensors = {
	parameters,
	total: Utils.sum(Object.values(parameters)),
};
```

- in 'regular' safetensors file all tensors are only stored once (but different tensors can be stored in different dtypes)

https://github.com/huggingface/huggingface.js/blob/92724130b4bd686857fee44dfad032ce0c3ecca5/packages/hub/src/lib/parse-safetensors-metadata.ts#L476-L502

julien-c: Good catch — but `sum` is actually correct here. As @julien-c confirmed, in regular safetensors files each tensor is stored exactly once; different tensors just happen to use different dtypes. HF Hub itself computes `total` the same way (`sum(Object.values(parameters))`). The `safetensors.total` field (checked first at L616) already provides this, so the `sum` fallback is only hit when `total` is missing.

scripts/scrape_hf_models.py
julien-c: Fair point — refactored into explicit lines above the dict:

```python
license_value = (info.get("cardData") or {}).get("license")
if isinstance(license_value, list):
    license_value = license_value[0] if license_value else None
```

scripts/scrape_hf_models.py
julien-c: Yes, intentional. `text2text-generation` models (T5, BART, etc.) are seq2seq encoder-decoder architectures — they're not really the decoder-only LLMs this tool is designed for. `image-text-to-text` replaces it to capture VLMs (Llava, Qwen-VL, etc.) which are much more relevant. Any seq2seq model already in the DB via manual addition would still be kept, this only affects auto-discovery.

scripts/scrape_hf_models.py
julien-c: Good question — the inline `info["config"]` from the HF API is a subset and doesn't reliably include `chat_template`. Updated to pass `full_config or config` so we prefer the full `config.json` fetched via `fetch_config_json()` (which has `tokenizer_config`/`chat_template`) and only fall back to the inline config when the full fetch fails.

julien-c: i think so bc that's the code we use in our internal codebase

scripts/scrape_hf_models.py
julien-c: `text2text-generation` is kinda a legacy pipeline at this point (no new model with this pipeline in transformers AFAIK), but i've added back in 24295858737f67feb8b23d1176ab80233ef060e7

scripts/scrape_hf_models.py
julien-c: dbd3c290af00cdbca93076229d40bbae9cc1e891

PR #263, opened 2026-03-18
dockerhub-search for Docker Hub ai/ namespace

ericcurtin:
Adds `llmfit dockerhub-search [QUERY]` command that queries
hub.docker.com/v2/repositories/ai/, filters by name/description,
and displays results sorted by pull count. Mirrors the behavior of
`model-cli search --source dockerhub` in model-runner.

PR #268, opened 2026-03-19
dockerhub-search for Docker Hub ai/ namespace

ericcurtin:
Adds `llmfit dockerhub-search [QUERY]` command that queries
hub.docker.com/v2/repositories/ai/, filters by name/description,
and displays results sorted by pull count. Mirrors the behavior of
`model-cli search --source dockerhub` in model-runner.

This is a clean re-cut of #263 against the latest main.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -2308,6 +2308,167 @@ pub fn ollama_pull_tag(hf_name: &str) -> Option<String> {
     lookup_ollama_tag(hf_name).map(|s| s.to_string())
 }
 
+// ---------------------------------------------------------------------------
+// Docker Hub model search
+// ---------------------------------------------------------------------------
+
+/// A model found on Docker Hub in the `ai/` namespace.
+#[derive(Debug, Clone)]
+pub struct DockerHubModel {
+    /// Full name, e.g. `ai/llama3.2`.
+    pub name: String,
+    /// Short description from the Hub listing.
+    pub description: String,
+    /// Total pull count.
+    pub pulls: u64,
+    /// Star count.
+    pub stars: u64,
+    /// Inferred backend: `"llama.cpp"`, `"vllm"`, or `"llama.cpp, vllm"`.
+    pub backend: String,
+    /// ISO-8601 last-updated timestamp.
+    pub updated_at: String,
+}
+
+/// Search Docker Hub's `ai/` namespace for models matching `query`.
+///
+/// Pass an empty string to list all models.  Results are sorted by
+/// pull count (descending) and capped at `limit`.
+pub fn search_dockerhub(query: &str, limit: usize) -> Vec<DockerHubModel> {
+    let limit = if limit == 0 { 32 } else { limit };
+    let query_lower = query.to_lowercase();
+    let mut results: Vec<DockerHubModel> = Vec::new();
+    let mut next_url: Option<String> = None;
+
+    eprintln!("Searching Docker Hub...");
+
+    loop {
+        let url = match &next_url {
+            Some(u) => u.clone(),
+            None => format!(
+                "https://hub.docker.com/v2/repositories/ai/?page_size=100&ordering=pull_count"
+            ),
+        };
+
+        let Ok(resp) = ureq::get(&url)
+            .config()
+            .timeout_global(Some(std::time::Duration::from_secs(15)))
+            .build()
+            .call()
+        else {
+            break;
+        };
+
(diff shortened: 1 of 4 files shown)


llmfit-tui/src/display.rs
```diff
@@ -399,6 +400,46 @@ pub fn display_search_results(models: &[&LlmModel], query: &str) {
...
+            name: m.name.clone(),
+            provider: "Docker Hub".to_string(),
+            size: "-".to_string(),
+            score: "-".to_string(),
+            tps: "-".to_string(),
+            quant: "-".to_string(),
+            runtime: "-".to_string(),
+            mode: "-".to_string(),
+            mem_use: "-".to_string(),
+            context: "-".to_string(),
```
AlexsJones: This reuses `ModelRow` but fills most fields with `"-"` (status, size, score, tps, quant, runtime, mode, mem_use, context). You're already fetching pulls, stars, backend, and updated_at in `DockerHubModel` but then throwing them away here. A dedicated row struct with just the relevant columns would make the output way more useful.

llmfit-core/src/providers.rs
```diff
@@ -2308,6 +2308,163 @@ pub fn ollama_pull_tag(hf_name: &str) -> Option<String> {
...
+        || combined.contains("gguf")
+        || combined.contains("llama-cpp");
+    match (has_vllm, has_llama) {
+        (true, true) => "llama.cpp, vllm".to_string(),
+        (true, false) => "vllm".to_string(),
+        (false, _) => "llama.cpp".to_string(),
+    }
+}
+
+fn truncate_str(s: &str, max: usize) -> String {
```
AlexsJones, line 2463: `&s[..max-3]` indexes by byte offset, so if the cut lands in the middle of a multi-byte character this will panic. Using `.chars().take(n)` or checking for a char boundary would fix that.

llmfit-core/src/providers.rs
```diff
@@ -2308,6 +2308,163 @@ pub fn ollama_pull_tag(hf_name: &str) -> Option<String> {
...
+            Some(r) => r.clone(),
+            None => break,
+        };
+
+        for repo in repos {
+            let is_private = repo
+                .get("is_private")
+                .and_then(|v| v.as_bool())
+                .unwrap_or(false);
+            if is_private {
```
AlexsJones, line 2375: `r.clone()` copies the whole vec on each page. You can just iterate by reference instead.

llmfit-core/src/providers.rs
```diff
@@ -2308,6 +2308,163 @@ pub fn ollama_pull_tag(hf_name: &str) -> Option<String> {
     lookup_ollama_tag(hf_name).map(|s| s.to_string())
 }
 
+// ---------------------------------------------------------------------------
+// Docker Hub model search
+// ---------------------------------------------------------------------------
+
+/// A model found on Docker Hub in the `ai/` namespace.
```
AlexsJones, line 2315: Worth adding `#[derive(Debug, Clone)]` here so it's easier to use in tests and downstream code.

llmfit-core/src/providers.rs
```diff
@@ -2308,6 +2308,163 @@ pub fn ollama_pull_tag(hf_name: &str) -> Option<String> {
...
+/// Search Docker Hub's `ai/` namespace for models matching `query`.
+///
+/// Pass an empty string to list all models.  Results are sorted by
+/// pull count (descending) and capped at `limit`.
+pub fn search_dockerhub(query: &str, limit: usize) -> Vec<DockerHubModel> {
+    let limit = if limit == 0 { 32 } else { limit };
+    let query_lower = query.to_lowercase();
+    let mut results: Vec<DockerHubModel> = Vec::new();
+    let mut next_url: Option<String> = None;
+
```
AlexsJones, line 2341: With a 15s timeout the CLI can look frozen on slow connections. Even just a `"Searching Docker Hub..."` print before the request starts would go a long way for UX.

AlexsJones (commented):
Nice addition! Being able to search the Docker Hub ai/ namespace directly from the CLI is really handy. Left a few inline notes.

three-foxes-in-a-trenchcoat:
Cool feature — searching Docker Hub for AI models. The concept makes sense for finding containerized model runners. Just needs `cargo fmt` to pass CI. Otherwise looks solid.

AlexsJones:
I've decided I don't want to integrate with docker hub

PR #181, opened 2026-03-08
feat: Add DGX Spark cluster support, TP awareness, use-case/availability filters, and live benchmarking

michaeljabbour:
## Summary

This PR adds several features to llmfit from the llmfit-spark fork, all applied additively on top of current upstream/main. The diff is **+2,408 / -75** across 14 files — purely additive.

---

### 1. DGX Spark Cluster Support (`llmfit cluster`)
- New `cluster` subcommand with `init`, `status`, `remove` actions
- New `llmfit-core/src/cluster.rs` (542 lines): `ClusterSpecs` struct, Ray API discovery, TOML config persistence
- `--cluster` / `--no-cluster` global flags to override hardware detection with cluster resources
- Cluster mode replaces local GPU detection with aggregated multi-node VRAM for fit analysis
- TUI system bar shows cluster info when cluster mode is active
- vLLM selected as inference runtime in cluster mode; tensor-parallel notes added to fit output

### 2. Tensor Parallelism Awareness
- New `supports_tp(tp_size)` and `valid_tp_sizes()` methods on `LlmModel`
- `num_attention_heads` / `num_key_value_heads` fields added to `LlmModel` (optional, heuristic inference fallback)
- TP filter in TUI: cycle All → TP=2 → TP=3 → TP=4 with `x` key
- TP column in TUI table showing max valid TP degree
- TP compatibility line in detail view (green when cluster TP matches, yellow otherwise)
- TP penalty (0.9×) added to speed estimation formula

### 3. Use-Case Filter + Availability Filter
- Cycling use-case filter (`u` key): All → General → Coding → Reasoning → Chat → Multi → Embed
- Availability filter (`a` key): All → GGUF Avail → Installed (checks llmfit-core GGUF source metadata)
- Both filters compose with existing provider/fit/capability filters in `apply_filters()`

### 4. Year Sort Column
- `SortColumn::ReleaseYear` — sort by release year (newest first), secondary sort by score
- `Year` column in TUI table
- `s` key cycles now includes Year in the rotation

### 5. Live Inference Benchmarking (`llmfit bench`)
- New `llmfit-core/src/bench.rs` (610 lines): `BenchTarget` enum, `BenchResult`, provider-specific benchmark functions
- Supports Ollama, vLLM (OpenAI-compat), and MLX endpoints
- Auto-detection of running providers (`--provider auto`)
- `--all` flag benchmarks every discovered model across all running providers
- `--url` flag overrides the provider endpoint
[...]

PR #278, opened 2026-03-22
feat: add llmfit bench for live inference benchmarking

michaeljabbour:
## Summary

- New `llmfit bench` subcommand for real inference benchmarking
- Measures TPS, TTFT, and total latency against running providers
- Supports Ollama, vLLM (OpenAI-compat), and MLX endpoints
- Auto-detection of running providers (`--provider auto`)
- `--all` flag benchmarks every discovered model across all providers
- `--url` flag overrides the provider endpoint
- TTFT is `Option<f64>`: measured for Ollama (from `eval_duration`), `None` for non-streaming vLLM/MLX
- JSON output (`--json`) for tool/agent integration
- `VLLM_PORT` env var for configurable vLLM port (default 8000)

### Files changed (3)

| File | Change |
|------|--------|
| `llmfit-core/src/bench.rs` | New benchmarking module (730 lines) |
| `llmfit-core/src/lib.rs` | Module export (+1) |
| `llmfit-tui/src/main.rs` | Bench subcommand, run_bench(), helpers (+286) |

### Example usage

```bash
# Benchmark a specific model on Ollama
llmfit bench qwen3:8b --provider ollama --runs 5

# Auto-detect provider and model
llmfit bench qwen3:8b

# Benchmark all discovered models
llmfit bench --all

# JSON output for CI/agents
llmfit bench qwen3:8b --json

# Custom vLLM endpoint
llmfit bench --provider vllm --url http://gpu-server:8000
```

## Test plan

- [x] All existing tests pass (1 pre-existing failure unrelated to this PR)
- [x] 5 new unit tests for `BenchSummary::from_runs` (multiple runs, single, empty, min/max)
- [ ] Manual: start Ollama, run `llmfit bench <model>`, verify output
- [ ] Manual: run `llmfit bench --all` with multiple providers

> Split from #181 — see [closing comment](#181) for context on the split.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #279, opened 2026-03-22
feat: add remote GPU cluster support for multi-node model fitting

michaeljabbour:
## Summary

- Hardware-agnostic remote GPU cluster support (any vendor — NVIDIA, AMD, etc.)
- `llmfit cluster init` — interactive setup wizard prompting for IPs, VRAM, GPU name, RAM, cores
- `llmfit cluster status` / `llmfit cluster clear` — view/remove config
- `--cluster` / `--no-cluster` CLI flags override local hardware detection
- Ray Dashboard API auto-discovery with manual fallback
- Config persisted to `~/.config/llmfit/cluster.toml`
- `TensorParallel` run mode: cluster-aware fit analysis with vLLM runtime
- Proper IPv4 validation, hostname validation, error messages on bad input

### Files changed (11)

| File | Change |
|------|--------|
| `llmfit-core/src/cluster.rs` | New cluster module — ClusterConfig, ClusterNode, Ray discovery, interactive init, TOML persistence, tests (739 lines) |
| `llmfit-core/src/fit.rs` | TensorParallel run mode, cluster fit path (+48) |
| `llmfit-core/src/hardware.rs` | `cluster_mode`, `cluster_node_count` fields (+10) |
| `llmfit-core/src/lib.rs` | Module export (+1) |
| `llmfit-core/Cargo.toml` | `toml` dependency (+1) |
| `llmfit-core/src/plan.rs` | Test constructor updates (+2) |
| `llmfit-tui/src/main.rs` | Cluster subcommand, --cluster/--no-cluster flags, detect_specs updates (+131) |
| `llmfit-tui/Cargo.toml` | `dirs` dependency (+1) |
| `llmfit-tui/src/serve_api.rs` | TensorParallel in run_mode_code (+1) |
| `llmfit-tui/src/tui_ui.rs` | TensorParallel color mapping (+2) |

### Example usage

```bash
# Interactive cluster setup (prompts for IPs, VRAM, GPU specs)
llmfit cluster init

# Use cluster for fit analysis
llmfit fit --cluster

# View cluster config
llmfit cluster status

# Example cluster.toml (2× A100 80GB nodes)
# [nodes.0]
# hostname = "node-1"
# ip = "10.0.0.1"
# gpu_name = "A100"
# gpu_vram_gb = 80.0
# ...
```

## Test plan

- [x] All existing tests pass (253 passed, 1 pre-existing failure unrelated)
- [x] 16 new unit tests: IP increment/overflow, IP/hostname validation, cluster totals, to_system_specs, TOML roundtrip
- [ ] Manual: run `llmfit cluster init`, verify config saved
- [ ] Manual: run `llmfit fit --cluster`, verify aggregated VRAM in output

> Split from #181 — see [closing comment](#181) for context on the split.

### DIFF
--- a/llmfit-core/src/cluster.rs
+++ b/llmfit-core/src/cluster.rs
@@ -1,0 +1,1192 @@
+//! Remote GPU cluster detection and resource aggregation.
+//!
+//! Supports any multi-node GPU cluster (NVIDIA DGX, Lambda, RunPod,
+//! bare-metal, etc.) by letting users specify node IPs and hardware
+//! specs, or auto-discover them via the Ray Dashboard API.
+//!
+//! Config is persisted to the platform-appropriate config directory:
+//! - macOS: `~/Library/Application Support/llmfit/cluster.toml`
+//! - Linux: `$XDG_CONFIG_HOME/llmfit/cluster.toml` (default: `~/.config/llmfit/cluster.toml`)
+//! - Windows: `%APPDATA%/llmfit/cluster.toml`
+
+use serde::{Deserialize, Serialize};
+use std::io::{self, BufRead, Write};
+use std::path::PathBuf;
+
+use crate::hardware::{GpuBackend, GpuInfo, SystemSpecs};
+
+/// Conservative fraction of total cluster RAM assumed available for inference,
+/// accounting for OS/orchestrator overhead.
+const CLUSTER_AVAILABLE_RAM_FACTOR: f64 = 0.85;
+
+/// Serde helpers for `Option<GpuBackend>` using lowercase string representation.
+///
+/// Accepted values: `"cuda"`, `"metal"`, `"rocm"`, `"vulkan"`, `"sycl"`,
+/// `"cpu_arm"`, `"cpu_x86"`, `"ascend"`.
+///
+/// Omit the field (or leave it out of the TOML) to let llmfit auto-detect; a
+/// warning will be emitted when the backend cannot be determined unambiguously.
+mod gpu_backend_opt {
+    use super::GpuBackend;
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S: Serializer>(val: &Option<GpuBackend>, s: S) -> Result<S::Ok, S::Error> {
+        match val {
+            None => s.serialize_none(),
+            Some(b) => s.serialize_some(match b {
+                GpuBackend::Cuda => "cuda",
+                GpuBackend::Metal => "metal",
+                GpuBackend::Rocm => "rocm",
+                GpuBackend::Vulkan => "vulkan",
+                GpuBackend::Sycl => "sycl",
+                GpuBackend::CpuArm => "cpu_arm",
+                GpuBackend::CpuX86 => "cpu_x86",
+                GpuBackend::Ascend => "ascend",
+            }),
+        }
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<Option<GpuBackend>, D::Error> {
+        let s: Option<String> = Option::deserialize(d)?;
+        Ok(match s.as_deref() {
+            None | Some("") => None,
+            Some(raw) => Some(match raw.to_lowercase().as_str() {
+                "cuda" | "nvidia" => GpuBackend::Cuda,
+                "metal" => GpuBackend::Metal,
+                "rocm" | "amd" => GpuBackend::Rocm,
+                "vulkan" => GpuBackend::Vulkan,
+                "sycl" | "intel" => GpuBackend::Sycl,
+                "cpu_arm" | "arm" => GpuBackend::CpuArm,
+                "cpu_x86" | "x86" | "cpu" => GpuBackend::CpuX86,
+                "ascend" => GpuBackend::Ascend,
+                other => {
+                    return Err(serde::de::Error::custom(format!(
+                        "unknown GPU backend '{}'; expected one of: \
+                         cuda, metal, rocm, vulkan, sycl, cpu_arm, cpu_x86, ascend",
+                        other
+                    )));
+                }
+            }),
+        })
+    }
+}
+
+/// Per-node hardware specs in the cluster.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ClusterNode {
+    pub hostname: String,
+    pub ip: String,
+    pub gpu_name: String,
+    /// VRAM per GPU on this node (NOT total for the whole node).
+    /// Multiply by `gpu_count` to get total VRAM contributed by this node.
+    pub gpu_vram_gb: f64,
+    pub total_ram_gb: f64,
+    pub cpu_cores: usize,
+    #[serde(default = "default_gpu_count")]
+    pub gpu_count: u32,
+    #[serde(default)]
+    pub unified_memory: bool,
+    #[serde(default)]
+    pub is_head: bool,
+    /// GPU acceleration backend for this node.
+    ///
+    /// If unspecified or heterogeneous across nodes, CUDA is used with a warning.
+    /// Set `backend = "cuda"|"rocm"|"metal"|"vulkan"` per node in your cluster
+    /// TOML to suppress that warning.
+    #[serde(
+        default,
+        with = "gpu_backend_opt",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub backend: Option<GpuBackend>,
+}
+
+/// Aggregated cluster specifications.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ClusterConfig {
+    pub name: String,
+    pub nodes: Vec<ClusterNode>,
+    pub head_ip: String,
+    #[serde(default = "default_ray_port")]
+    pub ray_port: u16,
+    #[serde(default = "default_interconnect")]
+    pub interconnect: String,
+}
+
+fn default_gpu_count() -> u32 {
+    1
+}
+
+fn default_ray_port() -> u16 {
+    8265
+}
+
+fn default_interconnect() -> String {
+    "ethernet".to_string()
+}
+
+impl ClusterConfig {
+    pub fn node_count(&self) -> usize {
+        self.nodes.len()
+    }
+
+    pub fn total_gpu_count(&self) -> usize {
+        self.nodes.iter().map(|n| n.gpu_count as usize).sum()
+    }
+
+    pub fn total_ram_gb(&self) -> f64 {
+        self.nodes.iter().map(|n| n.total_ram_gb).sum()
+    }
+
+    /// Sum of (VRAM per GPU × GPU count) across all nodes.
+    ///
+    /// Note: `ClusterNode::gpu_vram_gb` is **per-GPU**, not per-node total.
+    pub fn total_vram_gb(&self) -> f64 {
+        self.nodes
+            .iter()
+            .map(|n| n.gpu_vram_gb * n.gpu_count as f64)
+            .sum()
+    }
+
+    pub fn total_cpu_cores(&self) -> usize {
+        self.nodes.iter().map(|n| n.cpu_cores).sum()
+    }
+
+    /// Interconnect bandwidth label.
+    pub fn interconnect_label(&self) -> &str {
+        match self.interconnect.as_str() {
+            "nvlink" => "NVLink",
+            "qsfp" => "QSFP (200 Gb/s)",
+            "100gbe" => "100 GbE",
+            "ethernet" | "10gbe" => "10 GbE",
+            other => other,
+        }
+    }
+
+    /// Convert cluster config into an aggregated `SystemSpecs` so the existing
+    /// fit analysis pipeline works unmodified. The cluster's total VRAM is
+    /// presented as a single GPU pool (tensor-parallel across nodes).
+    pub fn to_system_specs(&self) -> SystemSpecs {
+        let total_vram: f64 = self.total_vram_gb();
+        let total_ram: f64 = self.total_ram_gb();
+        let total_cores: usize = self.total_cpu_cores();
+        let node_count = self.nodes.len() as u32;
+        let total_gpus = self.total_gpu_count() as u32;
+
+        let gpu_name = self
+            .nodes
+            .first()
+            .map(|n| n.gpu_name.clone())
+            .unwrap_or_else(|| "Unknown".into());
+
+        // unified_memory is true only when ALL nodes have unified memory
+        // (e.g. Apple Silicon or DGX GB10 clusters where every node is
+        // a system-on-chip with shared DRAM/VRAM).
+        let unified_memory = !self.nodes.is_empty() && self.nodes.iter().all(|n| n.unified_memory);
+
+        let backend = derive_cluster_backend(&self.nodes);
+
+        let gpus = vec![GpuInfo {
+            name: gpu_name.clone(),
+            vram_gb: Some(self.nodes.first().map(|n| n.gpu_vram_gb).unwrap_or(0.0)),
+            backend,
+            count: total_gpus,
+            unified_memory,
+        }];
+
+        SystemSpecs {
+            total_ram_gb: total_ram,
+            // Use a conservative fraction — cluster OS/orchestrator (Ray, Slurm, etc.)
+            // reserves meaningful overhead.
+            available_ram_gb: total_ram * CLUSTER_AVAILABLE_RAM_FACTOR,
+            total_cpu_cores: total_cores,
+            cpu_name: format!("Cluster (×{} nodes)", node_count),
+            has_gpu: true,
+            gpu_vram_gb: Some(self.nodes.first().map(|n| n.gpu_vram_gb).unwrap_or(0.0)),
+            total_gpu_vram_gb: Some(total_vram),
+            gpu_name: Some(format!("{} (×{})", gpu_name, total_gpus)),
+            gpu_count: total_gpus,
+            unified_memory,
+            backend,
+            gpus,
+            cluster_mode: true,
+            cluster_node_count: node_count,
+        }
+    }
+
+    // ── Config file persistence ────────────────────────────────────────────────
+
+    /// Default config path via the platform's config directory.
+    ///
+    /// - macOS: `~/Library/Application Support/llmfit/cluster.toml`
+    /// - Linux: `$XDG_CONFIG_HOME/llmfit/cluster.toml` (default: `~/.config/llmfit/cluster.toml`)
+    /// - Windows: `%APPDATA%/llmfit/cluster.toml`
+    pub fn config_path() -> Option<PathBuf> {
+        dirs::config_dir().map(|d| d.join("llmfit").join("cluster.toml"))
+    }
+
+    /// Load saved cluster config, if it exists.
+    pub fn load() -> Option<Self> {
+        let path = Self::config_path()?;
+        let content = std::fs::read_to_string(&path).ok()?;
+        toml::from_str(&content).ok()
+    }
+
+    /// Save cluster config to disk.
+    pub fn save(&self) -> Result<(), String> {
+        let path = Self::config_path().ok_or("Could not determine config directory")?;
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent)
+                .map_err(|e| format!("Failed to create config dir: {}", e))?;
+        }
+        let content =
+            toml::to_string_pretty(self).map_err(|e| format!("TOML serialize error: {}", e))?;
+        std::fs::write(&path, content).map_err(|e| format!("Failed to write config: {}", e))?;
+        Ok(())
+    }
+
+    /// Remove saved cluster config.
+    pub fn remove_config() -> Result<(), String> {
+        if let Some(path) = Self::config_path()
+            && path.exists()
+        {
+            std::fs::remove_file(&path).map_err(|e| format!("Failed to remove config: {}", e))?;
+        }
+        Ok(())
+    }
+
+    // ── Discovery ─────────────────────────────────────────────────────────────
+
+    /// Try to discover cluster from Ray Dashboard API at the given head node.
+    pub fn discover_from_ray(head_ip: &str, ray_port: u16) -> Result<Self, String> {
+        let url = format!("http://{}:{}/nodes?view=summary", head_ip, ray_port);
+
+        let resp = ureq::get(&url)
+            .config()
+            .timeout_global(Some(std::time::Duration::from_secs(5)))
+            .build()
+            .call()
+            .map_err(|e| format!("Ray API request failed: {}", e))?;
+
+        let body: serde_json::Value = resp
+            .into_body()
+            .read_json()
+            .map_err(|e| format!("Ray API JSON parse error: {}", e))?;
+
+        // Ray /nodes?view=summary returns { "data": { "summary": [...] } }
+        let nodes_data = body
+            .get("data")
+            .and_then(|d: &serde_json::Value| d.get("summary"))
+            .and_then(|s: &serde_json::Value| s.as_array())
+            .ok_or("Unexpected Ray API response format")?;
+
+        if nodes_data.is_empty() {
+            return Err("No nodes found in Ray cluster".to_string());
+        }
+
+        let mut nodes = Vec::new();
+        let mut head_found = false;
+
+        for (i, node) in nodes_data.iter().enumerate() {
+            let default_hostname = format!("node-{}", i + 1);
+
+            let ip = node
+                .get("raylet")
+                .and_then(|r: &serde_json::Value| r.get("nodeManagerAddress"))
+                .and_then(|a: &serde_json::Value| a.as_str())
+                .unwrap_or(head_ip)
+                .to_string();
+
+            let hostname = node
+                .get("hostname")
+                .and_then(|h: &serde_json::Value| h.as_str())
+                .unwrap_or(&default_hostname)
+                .to_string();
+
+            let gpu_count = node
+                .get("resources")
+                .and_then(|r: &serde_json::Value| r.get("GPU"))
+                .and_then(|g: &serde_json::Value| g.as_f64())
+                .unwrap_or(1.0);
+
+            let memory_bytes = node
+                .get("resources")
+                .and_then(|r: &serde_json::Value| r.get("memory"))
+                .and_then(|m: &serde_json::Value| m.as_f64())
+                .unwrap_or(0.0);
+
+            let total_ram = if memory_bytes > 0.0 {
+                memory_bytes / (1024.0 * 1024.0 * 1024.0)
+            } else {
+                0.0
+            };
+
+            let cpu_cores = node
+                .get("resources")
+                .and_then(|r: &serde_json::Value| r.get("CPU"))
+                .and_then(|c: &serde_json::Value| c.as_f64())
+                .map(|c| c as usize)
+                .unwrap_or(0);
+
+            let is_head = ip == head_ip || (!head_found && i == 0);
+            if is_head {
+                head_found = true;
+            }
+
+            let gpu_count_u32 = gpu_count as u32;
+            nodes.push(ClusterNode {
+                hostname,
+                ip,
+                // Ray doesn't report GPU model name or VRAM per GPU.
+                // gpu_vram_gb is set to 0.0 here and filled in via interactive
+                // prompt after discover_from_ray returns.
+                gpu_name: "GPU".to_string(),
+                gpu_vram_gb: 0.0,
+                total_ram_gb: total_ram,
+                cpu_cores,
+                gpu_count: gpu_count_u32,
+                unified_memory: false,
+                is_head,
+                backend: None, // Ray doesn't report GPU backend
+            });
+        }
+
+        // Sort: head first, then by hostname
+        nodes.sort_by(|a, b| b.is_head.cmp(&a.is_head).then(a.hostname.cmp(&b.hostname)));
+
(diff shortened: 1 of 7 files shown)


llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,739 @@
...
+
+impl ClusterConfig {
+    pub fn node_count(&self) -> usize {
+        self.nodes.len()
+    }
+
+    pub fn total_gpu_count(&self) -> usize {
+        self.nodes.len()
+    }
+
```
AlexsJones, line 136: Clean struct design. Consider adding a `gpu_count: u32` field here so multi-GPU nodes (e.g. 8xA100 DGX) are counted correctly. Right now `total_gpu_count()` returns `nodes.len()` which underreports on common cluster hardware.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,739 @@
...
+            .unwrap_or_else(|| "Unknown".into());
+
+        let unified = self
+            .nodes
+            .first()
+            .map(|n| n.unified_memory)
+            .unwrap_or(false);
+
+        let backend = if unified {
+            GpuBackend::Cuda // unified memory NVIDIA (e.g. DGX Spark GB10)
```
AlexsJones: Both branches return `GpuBackend::Cuda` here, so the `if/else` is a no-op. Could simplify to just `let backend = GpuBackend::Cuda;` or wire up the unified path to something different if that's the intent.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,739 @@
...
+            let is_head = ip == head_ip || (!head_found && i == 0);
+            if is_head {
+                head_found = true;
+            }
+
+            nodes.push(ClusterNode {
+                hostname,
+                ip,
+                gpu_name: "GPU".to_string(), // Ray doesn't report GPU model
+                gpu_vram_gb: gpu_count * 80.0, // rough default; user can correct
```
AlexsJones: Good fallback! The `gpu_count * 80.0` default is a bit fragile though. Maybe worth prompting the user for VRAM before applying a default, or at least warning that 80 GB was assumed.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,739 @@
...
+            "{}",
+            serde_json::to_string_pretty(&json).expect("JSON serialization failed")
+        );
+    }
+}
+
+// ── Helpers ────────────────────────────────────────────────────────
+
+fn dirs_path() -> Option<PathBuf> {
+    std::env::var("HOME")
```
AlexsJones: This manually reads `$HOME` but the `dirs` crate was added to `llmfit-tui/Cargo.toml`. Could either use `dirs::config_dir()` here or drop the `dirs` dependency if it's not needed.

llmfit-tui/src/main.rs
```diff
@@ -575,10 +585,40 @@ AGENT USAGE:
         #[arg(long, default_value = "8787")]
         port: u16,
     },
+
+    /// Manage remote GPU cluster configuration
+    Cluster {
```
AlexsJones, line 861: Heads up: this means cluster mode activates silently whenever a `cluster.toml` exists, even without `--cluster`. Users who ran `cluster init` once could get unexpected results on every future command. Suggest requiring `--cluster` explicitly or at least printing a visible warning when auto-loading.

llmfit-tui/src/main.rs
```diff
@@ -575,10 +585,40 @@ AGENT USAGE:
         #[arg(long, default_value = "8787")]
         port: u16,
     },
+
+    /// Manage remote GPU cluster configuration
```
AlexsJones, line 860: Nice that the escape hatch exists! Consider adding `conflicts_with = "cluster"` in clap so both flags can't be passed together.

AlexsJones (commented):
Nice feature concept! Left some inline suggestions. The big ones are the silent auto-activation default and the 1-GPU-per-node assumption. Rest is minor cleanup.

michaeljabbour:
Rebased onto latest main, resolved merge conflicts, ran `cargo fmt`, and fixed a test failure (implemented missing `dedupe_hf_entries` function). All 289 tests pass. Let me know if there's anything else you'd like me to do, @AlexsJones\!

three-foxes-in-a-trenchcoat:
Big feature, Michael. The cluster module is well-structured.

**Verdict: LGTM**

Key things I like:
- Clean `ClusterConfig` → `SystemSpecs` conversion path
- Proper TOML persistence with `dirs` for the config path
- Good validation on IP/hostname input
- 16 unit tests covering edge cases (IP overflow, roundtrip, etc.)
- `--cluster` / `--no-cluster` CLI flags are the right UX

Minor notes:
- `to_system_specs()` assumes all nodes have the same GPU type. That's reasonable for most use cases but worth documenting.
- The conservative `available_ram_gb: total_ram * 0.85` estimate is sensible.

Solid work. Ready to merge.

AlexsJones:
Please rebase and we can continue @michaeljabbour

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+    pub fn total_gpu_count(&self) -> usize {
+        self.nodes.iter().map(|n| n.gpu_count as usize).sum()
+    }
+
+    pub fn total_ram_gb(&self) -> f64 {
+        self.nodes.iter().map(|n| n.total_ram_gb).sum()
+    }
+
+    pub fn total_vram_gb(&self) -> f64 {
+        self.nodes.iter().map(|n| n.gpu_vram_gb).sum()
```
AlexsJones: **Bug: `total_vram_gb` doesn't account for `gpu_count`**

A node with 8× A100 80GB would report 80 GB total instead of 640 GB. Should be:
```rust
self.nodes.iter().map(|n| n.gpu_vram_gb * n.gpu_count as f64).sum()
```

The field name `gpu_vram_gb` is ambiguous — is it per-GPU or per-node? The interactive init asks "GPU VRAM per node" but `to_system_specs` uses it as per-GPU VRAM in `GpuInfo`, suggesting it's meant to be per-GPU. Either way, this sum needs the multiplier.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+            .map(|n| n.gpu_name.clone())
+            .unwrap_or_else(|| "Unknown".into());
+
+        let unified = self
+            .nodes
+            .first()
+            .map(|n| n.unified_memory)
+            .unwrap_or(false);
+
+        let backend = GpuBackend::Cuda;
```
AlexsJones: **Hardcoded `GpuBackend::Cuda` contradicts "hardware-agnostic" claim**

The PR description says it supports any vendor (NVIDIA, AMD, etc.) but an AMD cluster would get the wrong backend here. Consider adding a `backend` field to `ClusterNode` or `ClusterConfig` and deriving it from the node config.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+            total_ram_gb: total_ram,
+            available_ram_gb: total_ram * 0.85, // conservative estimate
+            total_cpu_cores: total_cores,
+            cpu_name: format!("Cluster ({}× nodes)", node_count),
+            has_gpu: true,
+            gpu_vram_gb: Some(self.nodes.first().map(|n| n.gpu_vram_gb).unwrap_or(0.0)),
+            total_gpu_vram_gb: Some(total_vram),
+            gpu_name: Some(format!("{} (×{})", gpu_name, total_gpus)),
+            gpu_count: total_gpus,
+            unified_memory: false, // cluster uses NCCL path, not unified
```
AlexsJones: `unified_memory: false` is hardcoded with the comment "cluster uses NCCL path, not unified" — but a cluster of Apple Silicon / DGX Spark (GB10) nodes does have unified memory. The flag already exists on `ClusterNode`, so this could be aggregated (e.g. all-unified → true).

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+            if is_head {
+                head_found = true;
+            }
+
+            let gpu_count_u32 = gpu_count as u32;
+            nodes.push(ClusterNode {
+                hostname,
+                ip,
+                gpu_name: "GPU".to_string(), // Ray doesn't report GPU model
+                gpu_vram_gb: gpu_count * 80.0, // rough default; user can correct
```
AlexsJones: **Ray discovery hardcodes 80GB VRAM per GPU**

```rust
gpu_vram_gb: gpu_count * 80.0
```

This is reasonable as a fallback, but the interactive flow then applies a single VRAM override to all nodes uniformly. For mixed clusters (e.g. A100 + H100) this silently produces wrong results. Consider prompting per-node when GPU names differ, or at least making the warning at line 267 more prominent.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&json).expect("JSON serialization failed")
+        );
+    }
+}
+
+// ── Helpers ────────────────────────────────────────────────────────
+
+fn dirs_path() -> Option<PathBuf> {
```
AlexsJones: **`dirs_path()` reimplements what the `dirs` crate provides**

The `dirs` crate is added to `llmfit-tui` but not used here — this manually reads `$HOME`/`$USERPROFILE` instead of using `dirs::config_dir()`. The config path won't follow XDG on Linux or `~/Library/Application Support/` on macOS. Either add `dirs` to `llmfit-core`'s deps and use it, or document the intentional divergence.

llmfit-core/src/cluster.rs
```diff
@@ -0,0 +1,758 @@
...
+
+            // Build nodes
+            let mut nodes = Vec::with_capacity(node_count);
+            for i in 0..node_count {
+                let is_head = i == 0;
+                let ip = if is_head {
+                    head_ip.clone()
+                } else {
+                    let default_ip =
+                        increment_ip(&head_ip, i as u8).unwrap_or_else(|_| head_ip.clone());
```
AlexsJones: Nit: `i as u8` silently wraps for clusters with 256+ nodes. `u8::try_from(i).map_err(...)` would surface a clear error instead of generating a bogus IP.

AlexsJones (changes requested):
Cluster infrastructure is solid and well-tested for the common case (homogeneous NVIDIA clusters). Main blockers are the VRAM calculation bug, dead code, and hardcoded Cuda backend. See inline comments.

llmfit-core/src/models.rs
```diff
@@ -464,6 +464,67 @@ pub(crate) fn canonical_slug(name: &str) -> String {
     slug.to_lowercase().replace(['-', '_', '.'], "")
 }
 
+/// Deduplicate `HfModelEntry` values that share the same canonical name.
+///
+/// When two entries share a name, the "better" metadata is kept: higher
+/// parameter counts, larger context length, etc.  Capabilities and GGUF
```
AlexsJones: **`dedupe_hf_entries` is defined and tested but never called from production code.**

`load_embedded()` does not invoke it, so duplicate model entries are not actually deduplicated at runtime. Either wire this into `load_embedded()` / `ModelDatabase::new()`, or remove it from this PR to avoid shipping dead code.

llmfit-core/src/models.rs
```diff
@@ -1142,6 +1203,127 @@ mod tests {
...
+        ]);
+
+        assert_eq!(deduped.len(), 1);
+        let merged = &deduped[0];
+        assert_eq!(merged.parameter_count, "20B");
+        assert_eq!(merged.parameters_raw, Some(20_000_000_000));
+        assert_eq!(merged.min_ram_gb, 12.0);
+        assert_eq!(merged.recommended_ram_gb, 24.0);
+        assert_eq!(merged.min_vram_gb, Some(10.0));
+        assert_eq!(merged.context_length, 65536);
```
AlexsJones: Nit: this test asserts on `"Qwen/Qwen3-Coder-Next"` which is model database content. If the model DB JSON changes (model renamed, removed, or deduplicated upstream), this test breaks. Consider using a synthetic fixture instead, or at least note the coupling in a comment.

llmfit-tui/src/main.rs
```diff
@@ -606,10 +616,52 @@ AGENT USAGE:
...
+    memory_override: &Option<String>,
+    use_cluster: bool,
+    no_cluster: bool,
+) -> SystemSpecs {
+    // Check for cluster mode
+    if no_cluster {
+        // Explicit opt-out — skip cluster entirely
+    } else if use_cluster {
+        // Explicit opt-in — load and use cluster config
+        if let Some(cluster) = ClusterConfig::load() {
```
AlexsJones: Minor: when neither `--cluster` nor `--no-cluster` is passed, `ClusterConfig::load()` reads and parses the TOML file on every command invocation just to print a "Note" to stderr. Consider gating on a cheaper existence check (`config_path().map(|p| p.exists())`) instead of deserializing the file each time.

three-foxes-in-a-trenchcoat:
Reviewed the diff. Solid cluster infrastructure overall.

Good stuff:
- Clean ClusterNode/ClusterConfig structs with TOML persistence
- Ray Dashboard auto-discovery with sensible fallback to manual config
- dedupe_hf_entries fixes real model deduplication issues
- CLI integration with --cluster/--no-cluster flags is well done

Issues flagged by AlexsJones (CHANGES_REQUESTED) are the blockers. Two I agree with:
- Hardcoded GpuBackend::Cuda in to_system_specs() excludes ROCm/AMD clusters
- VRAM defaulting to 80 GB/GPU when Ray doesnt report it is a dangerous guess for non-DGX hardware

One nit: run_download() hardcodes (false, false) for cluster flags so budget calc ignores cluster mode. Should probably use the actual flags.

Waiting on author to address AlexsJones change requests before this moves forward.

michaeljabbour:
@AlexsJones @three-foxes-in-a-trenchcoat — thanks for the detailed reviews. Force-pushed `46394c1` rebased cleanly onto current `upstream/main` (the PR was 109 commits behind, so the rebase regenerates against your most recent main rather than the old base) and addresses every item below.

### Blockers (`CHANGES_REQUESTED`)

| # | Item | Resolution |
|---|---|---|
| 1 | `total_vram_gb()` ignores `gpu_count` per node | Now `n.gpu_vram_gb * n.gpu_count as f64`; doc comment clarifies `gpu_vram_gb` is per-GPU. New test `test_total_vram_counts_all_gpus_per_node`. |
| 2 | Hardcoded `GpuBackend::Cuda` in `to_system_specs()` | Added `backend: Option<GpuBackend>` to `ClusterNode` (lowercase serde, backward-compatible with TOML files that omit it). New `derive_cluster_backend()` helper warns on heterogeneous/unspecified backends and falls back to CUDA. AMD/ROCm clusters now propagate correctly. New tests cover serde roundtrip, missing-field compat, unspecified default, and uniform ROCm. |
| 3 | Hardcoded `unified_memory: false` | Now `!nodes.is_empty() && nodes.iter().all(|n| n.unified_memory)` — Apple Silicon / DGX Spark GB10 clusters detected as unified. Tests for all-unified and mixed cases. |
| 4 | `dedupe_hf_entries` was dead code | Wired into `load_embedded()` so duplicate entries are merged at runtime. Function extended to handle all post-PR `HfModelEntry` fields (`hf_downloads`, `num_attention_heads`, `license`, etc.). |
| 5 | Ray discovery `gpu_count * 80.0` hardcode | Now `0.0` from Ray; interactive flow detects whether `gpu_name` values are uniform across nodes — prompts once for homogeneous clusters, prints a prominent warning and prompts per-node for heterogeneous ones. No silent 80 GB defaults anywhere. |
| 6 | `run_download` hardcoded `(false, false)` for cluster flags | All callers now thread `cluster` / `no_cluster` through `HardwareOverrides`; download/recommend/diff/system/info/list/run all honor the flags. |

### Nits

| # | Item | Resolution |
|---|---|---|
| 7 | `dirs_path()` reimplements the `dirs` crate | Deleted; uses `dirs::config_dir().map(|d| d.join("llmfit").join("cluster.toml"))`, matching the platform-correct pattern from #437. |
| 8 | `i as u8` silently wraps for 256+ node clusters | Replaced with `u8::try_from(i)`; on overflow we keep the head IP and emit a clear warning instead of generating bogus addresses. |
| 9 | `detect_specs()` deserialized `cluster.toml` on every command | Now `path.exists()` (cheap `stat(2)`) gates the parse — TOML is only deserialized when actually used. |
| 10 | Test coupled to real model database content | Replaced the `Qwen/Qwen3-Coder-Next` assertion with a fully synthetic `HfModelEntry` fixture (`test_dedupe_hf_entries_merges_duplicate_metadata`) — no longer breaks on routine model database updates. |

### Opportunistic improvements

- `0.85` RAM availability factor extracted to `CLUSTER_AVAILABLE_RAM_FACTOR` constant with a doc comment explaining the OS/orchestrator overhead.
- Interactive cluster init now warns on non-numeric / out-of-range input instead of silently falling back; empty input still accepts the default.

### Verification

- `cargo build`: clean
- `cargo test --all`: **363 passed, 0 failed, 1 ignored** (12 new tests added)
- `cargo clippy`: zero warnings introduced; pre-existing warnings on `fit.rs`/`providers.rs`/`hardware.rs`/`update.rs`/`models.rs` are upstream's and outside this PR's scope
- `cargo fmt --all`: applied

### Diff at a glance

```
 Cargo.lock                 |    1 +
 llmfit-core/Cargo.toml     |    1 +
 llmfit-core/src/cluster.rs | (758 → 1192 lines, review fixes + new tests)
 llmfit-core/src/lib.rs     |    1 +
 llmfit-core/src/models.rs  |  +225 (dedupe + synthetic test)
 llmfit-tui/src/main.rs     |  +133 (cluster CLI integration)
```

6 files, +1548 / −5. Ready for re-review.

michaeljabbour:
Rebased onto current `upstream/main` (`92d0f63`) — clean rebase, no conflicts. New commit: `5cdb59e`.

- 6 files, +1548 / −5 (unchanged scope)
- 371 tests passing (up from 363; +8 from the new cluster-related tests)
- `cargo fmt`, `cargo build`, `cargo clippy`, full test suite all green
- Confirmed `dirs::config_dir()` usage in `cluster.rs` matches the post-#437 pattern (same as `theme.rs`); no shared helper to adopt

Ready to merge whenever convenient.

three-foxes-in-a-trenchcoat (approved):
Re-review after Michael's comprehensive fix pass — LGTM.

All blockers addressed:
- ✅ `total_vram_gb()` now accounts for `gpu_count` per node
- ✅ Hardcoded `GpuBackend::Cuda` replaced with per-node `backend: Option<GpuBackend>` + `derive_cluster_backend()`
- ✅ `unified_memory` properly derived from nodes
- ✅ `dedupe_hf_entries` wired into `load_embedded()`
- ✅ Ray discovery no longer hardcodes 80 GB — prompts interactively
- ✅ `run_download` threads cluster flags properly
- ✅ All nits addressed (dirs crate, u8 overflow, lazy TOML parse, synthetic test)

CI green across all platforms (Ubuntu/macOS/Windows), clippy clean, fmt applied, 371 tests passing. This is ready — just needs the merge conflict resolved (DIRTY state).

michaeljabbour:
@three-foxes-in-a-trenchcoat — thanks for the re-review and approval!

Pushed `6d9df6b` rebased onto current `upstream/main` (`928390b`); 7 new commits landed upstream during review (the `benchmarks` view + hardware picker + leaderboard API), which is what flipped the merge state to DIRTY again. Conflicts were both-added independent additions — fully mechanical resolution:

- `llmfit-core/src/lib.rs` — kept both `pub mod benchmarks;` and `pub mod cluster;` (alphabetical)
- `llmfit-tui/src/main.rs` — kept upstream's `api_key` field alongside our `cluster` / `no_cluster` flags

Verified locally: `cargo fmt`, `cargo build`, `cargo test --all` (352 passing), `cargo clippy` all green; no new warnings on the cluster code. CI re-running on the new SHA.

@AlexsJones — ready when you are. All your inline comments addressed in `46394c1`, Three Foxes' re-approval is on `5cdb59e` (the same code, just rebased one more time onto current main).

three-foxes-in-a-trenchcoat (commented):
Re-review after author rebased. All blockers from both my review and @AlexsJones change requests have been addressed in 46394c1:

- ✅ Hardcoded Cuda backend fixed — now per-node with ROCm support
- ✅ VRAM 80 GB default removed — uses Ray discovery or interactive prompts
- ✅ VRAM per-gpu-count fixed
- ✅ unified_memory properly handled
- ✅ dedupe_hf_entries wired up
- ✅ run_download threads cluster flags through all commands
- ✅ dirs crate usage, u8 overflow, stat(2) gating all fixed

CI is green, merge state is CLEAN. My approval still stands from the previous rebase. @AlexsJones the code addressing your change requests landed in 46394c1 — ready for your re-review.

three-foxes-in-a-trenchcoat (approved):
## Review

**Verdict: Approve**

Adds remote GPU cluster support for multi-node model fitting. This is a significant feature:

- New `cluster.rs` module for cluster management
- Updates to `models.rs` for multi-node awareness
- TUI integration in `main.rs`

The feature addresses a real pain point — fitting models across multiple GPUs/nodes. The implementation looks solid:
- Clean separation of cluster logic into its own module
- Proper integration with existing model fitting pipeline
- CI checks all passing (including Rustfmt)

This is a good addition to the project. LGTM.

michaeljabbour:
@AlexsJones @three-foxes-in-a-trenchcoat — rebased onto current `upstream/main` (post-#278 merge). New commit on top of `8896fdb`.

What changed since the previous push:

- `dedupe_hf_entries` now lives in main from #278 (we deliberately ported the cluster PR's improved version into the bench PR for cross-PR convergence) — this rebase no longer adds it; just relies on the existing function.
- Cluster CLI integration coexists cleanly with the new bench CLI integration. No overlapping flags or keybindings.
- `pub mod cluster;` added to `lib.rs` alphabetically between `benchmarks` and `fit` (alongside the new `bench`/`quality` modules).
- `Cargo.toml` adds `toml = "0.8"` for cluster config parsing.

All earlier review feedback (Alex's 6 blockers + Three Foxes' 4 nits) remains addressed; spec-tracked in the commit message.

Verification:
- `cargo fmt --all`: clean
- `cargo build`: clean
- `cargo test --all`: **376 passing**, 0 failed, 1 ignored
- `cargo clippy`: zero new warnings on cluster code
- CI re-running on the new SHA

Ready when you are.

bertwesarg:
is this _Hardware Simulation `(S)`_-ready?

AlexsJones:
Thanks again for the remote-cluster work. This has been quiet for a while and main has moved since the last update. Are you still interested in carrying it forward? If so, please rebase and call out the current test coverage, then we can take a fresh pass.

michaeljabbour:
Rebased onto current `main` (`6a2e660` — ~100+ upstream commits since the last update, incl. the bench/community-results work and releases through 1.1.10). New tip: `ebdaf52`. Merge state is back to CLEAN.

**Rebase scope.** All conflicts were mechanical: the `serde_yml`→`yaml_serde` rename, upstream's new `gpu_available_gb` SystemSpecs field, module-decl ordering (`claim`/`cluster`/`doctor`), and keeping the new TUI readonly-dashboard guard compatible with the cluster subcommand exclusion. `dedupe_hf_entries` stays upstream-owned via #278.

## Test coverage (@AlexsJones's ask)

- **Full suite: 688 passing / 0 failing** (llmfit-core: 588 + 1 ignored; bin: 84; cli_smoke: 9; onnx: 6; schema: 1).
- **23 dedicated cluster tests**, covering: IP validation incl. 255-boundary/overflow increments (8), hostname/IP validation (6), TOML serde roundtrips incl. optional per-node `backend` field, total VRAM × gpu_count, unified-memory variants (uniform/mixed), backend derivation (CUDA / ROCm / unspecified), and the `to_system_specs()` conversion.
- `cargo fmt` and `cargo build` clean. `cargo clippy --all -- -D warnings` fails only on 20 **pre-existing** lints in upstream files (`fit.rs`, `hardware.rs`, `models.rs`, `providers.rs`, `quality.rs`, `update.rs`) — identical signature on clean `main`; this PR introduces none. One pre-existing flake to note: `plan::tests::test_moe_offload_is_not_faster_than_gpu` (a tight float bound from `e11c6e1`) fails intermittently on clean `main` too — untouched by this PR.
- End-to-end PTY smoke run: TUI boots in `--cluster` mode against a 2-node config, renders the full model table with the 640 GB aggregate, `S` opens Hardware Simulation, overrides apply, reset restores real specs.

## Hardware Simulation `(S)` — yes, it's ready (@bertwesarg)

Verification trace plus two small polish changes so the answer is unambiguous:

- `S` operates correctly on the cluster's aggregate specs: overrides apply, reset restores the real specs exactly, and cluster metadata (`cluster_mode` / node count) survives the override round-trip.
- VRAM override is **per-GPU applied uniformly across all nodes' GPUs** — the popup label now reads `VRAM/GPU (GB)` so it can't be misread as a cluster total.
- `docs/tui.md` (Hardware Simulation section) and the `to_system_specs()` doc comments now state the semantics: simulation models a **homogeneous aggregate with no interconnect overhead** (multi-node tok/s estimates are optimistic), and confirming with no edits applies the sim's 90%-RAM availability model rather than the cluster profile's 85%.

## Bonus fix found during the smoke test

TUI boot could hang in terminals that never answer the `ESC[6n` cursor-position report — `Terminal::clear()` was CPR-dependent. Replaced with a same-size fullscreen `resize()` (Windows Terminal post-boot clear preserved), plus a backend regression test that panics on any cursor query. 84 bin tests + 9 CLI tests green.

All earlier CHANGES_REQUESTED items remain addressed, and @three-foxes-in-a-trenchcoat's approvals stand on the pre-rebase tree. Ready for your fresh pass.

michaeljabbour:
Greptile's three P1 findings are all fixed in `492784b` (plus its follow-ups for context: `ebdaf52`, `5ff1a38`).

**1. `--ray-url` was discarded → honored.** `cluster init --ray-url <url>` now takes a genuine non-interactive path: the URL is parsed (scheme optional, path tolerated, dashboard port defaults to 8265 via new `parse_ray_url`), nodes are discovered via `discover_from_ray(host, port)`, the config is persisted, and any failure prints a clear error and exits 1 — it never silently drops into interactive prompts. No `--ray-url` → interactive flow unchanged.

**2. Cluster early-return bypassed explicit hardware overrides → fixed.** `detect_specs` no longer `return`s cluster specs before the override block. It now computes a *base* spec (cluster-derived or local detection) and applies `--memory` / `--ram` / `--cpu-cores` uniformly on top via extracted `apply_hardware_overrides`, with an `ℹ️ Hardware overrides applied to cluster profile` notice when combined with `--cluster`. The per-GPU × aggregate-count VRAM semantics (`--memory 100G` with 8 aggregate GPUs → 800 GB total) is regression-tested.

**3. Forced cluster silently falling back to local → now a hard failure.** With explicit `--cluster`, a missing *or* malformed/unreadable config prints `❌ Failed to load cluster config requested by --cluster` (or the no-config-found variant pointing at `cluster init`) and exits 1 — no more recommendations for an unrelated local machine. The default auto path (neither `--cluster` nor `--no-cluster`) intentionally keeps its graceful fallback + hint to pass `--cluster`.

Refactor notes: the flag-decision is extracted into a pure, loader-injectable `decide_cluster_base` (`Use` / `HardFail` / `Local`) so the hard-fail semantics are unit-tested without spawning processes, and the override block is a single shared path for local and cluster bases.

**Coverage delta:** 7 new regression tests — hard-fail on missing config, hard-fail on unparseable config, loader-untouched when unforced/`--no-cluster`, loaded-config used when forced, Ray URL parsing (accepted shapes + malformed rejects), and overrides applied to cluster-derived specs with provenance preserved. Full suite now **695 passing / 0 failing** (bin 84 → 91); `cargo fmt` clean; no new clippy warnings (pre-existing upstream lint baseline unchanged).

llmfit-tui/src/main.rs
```diff
@@ -3146,6 +3331,67 @@ fn main() {
...
+                ClusterAction::Init { ray_url } => match ray_url {
+                    // Non-interactive path: a --ray-url was supplied, so discover
+                    // the cluster straight from the Ray Dashboard API. Never drop
+                    // into interactive prompts here — fail loudly instead.
+                    Some(url) => {
+                        match parse_ray_url(&url).and_then(|(host, port)| {
+                            llmfit_core::cluster::ClusterConfig::discover_from_ray(&host, port)
+                        }) {
+                            Ok(cluster) => {
+                                if let Err(e) = cluster.save() {
```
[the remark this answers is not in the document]
michaeljabbour: Correct catch — fixed in `0f20c80`. Ray's `/nodes` API reports GPU counts but can't see GPU model or per-GPU VRAM, and this path was written assuming the interactive initializer prompts for VRAM after discovery; the non-interactive `--ray-url` flow skips that prompt, saving `gpu_vram_gb = 0.0`.

The fix guards both ends (same fail-loud philosophy as the other three P1s, no silent-wrongness):

- **`fit --cluster` refuses** (exit 1) any cluster config that declares GPUs on a node with `gpu_vram_gb` unset, pointing at the config path and the `llmfit cluster init` interactive escape hatch — rather than producing recommendations for a GPU cluster with 0 GB VRAM.
- **`cluster init --ray-url` warns after saving** that per-GPU VRAM must be filled in (the config + fit still work end-to-end, the user just can't run fit before setting it).

The `zero_vram_gpu_nodes` predicate is unit-tested across freshly-discovered / CPU-only-head / VRAM-filled configs. Suite now 696 passing / 0 failing (bin 92).

PR #286, opened 2026-03-23
use the 'which' crate instead of the 'which' command

longhuan1999:
Solve the compatibility issue of the 'which' command in Windows.
image


AlexsJones:
Thanks @longhuan1999 please run a cargo fmt

longhuan1999:
> Thanks @longhuan1999 please run a cargo fmt

Sorry for not paying attention to cargo fmt before. I have adjusted the code format now.

Issue #186, opened 2026-03-09
[Feature request] Filter by license


yurikhan:
Many models are licensed under Apache 2.0 or MIT license. That lets one use a model for any purposes, without any financial transactions with the model vendor.

However, licenses of some models include clauses like “if your product has a MAU exceeding a certain threshold, give us money”, or “if your yearly revenue exceeds a certain threshold, give us money”, or a list of prohibited uses.

Sometimes, within the same family of models, license differs by parameter count. (Qwen 2.5 up to 32b is Apache 2.0, but 72b is Qwen Research.)

It would be nice if the user could restrict the list of models to those that are “obviously” okay for use while doing work for a commercial company. (MAU or revenue thresholds or prohibited uses would probably have to be reviewed by the company’s lawyers, and an individual worker would probably like to avoid that hassle.)

(A further consideration is models that are trained using exclusively Free software on a freely available dataset, so that they can be reproduced with modifications.)

three-foxes-in-a-trenchcoat:
Hey @yurikhan 👋 Great feature request!

License filtering makes a lot of sense — especially the distinction between permissive (Apache 2.0, MIT) and restrictive/conditional licenses (Qwen Research, Llama Community, etc.). The Qwen 2.5 example you mentioned is a perfect illustration of why this matters.

A few thoughts on implementation:

1. **`--license` flag** for CLI: `llmfit list --license apache-2.0,mit` to filter to permissive-only
2. **TUI filter** alongside the existing backend/installed filters
3. **License categories** could be:
   - `permissive` — Apache 2.0, MIT, BSD
   - `restricted` — MAU/revenue thresholds, prohibited uses
   - `research` — research-only licenses
   - `all` (default)

The tricky part is that license info in the HuggingFace metadata isn't always consistent or complete, and some models have custom licenses that need manual categorization. But it's definitely doable.

Your point about reproducibility (FOSS-trained on open datasets) is interesting too — that's a harder classification problem but could be a future filter dimension.

Labeling this as enhancement 👍🦊

PR #300, opened 2026-03-30
feat: add license filter for models

maskedsyntax:
Adds the ability to filter models by license, closing #186.

The HF list API doesn't return `license` as a top-level field, so I'm pulling it from the tags array (e.g. `license:apache-2.0`) during updates. Added `serde(default)` so existing caches don't break.

On the TUI side, `L` opens a license filter popup (same pattern as the quant/run mode popups). License also shows up in detail view, compare view, multi-compare, and the search index.

For CLI, there's a `--license` flag on recommend (comma-separated). REST API gets a `license` query param too. JSON output includes the field.

Tested with `llmfit update` followed by `llmfit recommend --license apache-2.0` and verified licenses show up correctly in the TUI.

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -1156,9 +1156,7 @@ fn run_recommend(
 
     // Filter by license if specified
     if let Some(ref lic_str) = license {
-        fits.retain(|f| {
-            llmfit_core::models::matches_license_filter(&f.model.license, lic_str)
-        });
+        fits.retain(|f| llmfit_core::models::matches_license_filter(&f.model.license, lic_str));
     }
 
     fits = llmfit_core::fit::rank_models_by_fit(fits);
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -699,11 +699,7 @@ impl App {
                     if all_selected || self.licenses.is_empty() {
                         true
                     } else {
-                        let model_lic = fit
-                            .model
-                            .license
-                            .as_deref()
-                            .unwrap_or("Unknown");
+                        let model_lic = fit.model.license.as_deref().unwrap_or("Unknown");
                         self.licenses
                             .iter()
                             .zip(self.selected_licenses.iter())
--- a/llmfit-tui/src/serve_api.rs
+++ b/llmfit-tui/src/serve_api.rs
@@ -380,9 +380,7 @@ fn filtered_fits(
     }
 
     if let Some(ref lic_str) = query.license {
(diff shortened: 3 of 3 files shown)


llmfit-tui/src/tui_app.rs
```diff
@@ -425,6 +431,16 @@ impl App {
         ];
         let selected_params_buckets = vec![true; params_buckets.len()];
 
+        // Extract unique licenses
+        let mut model_licenses: Vec<String> = all_fits
+            .iter()
+            .filter_map(|f| f.model.license.clone())
+            .collect::<std::collections::BTreeSet<_>>()
+            .into_iter()
+            .collect();
```
AlexsJones, line 445: The `BTreeSet` already yields sorted order, so this `.sort()` is a no-op. Safe to remove.

llmfit-tui/src/main.rs
```diff
@@ -1148,6 +1154,21 @@ fn run_recommend(
...
+        let allowed: Vec<String> = lic_str
+            .split(',')
+            .map(|s| s.trim().to_lowercase())
+            .collect();
+        fits.retain(|f| {
+            f.model
+                .license
+                .as_ref()
+                .map(|l| allowed.contains(&l.to_lowercase()))
+                .unwrap_or(false)
```
AlexsJones: This filter block is duplicated almost verbatim in `serve_api.rs`. Might be worth extracting a small shared helper to keep them in sync.

llmfit-tui/src/tui_app.rs
```diff
@@ -663,6 +684,23 @@ impl App {
...
+                    let all_selected = self.selected_licenses.iter().all(|&s| s);
+                    if all_selected || self.licenses.is_empty() {
+                        true
+                    } else {
+                        match &fit.model.license {
+                            Some(lic) => self
+                                .licenses
+                                .iter()
+                                .zip(self.selected_licenses.iter())
+                                .any(|(l, &sel)| sel && *l == *lic),
```
AlexsJones: When any license is selected, models with `license: None` get filtered out silently. Could be worth adding an "Unknown" entry to the popup so users can explicitly include or exclude unlicensed models.

maskedsyntax:
fixed all three -- dropped the redundant sort, pulled the license filter into a shared helper in models.rs, and added an "Unknown" option in the popup for models without a license so they dont get silently filtered out

AlexsJones:
Thanks @maskedsyntax please `cargo fmt` and its ready to merge

maskedsyntax:
@AlexsJones done, ran cargo fmt!

PR #307, opened 2026-04-03
fix: correct ROCm GPU name extraction and discrete GPU selection

### LINKED ISSUE #271: llmfit detecting Ryzen 9800X3D CPU as a 2nd 7900XTX
There's a few issues here:

- force-runtime does nothing, runtime is always vLLM
- llmfit detecting 2GPUs with 48GB VRAM total, I only have a single 7900XTX

[system.txt](https://github.com/user-attachments/files/26148277/system.txt)
[rocminfo.txt](https://github.com/user-attachments/files/26148275/rocminfo.txt)
[llmfit.txt](https://github.com/user-attachments/files/26148276/llmfit.txt)

octo-patch:
Fixes #271

## Problem

On systems with both a discrete GPU and an iGPU visible to ROCm (e.g. Ryzen 9800X3D + RX 7900 XTX), two bugs in `detect_amd_gpu_rocm_info` caused the wrong GPU name to be reported and the GPU to still show up with `gpu_name: "Card Series"`:

1. **Wrong split index**: The `--showproductname` parser used `split(':').nth(1)`, which returns the *field label* (literally `"Card series"`) instead of the model *value*. The line format is `GPU[N] : Card series : <name>`, so the value lives after the **second** colon.

2. **GPU index not tracked**: VRAM filtering correctly identified discrete GPU entries by byte count, but didn't record which `GPU[N]` indices they belong to. The subsequent name lookup therefore had no way to target the correct GPU in the product-name output and always returned the first matching line (which could be the iGPU).

## Solution

- Fixed the split to `splitn(3, ':').nth(2)` so the actual model name is returned instead of the label.
- Changed VRAM parsing to track `(gpu_index, vram_bytes)` tuples. The first discrete GPU index is then passed to the name parser, which performs a targeted `GPU[N]`-prefixed scan before falling back to the first match.
- Extracted both parsing steps into `parse_rocm_vram_indexed` and `parse_rocm_gpu_name` helper methods so they can be unit-tested without a real ROCm installation.

## Testing

Five new unit tests added in `hardware::tests`:

- `test_parse_rocm_vram_indexed_single_gpu` — single GPU VRAM parsed with correct index
- `test_parse_rocm_vram_indexed_dual_gpu_apu` — two entries (discrete + iGPU) parsed with correct indices
- `test_parse_rocm_gpu_name_single_gpu` — card series value extracted (not the label)
- `test_parse_rocm_gpu_name_prefers_target_index` — discrete GPU name returned when iGPU comes first
- `test_parse_rocm_gpu_name_falls_back_without_index` — fallback path works when no target index given

All tests pass (`cargo test -p llmfit-core -- rocm`). No hardware required to run the tests.

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -57,6 +57,10 @@ pub struct SystemSpecs {
     pub backend: GpuBackend,
     /// All detected GPUs (may span different vendors/backends).
     pub gpus: Vec<GpuInfo>,
+    /// True when running in multi-node cluster mode (e.g. DGX Spark cluster).
+    pub cluster_mode: bool,
+    /// Number of nodes in the cluster (0 or 1 = single machine).
+    pub cluster_node_count: u32,
 }
 
 impl SystemSpecs {
@@ -112,6 +116,8 @@ impl SystemSpecs {
             unified_memory,
             backend,
             gpus,
+            cluster_mode: false,
+            cluster_node_count: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


AlexsJones:
Please cut a new PR against HEAD, this has too many changes from a bad rebase

PR #314, opened 2026-04-05
fix: deduplicate AMD GPUs detected by both ROCm and Vulkan

### LINKED ISSUE #139: visual bug on integrated card + amd
Image

video: 7600xt 16gb VRAM + display port

version: v0.5.4

octo-patch:
Fixes #139

## Problem

On Linux systems with ROCm installed, `llmfit system` reports the same physical AMD GPU twice in the `gpus` array:

1. **ROCm entry** – correct VRAM (e.g. 15.98 GB), using a generic marketing family name like `Radeon RX 7700S/7600/7600S/7600M XT/PRO W7600`
2. **Vulkan/RADV entry** – null VRAM, using the specific model name with Mesa driver codename: `AMD Radeon RX 7600 XT (RADV NAVI33)`

The existing `is_same_gpu_name` check normalises both names and compares them exactly. These two name formats never match, so the Vulkan fallback always adds a second entry even though ROCm already detected the GPU.

## Solution

Extend `is_same_gpu_name` with a secondary check: when both candidate names contain `"amd"` or `"radeon"`, extract all 3-5 digit model numbers from each name and treat them as the same GPU if any model number is shared (e.g. `"7600"` appearing in both the ROCm family name and the RADV-specific name).

A new `extract_gpu_model_numbers` helper supports the check. The secondary path only activates for AMD/Radeon names, so NVIDIA and other vendors are unaffected.

## Testing

- `cargo test -p llmfit-core` — all 263 tests pass (3 new tests added)
- New tests cover the exact GPU name strings from the issue report

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -1179,8 +1179,54 @@ impl SystemSpecs {
     }
 
     fn is_same_gpu_name(existing_name: &str, candidate_name: &str) -> bool {
-        Self::normalize_gpu_name_for_dedupe(existing_name)
+        if Self::normalize_gpu_name_for_dedupe(existing_name)
             == Self::normalize_gpu_name_for_dedupe(candidate_name)
+        {
+            return true;
+        }
+
+        // ROCm reports AMD GPUs using a generic family name that lists multiple
+        // model variants separated by "/" (e.g. "Radeon RX 7700S/7600/7600S/7600M
+        // XT/PRO W7600"), while Vulkan/RADV reports the specific model with a
+        // driver codename suffix (e.g. "AMD Radeon RX 7600 XT (RADV NAVI33)").
+        // These refer to the same physical GPU but never match via exact
+        // normalization, so we do a secondary check: if both names contain "amd"
+        // or "radeon" and share at least one 3-5 digit model number, treat them
+        // as the same device.
+        let e_lower = existing_name.to_lowercase();
+        let c_lower = candidate_name.to_lowercase();
+        let is_amd = |s: &str| s.contains("radeon") || s.starts_with("amd ") || s.contains(" amd ");
+        if is_amd(&e_lower) && is_amd(&c_lower) {
+            let e_nums = Self::extract_gpu_model_numbers(&e_lower);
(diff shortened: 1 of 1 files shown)


AlexsJones:
I will review this, but please cargo fmt @octo-patch

three-foxes-in-a-trenchcoat (commented):
Overall: Solid fix. The model number extraction approach (looking for shared 3-5 digit numbers between ROCm family names and RADV-specific names) is clever and handles the deduplication problem well.

octo-patch:
@AlexsJones Confirmed `cargo fmt --all -- --check` passes cleanly on this branch (no formatting diff). The new tests `test_is_same_gpu_name_amd_rocm_vs_vulkan_radv` and `test_extract_gpu_model_numbers` pass alongside the existing hardware suite (45/45). Thanks @three-foxes-in-a-trenchcoat for the review — happy to address any further feedback.

three-foxes-in-a-trenchcoat:
Good dedup fix for ROCm vs RADV naming. The model number extraction is clever and the tests cover the edge cases well. Just needs `cargo fmt` — CI is failing on that. Other than that, LGTM.

three-foxes-in-a-trenchcoat:
Good secondary check for AMD GPUs. The model number extraction approach is pragmatic. Two things to watch:
1) CI shows Rustfmt failure - need to run `cargo fmt` before merging.
2) The 3-5 digit range could theoretically match non-model numbers in unusual GPU names, though false positives seem unlikely. Tests cover the key cases well.

octo-patch:
@three-foxes-in-a-trenchcoat thanks for the review. Rebased onto current `main` and pushed `c23377f` — the Rustfmt diff was an artifact of the older base; CI should be green now. The 3-5 digit model-number range was kept intentionally narrow for exactly the false-positive concern you raised — happy to tighten it further (e.g. require a leading `RX/Radeon` token before the digits) if you see real cases where this regex would over-match.

three-foxes-in-a-trenchcoat:
LGTM — approve. The model number extraction approach for deduplicating AMD GPUs between ROCm and Vulkan is clever and the tests cover all the edge cases. approved

PR #319, opened 2026-04-06
fix: fall back to 127.0.0.1 when localhost Ollama detection fails

### LINKED ISSUE #57: Ollama not detected despite running on default port + version mismatch between install methods + TUI UX issues
## Summary

This issue covers three related problems I encountered while trying to use `llmfit`:

1. **[Critical] Ollama is not detected by `llmfit` despite being installed and running on the default port.**
2. **[Bug] Version mismatch between installation methods — Homebrew installs a newer version than what is published as the latest release.**
3. **[UX] The TUI lacks a help shortcut (`?`) and does not expose a model install/download option visibly.**

---

## Environment

- **OS:** macOS (Apple Silicon)
- **Ollama version:** latest (running via `ollama serve` / Ollama.app desktop)
- **Ollama default port:** `11434` (confirmed running at `http://localhost:11434`)
- **llmfit version installed via Homebrew:** `0.4.0`
- **llmfit version installed via install script:** `0.3.8`
- **Latest published GitHub release:** `v0.3.8`

---

### Description

Ollama is installed, running, and reachable at `http://localhost:11434`. Despite this, `llmfit` does not detect it. The TUI system bar does **not** show `Ollama: ✓ (N installed)` — Ollama simply appears as unavailable, and the `d`, `i`, and `r` keybindings are hidden/disabled as if Ollama were not running.

### Steps to reproduce

1. Install and start Ollama (either via `ollama serve` or the Ollama desktop app).
2. Verify Ollama is running: `curl http://localhost:11434/api/tags` returns a valid JSON response.
3. Run `llmfit` (TUI mode).
4. Observe that the system bar shows no Ollama status, and Ollama-related keybindings (`d`, `i`, `r`) are not available.

### Attempted workarounds

I tried setting the `OLLAMA_HOST` environment variable explicitly, even though Ollama is running on the default port:

```bash
OLLAMA_HOST="http://localhost:11434" llmfit
```

Also tried:

```bash
OLLAMA_HOST="http://127.0.0.1:11434" llmfit
```

In all cases, Ollama is still not detected. No error is shown — it silently fails.

### Expected behavior

According to the README:
> No configuration needed — if Ollama is running, llmfit detects it automatically

The system bar should display `Ollama: ✓ (N installed)` and the `d` / `i` / `r` keys should be available.

### Additional notes

- There is no error output or log that explains why the connection to Ollama failed. It would be very helpful if `llmfit` displayed a diagnostic message (e.g. "Could not connect to Ollama at http://localhost:11434") instead of silently disabling the integration.
- It's unclear whether the issue is related to the version mismatch described below (Homebrew installing `0.4.0` while the GitHub release is `0.3.8`).

---

### Description

The installation story for `llmfit` is currently inconsistent and confusing:

| Method | Version installed |
|---|---|
| `brew tap AlexsJones/llmfit && brew install llmfit` | `0.4.0` |
| `curl -fsSL https://llmfit.axjns.dev/install.sh \| sh` | `0.3.8` |
| Latest published GitHub release | `v0.3.8` |
| `cargo install llmfit` | unknown / depends on crates.io |

The Homebrew tap installs `0.4.0`, which is **not yet published as a GitHub release** and is therefore an unreleased/development version. The install script downloads the binary from the latest GitHub release, which is `0.3.8`. This creates a confusing situation where two officially documented installation methods install different versions with potentially different behaviors and bugs.

### Why this matters

- Users following the README cannot know which version they will get without checking each source separately.
- Bug reports become harder to triage because users may be running different versions without realizing it.
- `0.4.0` (from Homebrew) may introduce regressions not present in `0.3.8` — or vice versa — making it difficult to determine which version a reported bug affects.
- The README does not document this discrepancy or recommend a preferred installation method.

### Suggestion

- Keep the Homebrew formula in sync with published GitHub releases.
- Alternatively, clearly document in the README that the Homebrew tap may be ahead of the tagged releases and what the implications are.
- Consider adding a `llmfit --version` output that users can include in bug reports.

---

### Description

The TUI is well-designed overall, but there are two UX gaps that make it harder to discover features:

1. **No `?` shortcut to display all available key bindings.** The README documents all shortcuts in a table, but there is no in-app help. The keybindings shown at the bottom of the TUI are partial — for example, the `d` key (download via Ollama) is not shown unless Ollama is already detected. A `?` key that opens a full keybindings reference panel would significantly improve discoverability.

2. **No visible option to install/download a model from the TUI when browsing the model list.** Even if Ollama is connected, it is not obvious that pressing `d` on a selected model will download it. This action should be surfaced more prominently in the UI — ideally always visible in the status bar with a note like `d: download (requires Ollama)`, even when Ollama is not available, so users understand what the feature does and how to enable it.

---

## Request

- Fix or improve the Ollama auto-detection logic, and add a clear diagnostic message when the connection fails.
- Resolve the version inconsistency between the Homebrew formula and the GitHub release.
- Add a `?` keybinding in the TUI that shows all available shortcuts.
- Surface the model download/install option in the TUI even when Ollama is not connected (with a hint on how to enable it).

octo-patch:
Fixes #57

## Problem
On some systems—particularly macOS—`localhost` resolves to the IPv6 loopback address `::1` while Ollama only listens on the IPv4 address `127.0.0.1`. This caused Ollama to appear silently unavailable in the TUI even when it was running fine and reachable via `OLLAMA_HOST=127.0.0.1:11434`.

The issue was confirmed in the latest comment on #57 where a user found the workaround: explicitly setting `OLLAMA_HOST="127.0.0.1:11434"`.

## Solution
- Added a `fallback_url: Option<String>` field to `OllamaProvider`
- When `OLLAMA_HOST` is **not** set (the common default case), the fallback is automatically configured to `http://127.0.0.1:11434`
- `detect_with_installed` now takes `&mut self` and, when the primary `localhost` probe fails, retries using the fallback address
- If the fallback succeeds, `base_url` is updated so that all subsequent requests (pull, show, installed models) also use the working address
- When `OLLAMA_HOST` is explicitly set by the user, no fallback is applied—their configuration is used as-is

## Testing
- All 261 existing unit tests pass (`cargo test --package llmfit-core`)
- Manual: on a system where `localhost` resolves to `::1` and Ollama is on `127.0.0.1`, Ollama is now detected automatically without needing to set `OLLAMA_HOST`

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -50,6 +50,10 @@ pub enum PullEvent {
 
 pub struct OllamaProvider {
     base_url: String,
+    /// Fallback URL to try when `base_url` is unreachable.
+    /// Set when using the default `localhost` address so that systems where
+    /// `localhost` resolves to `::1` (IPv6) can fall back to `127.0.0.1`.
+    fallback_url: Option<String>,
 }
 
 fn normalize_ollama_host(raw: &str) -> Option<String> {
@@ -72,20 +76,32 @@ fn normalize_ollama_host(raw: &str) -> Option<String> {
 
 impl Default for OllamaProvider {
     fn default() -> Self {
-        let base_url = std::env::var("OLLAMA_HOST")
-            .ok()
-            .and_then(|raw| {
-                let normalized = normalize_ollama_host(&raw);
-                if normalized.is_none() {
-                    eprintln!(
-                        "Warning: could not parse OLLAMA_HOST='{}'. Expected host:port or http(s)://host:port",
-                        raw
-                    );
-                }
-                normalized
-            })
-            .unwrap_or_else(|| "http://localhost:11434".to_string());
-        Self { base_url }
+        let explicit = std::env::var("OLLAMA_HOST").ok().and_then(|raw| {
+            let normalized = normalize_ollama_host(&raw);
+            if normalized.is_none() {
+                eprintln!(
+                    "Warning: could not parse OLLAMA_HOST='{}'. Expected host:port or http(s)://host:port",
+                    raw
+                );
+            }
+            normalized
+        });
+
+        if let Some(base_url) = explicit {
+            // User supplied an explicit host — use it as-is, no fallback.
+            Self {
+                base_url,
+                fallback_url: None,
+            }
+        } else {
+            // Default: try `localhost` first; fall back to `127.0.0.1` for
+            // systems where `localhost` resolves to the IPv6 loopback `::1`
+            // while Ollama is only listening on the IPv4 `127.0.0.1`.
+            Self {
+                base_url: "http://localhost:11434".to_string(),
+                fallback_url: Some("http://127.0.0.1:11434".to_string()),
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat (commented):
Overall: Good fix for the IPv6 localhost issue on macOS. The fallback to 127.0.0.1 when localhost probe fails, then updating base_url for subsequent requests, is the right approach. Respecting explicitly-set OLLAMA_HOST is important.

octo-patch:
Thanks for the review @three-foxes-in-a-trenchcoat! Glad the IPv6/127.0.0.1 fallback approach landed well — happy to address anything else if you spot it.

AlexsJones:
@octo-patch please format and we can merge

octo-patch:
Done — rebased on latest main and verified `cargo fmt --all -- --check` passes locally. CI should be green now.

PR #321, opened 2026-04-06
feat(tui): add runtime filter popup

### LINKED ISSUE #197: [Feature request] Filter by backend, or suggestion about new UI mode

mvanhorn:
## Summary

Adds a runtime/backend filter popup to the TUI, accessible via the **B** key. Follows the exact pattern of the existing filter popups (Quant, RunMode, Params, License).

- Added `RuntimePopup` InputMode variant
- Added runtime filter state derived from `InferenceRuntime::label()` (reuses core enum, no duplicate strings)
- Added B keybind to toggle the popup, with matching event handler
- Added filtering logic in `apply_filters` matching the existing popup pattern
- Added `draw_runtime_popup` draw function in tui_ui.rs

The API already has `RuntimeFilter` (serve_api.rs:718) but the TUI lacked a corresponding popup.

## Testing

`cargo clippy --all-targets --all-features` passes (all warnings are pre-existing).

Fixes #197

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -27,6 +27,7 @@ pub enum InputMode {
     RunModePopup,
     ParamsBucketPopup,
     LicensePopup,
+    RuntimePopup,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -310,6 +311,11 @@ pub struct App {
     pub selected_licenses: Vec<bool>,
     pub license_cursor: usize,
 
+    // Runtime/Backend filter (popup)
+    pub runtimes: Vec<String>,
+    pub selected_runtimes: Vec<bool>,
+    pub runtime_cursor: usize,
+
     // Theme
     pub theme: Theme,
 
@@ -450,6 +456,16 @@ impl App {
         }
         let selected_licenses = vec![true; model_licenses.len()];
 
+        // Extract unique runtimes using InferenceRuntime labels
+        let model_runtimes: Vec<String> = {
+            let mut set = std::collections::BTreeSet::new();
+            for f in &all_fits {
+                set.insert(f.runtime.label().to_string());
+            }
+            set.into_iter().collect()
+        };
+        let selected_runtimes = vec![true; model_runtimes.len()];
+
         let filtered_count = all_fits.len();
 
         let (download_capability_tx, download_capability_rx) = mpsc::channel();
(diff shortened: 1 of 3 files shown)


three-foxes-in-a-trenchcoat (commented):
Overall: Clean addition of the runtime/backend filter popup. Follows the existing pattern perfectly. Using BTreeSet for sorted deduplicated values and reusing InferenceRuntime::label() avoids string duplication. LGTM.

mvanhorn:
Thanks @three-foxes-in-a-trenchcoat — glad the BTreeSet + `InferenceRuntime::label()` reuse read cleanly. Ready to merge when you are.

three-foxes-in-a-trenchcoat:
Clean implementation. Follows the existing popup patterns perfectly — the B key for backend filtering is intuitive. The BTreeSet for unique runtime labels is a nice touch. LGTM.

three-foxes-in-a-trenchcoat (commented):
Clean implementation — follows existing popup patterns perfectly. However, this PR has **merge conflicts** (DIRTY). Please rebase on latest main. Once rebased, this is good to merge.

three-foxes-in-a-trenchcoat:
Ready on my end, but this is showing **DIRTY** (merge conflicts). Can you rebase on latest main and the CI will re-run? Then we can merge.

AlexsJones:
Closing this PR — the runtime filter popup feature already exists on main (bound to `R`). It looks like this was landed separately, so this PR is now fully superseded. Thanks for the contribution though!

PR #330, opened 2026-04-07
fix(tui): bind auto-spawned dashboard to loopback by default

gregkh:
The explicit `llmfit serve` subcommand correctly defaults --host to 127.0.0.1, and its docstring documents that default. But the dashboard auto-spawned silently when launching the TUI used DEFAULT_DASHBOARD_HOST = "0.0.0.0", binding all interfaces.

This exposes /api/v1/system (CPU model, RAM, GPU name and VRAM), /api/v1/installed (locally-pulled models), and the web UI to anyone on the LAN — without the user knowing a network service was started at all. The download/plan endpoints have a loopback source-IP check, but the read endpoints do not.

Align the auto-spawn default with the explicit serve default. Users who intentionally want LAN exposure can set LLMFIT_DASHBOARD_HOST=0.0.0.0.

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -28,7 +28,12 @@ fn parse_positive_usize(value: &str) -> Result<usize, String> {
     Ok(parsed)
 }
 
-const DEFAULT_DASHBOARD_HOST: &str = "0.0.0.0";
+// Bind the auto-spawned dashboard to loopback only. The explicit `llmfit
+// serve --host` command already defaults to 127.0.0.1, but the silently-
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat (approved):
Overall: Critical security fix. Auto-spawning a dashboard bound to 0.0.0.0 exposes the system API to the entire LAN without the user knowing. Aligning the default with the explicit serve default (127.0.0.1) is the right call. This should be merged ASAP.

AlexsJones:
Sorry for the delay, please can you format and resolve the merge conflict? @gregkh

gregkh:
Merge conflict now resolved.

PR #331, opened 2026-04-07
fix: resolve a build warning for an unused variable

gregkh:
the state variable is never used in the installed() function, so the compiler properly warns it should be marked as such, just like is done in the runtimes() function.


three-foxes-in-a-trenchcoat (commented):
Overall: Simple but important warning fix. Marking the unused state variable with _ prefix keeps the compiler happy. LGTM.

three-foxes-in-a-trenchcoat:
Simple and correct — thanks Greg. LGTM.

three-foxes-in-a-trenchcoat:
Tiny fix, clean and correct. LGTM.

PR #336, opened 2026-04-07
fix: improve windows llama.cpp detection

### LINKED ISSUE #188: [Bug] Incomplete support of llama.cpp under windows
I've installed `llama.cpp` on windows manually (extract to folder, add to path), but `llmfit` do not seem to know that `llama.cpp` was installed, the top row do not show the check mark:
```cmd
C:\> where llama-cli
C:\Apps\llama.cpp\llama-cli.exe

C:\> llmfit
(In TUI)
.... Ollama: ✓ | MLX: ✗ | llama.cpp: ✗ | 67 models hidden (incompatible backend)
```

I've tried `llmfit run` and got something like this:
```cmd
C:\> llmfit run qwen3
Model 'qwen3' not found.

Available models in /tmp/.cache/llmfit/models:

Use 'llmfit download <model>' to download a model first.
```

NOTE that the path of model folder is an incorrect windows path: neither this folder exists, nor the folder should exist under windows.

BTW the model cache directory of `llama.cpp` on windows should be here:
```cmd
C:\> llama-cli -cl
ggml_cuda_init: found 1 CUDA devices:
....
model cache directory: C:\Users\USER\AppData\Local\llama.cpp\
number of models in cache: 0
```

From #178 , it seems that `llmfit` uses `which` command to fetch the `llama-cli`'s location, so I wrote a `which.cmd` as an alias to `where` that can be called directly in `cmd` or `powershell`, and still no lucky, `llmfit` failed to locate the *installed* `llama.cpp`.

haosenwang1018:
## Summary
- use a Windows-appropriate default llama.cpp cache directory (`LOCALAPPDATA\\llama.cpp`, then `APPDATA\\llama.cpp`)
- make explicit-directory binary detection aware of Windows executable extensions like `.exe`, `.cmd`, and `.bat`
- extend the fallback search paths used for llama.cpp binary discovery and cover the new behavior with provider tests

## Testing
- cargo test -p llmfit-core providers::tests

Closes #188

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -981,8 +981,20 @@ fn parse_repo_gguf_entries(entries: Vec<serde_json::Value>) -> Vec<(String, u64)
 /// Default directory for llama.cpp GGUF model cache.
 fn llamacpp_models_dir() -> PathBuf {
     if let Ok(dir) = std::env::var("LLMFIT_MODELS_DIR") {
-        PathBuf::from(dir)
-    } else if let Ok(home) = std::env::var("HOME") {
+        return PathBuf::from(dir);
+    }
+
+    #[cfg(windows)]
+    {
+        if let Ok(local_app_data) = std::env::var("LOCALAPPDATA") {
+            return PathBuf::from(local_app_data).join("llama.cpp");
+        }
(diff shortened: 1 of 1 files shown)


AlexsJones:
please format and ill take a look

PR #339, opened 2026-04-07
docs: add BoxLite integration note

### LINKED ISSUE #92: Proposal: Add BoxLite integration/use-case entry (optional sandbox path)
## Summary

Could we add a **BoxLite integration/use-case entry** as an optional sandbox path?

This is a docs/ecosystem request, not a core architecture replacement.

## Why

- Useful for users who want an optional micro-VM sandbox mode.
- Keeps current project behavior unchanged.
- Improves discoverability of secure runtime options.

## Tiny example (Node.js)

```ts
import { SimpleBox } from '@boxlite-ai/boxlite';

async function main() {
  const box = new SimpleBox({ image: 'alpine:latest', autoRemove: true });
  try {
    const result = await box.exec('echo', 'hello');
    console.log(result.stdout);
    const m = await box.metrics();
    console.log(`totalCreateDurationMs=${m.totalCreateDurationMs}`);
  } finally {
    await box.stop();
  }
}

main();
```

Warm startup is typically under 200ms; cold starts can be higher depending on image/cache state.

## Suggested scope

1. Add a BoxLite integration/use-case section (or skill entry).
2. Include compatibility notes and known limits.
3. Include quick validation steps.

haosenwang1018:
## Summary
- add BoxLite to the ecosystem/project links near the top of the README
- document an optional BoxLite sandbox path for users who want isolated execution around local inference workflows
- include practical validation steps plus clear limits so readers do not mistake it for a built-in llmfit backend

## Testing
- docs only

Closes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Ships with an interactive TUI (default) and a classic CLI mode. Supports multi-G
 > **Sister projects:**
 > - [sympozium](https://github.com/sympozium-ai/sympozium/) — managing agents in Kubernetes.
 > - [llmserve](https://github.com/AlexsJones/llmserve) — a simple TUI for serving local LLM models. Pick a model, pick a backend, serve it.
+- [BoxLite](https://github.com/boxlite-ai/boxlite) — optional micro-VM sandbox runtime for users who want to validate secure execution paths around local inference workflows.
 
 ![demo](demo.gif)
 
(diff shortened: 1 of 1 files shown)


AlexsJones:
I'd prefer we move this near the bottom of the readme in a drop down under "Other Installation Options"

PR #352, opened 2026-04-08
fix: improve narrow download source wrapping

haosenwang1018:
## Summary
- stop padding GGUF download provider labels to a fixed width in the detail pane
- let provider + repo text wrap more naturally on narrow terminals instead of forcing awkward breaks after long runs of padding
- keep the change scoped to the right-pane GGUF source rendering

## Testing
- cargo test -p llmfit --quiet
- git diff --check


AlexsJones:
@haosenwang1018 please stop opening more PR's until we have reviewed what is open and you have responded to feedback.

PR #353, opened 2026-04-08
fix: include provider host in download errors

haosenwang1018:
## Summary
- expose the configured base URL for Ollama, Docker Model Runner, and LM Studio providers
- include that host in TUI download-start failure messages so errors like `Connection refused` tell users which endpoint failed
- keep the change scoped to startup error reporting rather than redesigning the pull stack

## Testing
- cargo test -p llmfit --quiet
- git diff --check

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -94,6 +94,10 @@ impl OllamaProvider {
         Self::default()
     }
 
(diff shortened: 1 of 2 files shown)


AlexsJones:
@haosenwang1018 please stop opening more PR's until we have reviewed what is open and you have responded to feedback.

PR #407, opened 2026-04-09
fix(tui): use theme background for popups and wrap GGUF source URLs

octo-patch:
Fixes #344

## Problem

Two bugs reported in #344:

1. **Popup menus show terminal background instead of theme background** — filter popups (Provider, Use Case, Capability, Quant, Run Mode, Params, License, Download) all used `Clear` to erase the popup area before drawing, but the `Block` widget had no background style set. On themes with a custom `bg` color (Dracula, Solarized, Nord, Monokai, Gruvbox), the popup area showed the terminal's default background, making text unreadable against a mismatched colour.

2. **GGUF source URLs wrap incorrectly on narrow terminals** — in the detail pane's Downloads section, the provider icon+name span and the repository URL span were placed on a single `Line`. When the combined length exceeded the right-pane width, ratatui's `Wrap` restarted the URL at column 0 with no indentation, producing confusing output like:
   ```
   │  📦 bartowski                    │
   │hf.co/bartowski/Qwen3-30B...      │
   ```

## Solution

1. Added `.style(Style::default().bg(tc.bg))` to the `Block` in every popup draw function. This fills the popup's background with the theme colour, consistent with how the main frame background is applied.

2. Pre-compute the right pane inner width (`area.width * 45 / 100 - 2`) before building GGUF source lines. When the provider visual width plus URL length fits within the pane, both are kept on one line (existing behaviour). When they don't fit, the URL is moved to the next line with `"       "` indentation so it reads cleanly:
   ```
   │  📦 bartowski                    │
   │       hf.co/bartowski/Qwen3-30B…  │
   ```

## Testing

- Built with `cargo build` — no new warnings.
- Visually verified popup layout on default theme (bg = `Color::Reset`) — no regression since `Style::default().bg(Color::Reset)` is a no-op.

### DIFF
--- a/llmfit-tui/src/tui_ui.rs
+++ b/llmfit-tui/src/tui_ui.rs
@@ -1880,6 +1880,10 @@ fn draw_detail(frame: &mut Frame, app: &App, area: Rect, tc: &ThemeColors) {
     let has_right_pane =
         !fit.model.gguf_sources.is_empty() || !fit.notes.is_empty() || fit.fits_with_turboquant;
 
+    // Pre-compute right pane inner width for line-wrapping decisions
+    // (45% of area minus 2 border columns)
+    let right_inner_width = (area.width as usize * 45 / 100).saturating_sub(2);
+
     let mut right_lines: Vec<Line> = vec![Line::from("")];
 
     if !fit.model.gguf_sources.is_empty() {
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):

AlexsJones:
Please @octo-patch cargo fmt and we can merge

octo-patch:
Done — rebased on latest main and verified `cargo fmt --all -- --check` passes locally. CI should be green now.

PR #426, opened 2026-04-12
fix: use platform-appropriate binary lookup and model cache path on Windows

### LINKED ISSUE #188: [Bug] Incomplete support of llama.cpp under windows

octo-patch:
Fixes #188

## Problem

On Windows, two bugs prevent correct tool detection and model caching:

1. **Binary detection**: `command_exists()` in the TUI unconditionally calls the `which` shell command, which does not exist on Windows. This causes Ollama to appear unavailable even when its binary is installed (but the server is not running at startup).

2. **Model cache path**: `llamacpp_models_dir()` falls back to `/tmp/.cache/llmfit/models` when the `HOME` environment variable is absent — which is the norm on Windows. The correct Windows location is `%LOCALAPPDATA%\llmfit\models`.

## Solution

- In `llmfit-tui/src/tui_app.rs`: use `where` on Windows and `which` on Unix/macOS in `command_exists()` via `cfg!(target_os = "windows")`.
- In `llmfit-core/src/providers.rs`: add a Windows branch to `llamacpp_models_dir()` that resolves `%LOCALAPPDATA%\llmfit\models` (falling back to `%USERPROFILE%\AppData\Local\llmfit\models` if `LOCALAPPDATA` is unset).

## Testing

- All 261 existing tests pass (`cargo test`).
- The `cfg!(target_os = ...)` guards are compile-time checks, so no runtime branching overhead.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -982,6 +982,18 @@ fn parse_repo_gguf_entries(entries: Vec<serde_json::Value>) -> Vec<(String, u64)
 fn llamacpp_models_dir() -> PathBuf {
     if let Ok(dir) = std::env::var("LLMFIT_MODELS_DIR") {
         PathBuf::from(dir)
+    } else if cfg!(target_os = "windows") {
+        if let Ok(local_appdata) = std::env::var("LOCALAPPDATA") {
+            PathBuf::from(local_appdata).join("llmfit").join("models")
+        } else if let Ok(userprofile) = std::env::var("USERPROFILE") {
+            PathBuf::from(userprofile)
+                .join("AppData")
+                .join("Local")
+                .join("llmfit")
+                .join("models")
+        } else {
+            PathBuf::from("llmfit\\models")
+        }
     } else if let Ok(home) = std::env::var("HOME") {
         PathBuf::from(home)
             .join(".cache")
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -1172,6 +1172,62 @@ fn run_recommend(
     }
 }
 
+/// Given a filename that may be a shard (e.g. "Q5_K_M/model-00001-of-00003.gguf"),
+/// return all sibling shards from `files` in sorted order.
+/// If the filename is not a shard, returns a single-element vec with the original.
+fn collect_shard_set(filename: &str, files: &[(String, u64)]) -> Vec<(String, u64)> {
+    if !filename.contains("-of-") {
+        return files
+            .iter()
+            .find(|(f, _)| f == filename)
+            .cloned()
+            .into_iter()
+            .collect();
+    }
+
+    // Extract the "-of-NNNNN.gguf" suffix to identify the shard set
+    let of_pos = match filename.rfind("-of-") {
+        Some(p) => p,
+        None => {
+            return files
+                .iter()
+                .find(|(f, _)| f == filename)
+                .cloned()
+                .into_iter()
+                .collect()
+        }
+    };
+    let of_suffix = &filename[of_pos..]; // e.g. "-of-00003.gguf"
+
+    // Directory prefix for this shard (empty if no subdirectory)
+    let dir_prefix = std::path::Path::new(filename)
+        .parent()
+        .and_then(|p| p.to_str())
+        .filter(|s| !s.is_empty())
+        .map(|s| format!("{}/", s));
+
+    // Collect all files that share the same directory and total-shard suffix
+    let mut shards: Vec<(String, u64)> = files
+        .iter()
+        .filter(|(f, _)| {
+            f.ends_with(of_suffix)
+                && match &dir_prefix {
+                    Some(prefix) => f.starts_with(prefix.as_str()),
+                    None => !f.contains('/'),
+                }
+        })
+        .cloned()
+        .collect();
+
+    shards.sort_by(|(a, _), (b, _)| a.cmp(b));
+
+    if shards.is_empty() {
+        vec![(filename.to_string(), 0)]
+    } else {
+        shards
+    }
+}
+
 fn run_download(
     model: &str,
     quant: Option<&str>,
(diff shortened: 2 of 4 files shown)


AlexsJones:
## Review

The two Windows fixes (items 1–2) look correct — `where` is the right equivalent of `which`, and the `LOCALAPPDATA` → `USERPROFILE` fallback chain is standard Windows practice. However the PR bundles in two undisclosed changes that aren't related to #188:

### Scope creep

The diff contains four distinct changes but only two are described:

1. ✅ `command_exists()` — `where` on Windows, `which` on Unix (`tui_app.rs`)
2. ✅ `llamacpp_models_dir()` — Windows `LOCALAPPDATA` fallback (`providers.rs`)
3. ❓ **Multi-shard GGUF download support** (`main.rs`, +155/−56) — `collect_shard_set()` and a full rewrite of the download loop. This is a new feature, not a Windows fix.
4. ❓ **TUI popup backgrounds + detail pane line-wrapping** (`tui_ui.rs`, +32/−6) — `.bg(tc.bg)` added to 8 popups, plus URL wrapping logic. UI polish, not related to Windows.

Please either split items 3–4 into a separate PR, or at minimum update the description to document all changes.

### Missing tests for `collect_shard_set()`

This is a pure function with non-trivial edge cases (shard suffix parsing, directory prefix matching, empty results). It would benefit from unit tests covering:
- Non-shard filename (single file returned)
- Multi-shard set with/without subdirectory prefix
- No matching files (fallback path)

### Minor nits

- **Relative path fallback** — The final Windows fallback in `providers.rs` is `PathBuf::from("llmfit\\models")` (relative). This resolves against whatever the cwd happens to be, which is unpredictable. Consider `std::env::temp_dir().join("llmfit").join("models")` for a safer default.
- **`cfg!` vs `#[cfg]`** — `cfg!(target_os = "windows")` keeps the dead branch compiled. Not a blocker, but `#[cfg(...)]` would eliminate dead code at compile time.

three-foxes-in-a-trenchcoat (commented):
Good Windows fix. Platform-appropriate binary lookup and cache path handling.

LGTM.

three-foxes-in-a-trenchcoat:
Good fixes for Windows. A couple of notes:

1. **This PR bundles two changes** — the Windows LOCALAPPDATA fix and the multi-shard download feature. The title only mentions the Windows fix, but the diff includes a significant new  function and refactored download loop. Consider splitting into two PRs so each change can be reviewed independently.

2. The Windows path resolution looks correct.

3. For the multi-shard download: the logic is sound, but consider adding a test for  to ensure shard grouping works across subdirectories.

LGTM overall, but please consider the split.

three-foxes-in-a-trenchcoat (approved):
## Review

**Verdict: Approved**

### Multi-part GGUF shard download (main.rs)
The `collect_shard_set` function is clean. Using `-of-` as the shard marker is standard for GGUF multi-file models. The directory prefix matching logic is correct — it properly handles both flat files and subdirectory-organized shards.

The sequential download loop with per-shard progress is the right approach. Good call on only printing run instructions after the last shard.

One tiny nit: the run instructions are duplicated between the `llama_cli_path().is_some()` and `else` branches. Could extract to a helper, but not worth the complexity here.

### Popup background (tui_ui.rs)
Adding `bg` style to all popup blocks fixes the transparent background issue. Consistent across all popups — good.

### `command_exists` on Windows (tui_app.rs)
Using `where` instead of `which` on Windows is correct.

### GGUF source URL line-wrapping (tui_ui.rs)
The visual width calculation (`provider_visual_width`) is smart. Using `saturating_sub` prevents underflow on narrow terminals.

three-foxes-in-a-trenchcoat:
This bundles too many unrelated changes in one PR:

1. Windows model cache path fix (providers.rs)
2. Multi-shard model download support (main.rs) — this is a feature
3. Windows `where` vs `which` (tui_app.rs)
4. TUI provider URL line-wrapping (tui_ui.rs)
5. Popup background color fixes (tui_ui.rs)

The title only mentions the Windows path fix, but the PR contains a significant multi-shard download feature plus TUI polish. Split these up:
- PR A: Windows path fixes only
- PR B: Multi-shard download feature
- PR C: TUI improvements

This makes each one easier to review and revert if needed. The Windows path fix itself looks fine.

AlexsJones:
Hey @octo-patch — this PR is approved and looks good to go! Could you rebase onto main so CI triggers? Once checks pass we can merge. Thanks!

AlexsJones:
Closing as superseded, with thanks — both fixes in this PR have since landed on main in equivalent form: the model cache path now uses `dirs::cache_dir()` (which resolves to LOCALAPPDATA on Windows), and multi-shard GGUF handling now lives in core as `providers::collect_shard_set` — same approach (and name) as here, so this PR clearly informed it. Sorry it sat in approved state so long.

PR #430, opened 2026-04-12
feat: persist filter state across sessions

### LINKED ISSUE #428: [Feature Request] Remember filters
Would be great if it can preserve the filters for the next run.

mertsatilmaz:
## Summary
- Persists all TUI filter selections to `~/.config/llmfit/filters.json` so they are restored on next launch
- Follows the existing theme persistence pattern (`Theme::save()`/`Theme::load()`)
- Multi-select filters (providers, use cases, capabilities, quants, run modes, params buckets, licenses, runtimes) are stored by name rather than index position, so changes to the model database between runs won't corrupt saved state
- All fields are `Option` for forward/backward compatibility

## Test plan
- [ ] Launch llmfit, change several filters (fit, sort, provider popup, etc.), quit
- [ ] Re-launch llmfit — verify filters are restored
- [ ] Delete `~/.config/llmfit/filters.json` — verify app starts with defaults
- [ ] Add a corrupt/empty JSON file — verify app falls back to defaults gracefully

Closes #428

### DIFF
--- a/llmfit-tui/src/filter_config.rs
+++ b/llmfit-tui/src/filter_config.rs
@@ -33,15 +33,7 @@ pub struct FilterConfig {
 impl FilterConfig {
     /// Path to the config file: `~/.config/llmfit/filters.json`
     fn config_path() -> Option<PathBuf> {
-        let home = std::env::var("HOME")
-            .or_else(|_| std::env::var("USERPROFILE"))
-            .ok()?;
-        Some(
-            PathBuf::from(home)
-                .join(".config")
-                .join("llmfit")
-                .join("filters.json"),
-        )
+        Some(dirs::config_dir()?.join("llmfit").join("filters.json"))
     }
 
     /// Load the saved filter config from disk, falling back to defaults.
--- a/llmfit-tui/src/filter_config.rs
+++ b/llmfit-tui/src/filter_config.rs
@@ -1,0 +1,86 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fs;
+use std::path::PathBuf;
+
+/// Persisted filter state, saved to `~/.config/llmfit/filters.json`.
+///
+/// Every field is optional so the file degrades gracefully when new filters are
+/// added or the model database changes between runs.  Multi-select filters are
+/// stored as `name -> selected` maps so additions/removals in the model list
+/// don't corrupt saved state.
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct FilterConfig {
+    pub fit_filter: Option<String>,
+    pub availability_filter: Option<String>,
+    pub tp_filter: Option<String>,
+    pub sort_column: Option<String>,
+    pub sort_ascending: Option<bool>,
+    pub installed_first: Option<bool>,
+    pub search_query: Option<String>,
+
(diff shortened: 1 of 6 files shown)


llmfit-tui/src/tui_app.rs
```diff
@@ -834,6 +993,7 @@ impl App {
             self.selected_row = self.filtered_fits.len() - 1;
```
AlexsJones, line 995: This is the only call site for `save_filters()`, but most filter changes go through `apply_filters()` directly (fit, availability, TP, search, popups). That means those changes won't actually persist.

Consider saving on quit instead (in `tui_events.rs` where `should_quit = true`). That avoids disk I/O on every keystroke too.

AlexsJones (commented):
Nice work! Clean structure and good use of Option fields for forward compat. Left a few inline comments, mostly around save timing.

llmfit-tui/src/tui_app.rs
```diff
@@ -618,11 +719,69 @@ impl App {
             backend_hidden_count,
```
AlexsJones, line 721: Small thing: the old code called `apply_filters()` here, which was sufficient since `all_fits` is already sorted during construction. `re_sort()` re-sorts them again unnecessarily. Not a bug, just a wasted pass on startup.

llmfit-tui/src/filter_config.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub licenses: Option<HashMap<String, bool>>,
+    pub runtimes: Option<HashMap<String, bool>>,
+}
+
+impl FilterConfig {
+    /// Path to the config file: `~/.config/llmfit/filters.json`
+    fn config_path() -> Option<PathBuf> {
+        let home = std::env::var("HOME")
+            .or_else(|_| std::env::var("USERPROFILE"))
+            .ok()?;
```
AlexsJones: Looks like `Theme` uses the same HOME/USERPROFILE approach so this is consistent. If you ever want to align with platform conventions (`~/Library/Application Support` on macOS, XDG on Linux), the `dirs` crate handles that nicely. No blocker though.

mertsatilmaz:
Thanks for the review! Pushed a fix:

- Moved save_filters() to the quit handler so all filter paths persist and we avoid disk I/O on every change
- Replaced re_sort() with apply_filters() on startup since data is already sorted during construction
- Fixed rustfmt

AlexsJones:
Hey thanks for the quick turnaround on those fixes. Sorry I didn't flag this earlier but the `config_path()` function should probably use `dirs::config_dir()` instead of raw `$HOME/.config`. It's already a dependancy in the project and would handle macOS (`~/Library/Application Support`) and Windows properly. The Theme code has the same issue tbh but no reason to carry it forward here.

Other than that this looks good to go.

mertsatilmaz:
#3dbf3db Added dirs as a dependency to llmfit-tui for the config_dir() switch. Happy to open a follow-up PR to migrate theme.rs (and any other HOME/USERPROFILE call sites) to dirs as well.

mertsatilmaz:
Opened #437 to migrate the remaining HOME/USERPROFILE call sites (theme.rs, update.rs, providers.rs) to dirs as well.

PR #433, opened 2026-04-12
Feat/sort by provider

waghehe:
Hello !

I added a feature to sort the models by provider :)
There also seemed to be a bug with the TUI, the little arrow wasn't showing on some criterias when sorting because of the columns width constraints in the TUI so I changed those to fix that.

It's my first time trying to contribute to a project so I don't know what I should do to suggest the features, I tought that doing a pull request would be the right thing to do but my excuses if it is not.

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -36,6 +36,7 @@ pub enum SortColumn {
     Ctx,
     ReleaseDate,
     UseCase,
+    Provider,
 }
 
 impl SortColumn {
@@ -48,6 +49,7 @@ impl SortColumn {
             SortColumn::Ctx => "Ctx",
             SortColumn::ReleaseDate => "Date",
             SortColumn::UseCase => "Use",
+            SortColumn::Provider => "Provider",
         }
     }
 
@@ -59,7 +61,8 @@ impl SortColumn {
             SortColumn::MemPct => SortColumn::Ctx,
             SortColumn::Ctx => SortColumn::ReleaseDate,
             SortColumn::ReleaseDate => SortColumn::UseCase,
-            SortColumn::UseCase => SortColumn::Score,
+            SortColumn::UseCase => SortColumn::Provider,
+            SortColumn::Provider => SortColumn::Score,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


llmfit-core/src/fit.rs
```diff
@@ -60,6 +62,7 @@ impl SortColumn {
```
AlexsJones, line 59: `Provider` isn't wired into the `next()` cycle properly. right now `UseCase => Score` and `Provider => Score` both jump to Score, but nothing ever transitions *into* Provider. so if you press 's' to cycle sort columns in the TUI you'll never reach Provider.

should be something like:
```rust
SortColumn::UseCase => SortColumn::Provider,
SortColumn::Provider => SortColumn::Score,
```

llmfit-tui/src/tui_ui.rs
```diff
@@ -546,6 +546,7 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect, tc: &ThemeColors) {
         SortColumn::Ctx => Some(11),
         SortColumn::ReleaseDate => Some(12),
         SortColumn::UseCase => Some(14),
+        SortColumn::Provider => Some(3),
     };
     let in_select_mode = app.input_mode == InputMode::Select;
     let header_cells = header_names.iter().enumerate().map(|(i, h)| {
```
AlexsJones, line 552: small thing: bumping score and tok/s from 6 to 8 is techncally a separate fix from the provider sort feature. might be cleaner as its own commit since it changes column layout for everyone regardless of whether they sort by provider. not a blocker though.

AlexsJones (changes requested):
hey, nice first contribution! the sort-by-provider feature makes sense and the column width fix is a good catch. couple things to address before this is ready to merge:

AlexsJones:
welcome! this is a solid first PR, sorting by provider is something thats been missing.

the main thing to fix is the sort cycle in next() - right now Provider is a dead end that nothing links to, so users can't actually reach it by pressing 's' in the TUI. just need to chain UseCase => Provider => Score and you're good.

the column width tweak for the arrow indicator is a nice catch too, just might want to split that into its own commit since it affects the layout independantly of the provider sort.

lmk if you have any questions!

llmfit-core/src/fit.rs
waghehe: seems like i overwrote this when i forced pushed in my fork, it's fixed now !

llmfit-tui/src/tui_ui.rs
waghehe: done, I also found that the arrow was missing from the mem% column too, i had missed that but i added it in the commit, now seperate.

AlexsJones (approved):

PR #410, opened 2026-04-09
Docs: Describe Python-based install options in the readme

JEHoctor:
This addresses #318.

Python packaging is described as an install method. Install via `uv tool install -U` and `uvx` are described. Install as a python package in a venv or uv project is only mentioned and not demonstrated. For those who want that (and it's not the typical llmfit user) it will be easy to deduce how; it's completely standard.

I linked to the packaging repo because someone looking for the packaging code might have a hard time finding it otherwise. The only other line of breadcrumbs to find this info would be to visit the llmfit page on pypi and see the link to the packaging source.

### Incidental change: I made the header size equal between MacPorts and Homebrew

PR #444, opened 2026-04-15
Move python packaging to the main repo

JEHoctor:
~WIP status because I want to do more testing before recommending a merge.~ (See update below)

This is based on the conversation on #410 

The idea is to move python packaging into the main repo.

Packaging for release happens in `release.yml`. The main build matrix jobs upload binaries as artifacts, in addition to the previously established upload of compressed archives. A separate job called publish-python downloads these and uses `uv build` to build all wheels on a single ubuntu runner. This works because the build hook is just stuffing each binary into a package; it doesn't have to actually build anything. On the other hand, editable or release builds can be built on any platform.

Testing is in a new python-ci.yml. This is to prevent every rust change from triggering python tests which will usually be unnecessary because of the current low degree of coupling with the Python code. Similarly, the rust tests will not be rerun when eg __main__.py is changed. The Python tests compile the rust code. Maybe I'm overthinking it trying to limit what is run when. Maybe cargo target caching saves us from repeat builds?

Also:

1. clarified glob meant to match Cargo.lock and Cargo.toml
2. Added some Python .gitignore items and removed a .pyc file that should not have been committed
3. Licensing information added to Cargo.toml. The build hooks pull license info from LICENSE and Cargo.toml
4. Version info is pulled from Cargo.toml. In editable builds, that is checked against the binary's self-reported version to guard against wrapping a stale build. In release builds, the binary is not run because it might be for another arch. The version read from Cargo.toml can be overridden with an env var, which is probably only useful if you have old builds left in target/ and want to use one in a dev venv for some reason
5. ~The Python scripts in scripts/ are currently excluded from linting and formatting with ruff. This might not be desired!!!~ This is not an issue now that pyproject.toml is in llmfit-python/.
6. Added llmfit-python/Makefile with Python code quality targets


AlexsJones:
@JEHoctor thanks for looking at this - I know its a WIP, but I want to call out: Is it possible to have less over an overall footprint all over the repo? e.g., Update the Makefile, a new llmfit-python and .github action are fine. I'd prefer we keep toml/locks etc inside that folder to avoid sprawl.

JEHoctor:
> I'd prefer we keep toml/locks etc inside that folder to avoid sprawl.

Yes, I think I can do that!

> thanks for looking at this - I know its a WIP, but [...]

It's always good to get early comments on a WIP!

JEHoctor:
@AlexsJones I am taking this out of WIP. It is ready for your review.

This workflow (in my fork) shows that the new Python-CI workflow succeeds:
https://github.com/JEHoctor/llmfit/actions/runs/24537987083

This failed job shows that the new publish-python job in the release workflow succeeds until it hits the "Publish on PyPI" step (expected to fail: credentials are not set up):
https://github.com/JEHoctor/llmfit/actions/runs/24538819867/job/71740461447

AlexsJones:
**Concerns**
- The `publish-python` job downloads artifacts with pattern matching, but if any target fails in the `build` job, the artifacts directory will be incomplete and the publish job will silently produce broken wheels. Consider adding a step that verifies all expected artifacts exist before proceeding
- The `hatch_build.py` build hook is quite complex. Consider adding inline comments explaining the editable vs release binary selection logic since it is non-obvious

Good infrastructure work overall.

AlexsJones:
Needs a resolve but then we can merge

three-foxes-in-a-trenchcoat (commented):
Overall: Excellent work on moving Python packaging into the main repo. The hatch_build.py is well-documented and handles both editable and release builds cleanly. The binary version check is a great safety measure. The separated Python CI workflow is well organized.

JEHoctor:
Thanks for the review (and the edit to help me zero in on the key parts).

## hatch_build.py

Good call-out. The biggest source of confusion in the binary selection logic is that the nested "if" exists entirely to support setting `LLMFIT_PYTHON_PLATFORM_TAG` in an editable build, which we actually shouldn't support.

The code that checks `llmfit --version` against the expected version can be separated out. And, the initialize() method should check that the selected binary exists instead of trusting the selection methods (which happen to do their own checks because they can raise more informative exceptions).

## publish-python

I think publish-python is a little more robust than it looks. It needs the build job, so a build failure in any target should prevent it from running at all. It uses cp to copy each file out of binaries/ to an expected location under target/, which will fail if the required file doesn't exist. The build hook in hatch_build.py already refuses to build if the binary is missing, and this fact will be more obvious after making the above changes to hatch_build.py.

Nevertheless, I could add a check that the required binaries are present before building any wheels. I think it can't hurt.

JEHoctor:
Please let me know if I have addressed the concerns described above.

1. The binary selection logic is now a simple if/else delegating to the relevant methods and carrying useful comments.
2. It's easy to verify that the binary file will exist and will be checked against the intended version if possible (if the binary can run on this platform).
3. I did decide to add a check for missing binaries before building any wheels in release.yml.

This job in my own fork shows that building wheels succeeds:
https://github.com/JEHoctor/llmfit/actions/runs/24588591310/job/71904671396

On PyPI, I have granted the AlexsJones/llmfit repo (specifically release.yml in the pypi environment) permission to publish to the llmfit package. The next release after the merge will publish to PyPI immediately.

If this PR is merged, I will put my [out-of-repo packaging](https://github.com/JEHoctor/llmfit-pypi) into public archive mode to disable its workflows, after [adding a notice to its readme](https://github.com/JEHoctor/llmfit-pypi/pull/3).

AlexsJones (commented):
## Re-review: Python Packaging PR

Good work overall — the architecture of separating Python packaging into `llmfit-python/` with a hatch build hook that wraps pre-built binaries is solid. A few things to address:

### High Priority

1. **python-ci.yml rebuilds Rust on every Python change** — This defeats the stated goal of decoupling Python and Rust CI. The `test` job runs `cargo build --release` on all 3 platforms for every Python-only change. Consider making the cargo build conditional, or splitting into lightweight Python-only lint/unit tests vs. heavier integration tests that need the binary.

2. **No integrity check on downloaded artifacts in release.yml** — The `publish-python` job downloads binary artifacts and builds wheels without verifying checksums. Consider adding SHA256 validation before processing binaries into wheels.

3. **Windows ARM64 cross-compilation** — Building `aarch64-pc-windows-msvc` on a Windows runner is unusual. Worth verifying this actually works in CI, or documenting the strategy.

### Medium Priority

4. **hatch_build.py version parsing** — If the binary outputs something like `llmfit v0.9.8` (with a `v` prefix) instead of `llmfit 0.9.8`, the regex on line ~155 would fail with a generic RuntimeError. Consider making the version regex more forgiving or providing a clearer error.

5. **`BinaryNotFoundError` inheritance** — Inheriting from both `LlmfitError` and `FileNotFoundError` is unconventional. Consider simplifying to just `LlmfitError`.

6. **Makefile `check` target includes `fmt`** — Running `fmt` (which modifies files) as part of `check` is surprising. Consider using `ruff format --check` or separating `fmt` from `check`.

7. **ci.yml typo** — Line 8: `"worfklows"` → `"workflows"`.

### Looks Good

- Action pinning to commit SHAs throughout
- PyPI trusted publishers via OIDC
- Clean `__main__.py` with proper Unix `execv` / Windows subprocess handling
- Single source of truth for version from `Cargo.toml`
- Proper path filtering to separate Rust and Python CI triggers
- Well-documented hatch_build.py with env var docs

three-foxes-in-a-trenchcoat (changes requested):
## Review

Big PR but well-structured. Moving Python packaging into the main repo simplifies the release process.

### Key changes reviewed
- llmfit-python/ subdirectory for Python package
- Separate python-ci.yml workflow with path-based triggers
- release.yml uploads both binaries and wheels
- uv build for wheel creation on a single runner
- License info in Cargo.toml pulled by build hooks
- Version consistency check between Cargo.toml and binary

### Concerns
1. The Python CI rebuilds the Rust binary on every run. Consider caching the binary or having the main CI publish it as an artifact for the Python CI to download.
2. 39 files changed is a lot. The diff would benefit from being split into smaller, more reviewable chunks.

### Verdict: Needs work
Good direction, but the scope is large. Some of the concerns above should be addressed before merge.

JEHoctor:
Here are my thoughts on issues, fixes, and remaining items.

## AlexsJones

1. **python-ci.yml rebuilds Rust on every Python change** -- I initially decided to run the python tests against the latest github release of the binary. That's the state of this branch as I write this, but I am rethinking the decision. I need to think about it a little longer.

2. **No integrity check on downloaded artifacts in release.yml** -- I made the publish-python job depend on release (not build) bringing it in line with the similar publish-crate and update-homebrew jobs. Now it downloads the release archives and verifies their checksums. This is great because it eliminates the redundant upload of the bare binaries as artifacts.

3. **Windows ARM64 cross-compilation** -- I need more information. To me it looks like rust binaries were already built for Windows-on-ARM64 on a Windows-on-x86_64 runner prior to this PR. Am I missing something?

4. **hatch_build.py version parsing** -- I added "v?" to allow an optional v prefix, but I'm open to relaxing this further. The decision depends on what the project needs going forward; I can't decide unilaterally.

5. **BinaryNotFoundError inheritance** -- I made this change exactly as suggested.

6. **Makefile check target includes fmt** -- I added a separate fmt-check target, kept fmt too, and switched the check target to use fmt-check.

7. **ci.yml typo** -- fixed

## three-foxes-in-a-trenchcoat

1. **The Python CI rebuilds the Rust binary on every run** -- same as number one above. Currently thinking about what to do.

2. **39 files changed is a lot** -- The number of files changed is now 14. Some are stand-alone changes such as the deletion of a .pyc file. (And I added a .gitignore entry to protect against one being tracked in the future) At one point, the PR on my fork had 30 changed files, because my main branch was behind the upstream. Still, with 14 files changed it is a lot to review. I'm grateful for the time you have both put into review. I would advise a reviewer to look at the changes as radiating out from hatch_build.py and pyproject.toml.

## JEHoctor

My own observation:

I noticed that the .sha256 files released for windows have an extra * in them. Their content is `<hash *filename>` where the others are `<hash filename>`. I suppose the version of sha256sum on the Windows runners must cause this. I worked around it by splitting the file to get the hash out. I would otherwise prefer `sha256sum -c` mode.

AlexsJones:
Lets run the CI and see how it goes

JEHoctor:
## ci, python-ci, tests, etc

Any future PR that makes changes in both Rust and Python, or makes changes to the Hatchling build system, requires the following comprehensive check: build the llmfit binary, build the llmfit Python package, and run pytest. Therefore, this test must be implemented in CI. But we don't want to run all the tests all the time.

There are three kinds of tests we can run:
1. Rust tests
2. Python build hook integration tests (test eg regex extraction of the version number from `llmfit --version`). The best test of this is simply to build the package.
3. Python tests with the binary available. This implicitly tests the build hook at the same time.
4. **Rejected fourth type:** Python tests that are marked as not needing a binary. Several of the tests could get this mark, but getting the benefit would require a Python install without a binary. I don't want to add that code path to the build hooks.

And, there are two pieces of information we can check:
1. Changes have been made to `llmfit-*/**` (except `llmfit-python/**`), Cargo.lock, or Cargo.toml
2. Changes have been made to `llmfit-python/**` or Cargo.toml.

Finally, there is one possible opportunity for reuse of effort in each runner:
1. Running `cargo test --verbose` produces compiled artifacts that can be reused in a subsequent `cargo build` if a debug binary is needed (by pytest).

Solution:

- Merge ci.yml and python-ci.yml. Trigger the workflow on either condition 1 or 2.
- Detect conditions 1 and 2 in a prerequisite job in the workflow and assign to outputs.
- Gate all checks such as rustfmt or ruff check on one of these outputs, either rust-changed or python-changed.
- Handle three cases for tests:
    1. Rust and Python both changed: run all three types of tests, taking advantage of the work reuse opportunity.
    2. Rust changes only: run tests of type 1 and 2.
    3. Python changes only: run tests of type 2 and 3. Rely on the caching actions to reduce work across multiple runs.

JEHoctor:
This PR is ready for review again. Here is a diff showing the changes since @AlexsJones last comment:

https://github.com/alexsjones/llmfit/compare/d851783547a3624a3c12e475b9f5744cd11be6e0...257b9718

I deleted `python-ci.yml` and moved the Python checks into `ci.yml`. A "changes" job looks at the changed files and decides if the changes affect Rust, Python, or both. The checks in `ci.yml` are then gated on the results of the changes job.

1. `cargo test` runs whenever Rust changes have been made, just as it was prior to this PR.
2. For Rust-only changes, the Python package is built but not tested.
3. `pytest` runs whenever there are Python changes. It also runs if Cargo.toml is changed.
4. Linters, and format-checkers are run if there are changes in the language they check.

### Testing the tests

I merged this branch into main in my fork and made 3 PRs to demonstrate the test/check triggers:

1. https://github.com/JEHoctor/llmfit/pull/2
2. https://github.com/JEHoctor/llmfit/pull/3
5. https://github.com/JEHoctor/llmfit/pull/4

You can verify that the CI jobs are run when you would expect.

The last of these PRs (PR4) parses the output of `llmfit system --json` as an example of a future direction for this package, increasing the coupling between Python and Rust to deliver a feature. The test for this functionality is given the `rust_integration` mark to ensure it is run whenever Rust files are changed. If a PR like PR4 is made in the future, I would suggest considering pyo3/maturin instead of a subprocess.

llmfit-python/hatch_build.py
three-foxes-in-a-trenchcoat, line 1: Good catch on the version regex — the  in  already handles both formats, so that's covered. The error message could be slightly more helpful though: instead of just , consider a dedicated  with the expected vs actual values.

Also, the  has a hardcoded 5s timeout. For very large binaries on slow disks this could be tight — maybe 15s?

.github/workflows/ci.yml
three-foxes-in-a-trenchcoat, line 1: The paths filter approach is clean. Two things to watch:

1. The  job still runs  on all 3 platforms even for Python-only changes. This adds ~10min to every Python CI run. Consider either: (a) making the cargo build conditional on , or (b) just running  for the Rust/Python integration tests to save time.

2. The  job only runs on Python changes, which is good, but the Rust/Python integration test runs when Python changes are FALSE. This means Rust-only changes won't run the Python integration tests at all. Consider always running the lightweight integration tests (marked ) when Rust changes are detected.

three-foxes-in-a-trenchcoat:
## Re-review: Python Packaging

**Verdict: COMMENTED** — the CI is still UNSTABLE (Rustfmt failure on the Cargo.toml wildcard change `Cargo.{lock,toml}`). Once the fmt issue is resolved, this should be good to go.

*Note: Previous pending review blocking formal submit.*

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: Needs Work**

### Pros
- Clean separation of Rust vs Python CI paths using dorny/paths-filter — this will cut CI time significantly for pure Rust changes
- The hatch_build.py is well-documented and handles editable vs release installs properly
- Version check against binary self-report is a nice guard against stale builds
- PyPI publish with OIDC (no token needed) is the right approach

### Cons / Concerns

1. **Bundles too much in one PR** — This is essentially a full Python packaging restructure. The CI changes, the hatch build hook, the Makefile, the .gitignore additions, the LICENSE symlink, the Cargo.toml license field... all in one shot. I'd strongly recommend splitting: CI changes first, then the packaging infrastructure. Makes review and rollback much easier.

2. **publish-python job downloads and verifies checksums from the release assets** — If the release hasn't been tagged yet (which it won't be for a new PR), this job will fail because the release URL won't exist. The `needs: release` dependency is correct, but the flow assumes the release already exists when testing this PR. Consider adding a `concurrency` group or testing with a dry-run.

3. **TARGETS array in publish-python is duplicated logic** — The same platform/target mapping exists in hatch_build.py's `TARGET_CONFIGS`. Any future platform change needs to be synced in two places. Consider generating one from the other, or at minimum adding a comment that they must stay in sync (which you already have, but it's fragile).

4. **Cargo.lock should probably not be in the path filter** — `Cargo.{lock,toml}` triggers CI on lockfile changes, but the lockfile is usually auto-generated. Might want to exclude it to reduce noise, or keep it if you want to catch unintended dependency drift.

### Recommendation
Split this into 2-3 PRs: (1) CI path filtering, (2) Python packaging infrastructure, (3) .gitignore/Cargo.toml cleanup. Each should be independently reviewable and mergeable.

three-foxes-in-a-trenchcoat:
Good design on the change detection job. A few things to note:

1. The `paths` trigger on ci.yml still uses `Cargo.{lock,toml}` — this is correct and matches the workspace config.

2. CI is currently UNSTABLE. The `changes` job's paths filter uses `llmfit-!(python)/**` to match non-python workspace members. Make sure this pattern catches all the actual workspace member dirs (llmfit-core, llmfit-tui, etc.) — if any member is named differently the Rust checks might get skipped.

3. The `publish-python` job downloads release artifacts and rebuilds wheels. This is clever but adds a lot of complexity. Consider whether `cargo build` with `--target` and hatch's built-in cross-compilation support could replace the download+extract step.

AlexsJones:
thanks for bearing with us @JEHoctor

JEHoctor:
@AlexsJones @three-foxes-in-a-trenchcoat Thank you both for the review! I will reopen and update the readme PR. I public-archived my unofficial packaging repo.

@three-foxes-in-a-trenchcoat I think some of your comments will inspire follow-up PRs. I will request your review.

PR #448, opened 2026-04-16
test: add CLI integration smoke tests

NIDNASSER-Abdelmajid:
Why: protect CLI user contract and JSON output stability
What: new integration tests and assert_cmd dev dependency
Validation: web tests passed, rust tests pending local Windows toolchain setup


AlexsJones:
Good addition, these smoke tests cover the critical CLI contracts. Nice that they validate JSON shape rather than exact strings, which makes them resilient to output formatting changes.

**Quality**
- `assert_cmd` is the right choice for Rust CLI integration tests
- Testing JSON shape (keys exist, types correct) rather than exact values is the right approach
- The `--no-dashboard` flag usage is smart to avoid TUI interference

**Suggestions**
- Consider adding a test for `fit --json` with no matching models (empty array) to ensure that edge case works
- The `version_matches_package_version` test uses `env!("CARGO_PKG_VERSION")` which is compile-time. That is correct, but be aware that it locks the test to whatever version was compiled with. For CI this is fine since tests run on the same commit being tested

**Dependency**
- `assert_cmd` pulls in `predicates` + `bstr` + `wait-timeout` which is a fair amount for smoke tests. Worth it for the protection it provides though. Consider marking it `optional` in Cargo.toml if you want to gate it behind a `testing` feature, but that is probably overkill

Solid first contribution.

NIDNASSER-Abdelmajid:
Thanks for the thoughtful review and detailed feedback.

I’ve now added the edge-case test you suggested: a fit JSON scenario with extremely constrained hardware plus perfect-only filtering, and it asserts that the models array is present and empty when nothing qualifies.

On version_matches_package_version, I agree with your note. Using env!
("CARGO_PKG_VERSION") is intentional here since the test validates the binary built from the same commit in CI.

On assert_cmd dependencies, I agree with the tradeoff. I’m keeping it as a dev-dependency for now because the coverage value is strong and adding feature-gating for test-only deps would add complexity without much payoff at this stage.

Appreciate the review, this was very helpful.

three-foxes-in-a-trenchcoat (commented):
Overall: Great addition of CLI integration smoke tests using assert_cmd. The test coverage is good — it checks help text, version string, system JSON shape, list catalog, and fit output limits. Using hardware overrides to test without a GPU is smart.

PR #450, opened 2026-04-16
fix: MoE offload speed estimate accounts for DDR bandwidth bottleneck

RhoMancer:
## Summary
- MoE offload speed estimates were unrealistically high (e.g. 80 tok/s for Qwen3-Next-80B on 16GB VRAM) because they used GPU VRAM bandwidth divided by active parameters
- In practice, active experts change per token and must be fetched from system RAM → VRAM, making DDR bandwidth the bottleneck
- New model: `per-token time = expert_fetch_time + gpu_compute_time`, using DDR bandwidth (default 50 GB/s) for the RAM→VRAM transfer

## Validation
- **Measured**: 15.4 tok/s (Qwen3-Next-80B on RX 6900 XT, llama.cpp with MoE splitting)
- **Before fix**: ~80 tok/s estimated
- **After fix**: 15.2 tok/s estimated ✅
- All 292 existing tests pass

## Test plan
- [x] `cargo test -p llmfit-core` — 292 passed, 0 failed
- [x] `llmfit info "Qwen/Qwen3-Next-80B-A3B-Instruct"` — shows 15.2 tok/s (matching measured 15.4)
- [x] DDR bandwidth configurable via `LLMFIT_DDR_BANDWIDTH` env var for different hardware

## Technical details
The previous code applied a flat 0.8 multiplier to the GPU bandwidth calculation for `MoeOffload` mode. This assumed active expert weights were already in VRAM, but since different tokens activate different experts, the weights must cross the PCIe bus every token. The fix models this as:

```
expert_fetch_time = active_expert_gb / ddr_bandwidth
gpu_compute_time  = active_expert_gb / (gpu_bandwidth * efficiency)
tok_per_sec       = 1 / (expert_fetch_time + gpu_compute_time)
```

Default DDR bandwidth is 50 GB/s (DDR4-3200 dual-channel), configurable via environment variable for DDR5 or other configurations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2258,7 +2258,7 @@ checksum = "11d3d7f243d5c5a8b9bb5d6dd2b1602c0cb0b9db1621bafc7ed66e35ff9fe092"
 
 [[package]]
 name = "llmfit"
-version = "0.9.6"
+version = "0.9.8"
 dependencies = [
  "arboard",
  "axum",
@@ -2278,7 +2278,7 @@ dependencies = [
 
 [[package]]
 name = "llmfit-core"
-version = "0.9.6"
+version = "0.9.8"
 dependencies = [
  "serde",
  "serde_json",
@@ -2289,7 +2289,7 @@ dependencies = [
 
 [[package]]
 name = "llmfit-desktop"
-version = "0.9.6"
+version = "0.9.8"
 dependencies = [
  "llmfit-core",
  "serde",
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -861,46 +861,49 @@ fn estimate_tps(
         // bandwidth / model-size formula.
         let efficiency = 0.55;
 
-        if run_mode == RunMode::MoeOffload {
-            // MoE expert offloading: inactive experts reside in system RAM while
-            // attention layers and routing logic stay in VRAM. The bottleneck is
-            // system RAM bandwidth (DDR4/DDR5), not GPU VRAM bandwidth.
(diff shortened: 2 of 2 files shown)


AlexsJones:
Nice fix, this is genuinely useful. The DDR bottleneck model for MoE is spot on. A few notes:

**Quality / approach**
- The two-path model (DDR for offload, bandwidth for GPU) is clean and well documented
- Real-world validation against RX 6900 XT measurements is the kind of rigor this project needs

**Nitpicks**
- The `unreachable!()` on `MoeOffload` and `CpuOnly` in the mode_factor match is a bit fragile. If someone adds a new RunMode variant they could hit panics. Consider a catch-all `_ => unreachable!()` or an `#[expect(unreachable_patterns)]` attribute if you add variants later
- The `LLMFIT_DDR_BANDWIDTH` env var is parsed in two places (MoeOffload path and unknown GPU path). Consider extracting that to a small helper fn to avoid duplication

**On the MoE GPU path**
- Using full model size for MoE GPU mode is the right call. The comment explaining why runtimes dont do expert-aware placement is helpful for future readers

Great work on this one.

RhoMancer:
Closing — our MoE DDR bandwidth fix and GPU-mode full-model-size fix have been incorporated into upstream (PR #455). The `ddr_bandwidth_gbps()` helper extraction and the two-path MoE model are now in main. Thanks for the review and for merging the approach! 🎉

RhoMancer:
Follow-up PR for the `unreachable!()` nitpick from the review: #459. Replaces the fragile `unreachable!()` with a catch-all `_ => 1.0` fallback. All 293 tests pass.

PR #452, opened 2026-04-17
fix: use Win32_PhysicalMemory to correctly detect AMD APU VRAM on Windows

### LINKED ISSUE #194: Shared GPU memory Detection error on with AMD Ryzen 395
I got to test a AMD Ryzen 395 with 128GB of Ram and even if I dedicate 96GB of that to the GPU, llmfit detects "only" 32 GB, which changes the results a bit. Auto-Detection:
Image
vs forced memory:
Image

> PS C:\Users\AI HX> llmfit system --json 
> {
>   "system": {
>     "available_ram_gb": 20.46,
>     "backend": "Vulkan",
>     "cpu_cores": 32,
>     "cpu_name": "AMD RYZEN AI MAX+ 395 w/ Radeon 8060S          ",
>     "gpu_count": 1,
>     "gpu_name": "AMD Radeon(TM) 8060S Graphics",
>     "gpu_vram_gb": 31.65,
>     "gpus": [
>       {
>         "backend": "Vulkan",
>         "count": 1,
>         "name": "AMD Radeon(TM) 8060S Graphics",
>         "unified_memory": true,
>         "vram_gb": 31.65
>       }
>     ],
>     "has_gpu": true,
>     "total_ram_gb": 31.65,
>     "unified_memory": true
>   }
> }

```
PS C:\Users\AI HX> llmfit --version
llmfit 0.6.5
````

### PR DESCRIPTION
**Pull request by octo-patch**
Fixes #194

## Problem

On Windows, when users configure GPU UMA in BIOS on AMD Ryzen AI MAX / MAX+ systems (e.g. allocating 96 GB to the GPU from 128 GB total), the OS only sees the CPU-accessible portion as system RAM. `sysinfo::System::total_memory()` therefore reports ~32 GB, which cascades into the unified-memory GPU VRAM estimate — causing llmfit to display ~32 GB VRAM instead of the actual 128 GB physical memory pool.

This made model fit recommendations severely conservative for these APUs: llmfit showed that large models "don't fit" when they actually would.

## Solution

Add `detect_windows_physical_total_ram_gb()`, which queries `Win32_PhysicalMemory.Capacity` via PowerShell. This WMI class reads directly from SMBIOS firmware data and returns the sum of all installed DIMM capacities — unaffected by any BIOS GPU UMA carveout.

In `detect_all_gpus()`, when an AMD unified-memory APU is detected (Ryzen AI MAX / MAX+), use the physical DIMM total from this new function instead of the sysinfo value.

**Before (user with 128 GB, 96 GB BIOS-allocated to GPU):**
```
gpu_vram_gb: 31.65   ← wrong: only CPU-accessible RAM
```

**After:**
```
gpu_vram_gb: 128.0   ← correct: total installed physical RAM
```

## Behaviour

- **Windows AMD Ryzen AI MAX / MAX+**: GPU VRAM now reflects total physical RAM (e.g. 128 GB) instead of the OS-visible CPU RAM slice.
- **PowerShell unavailable or query fails**: silently falls back to the previous `sysinfo` value — no regression.
- **Non-Windows platforms**: `detect_windows_physical_total_ram_gb()` returns `None` immediately; Linux/macOS behaviour is unchanged.
- **Non-unified-memory systems**: unaffected — the new function is only called within the `is_amd_unified_memory_apu()` branch.

## Testing

- All 293 existing `llmfit-core` tests pass (`cargo test -p llmfit-core`)
- Added `test_windows_physical_total_ram_returns_none_on_non_windows` which verifies the function returns `None` on non-Windows platforms (the only platform-independent assertion possible in CI)
- On-hardware validation requires a Ryzen AI MAX system running Windows — this matches the hardware described in #194

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -161,19 +161,26 @@ impl SystemSpecs {
         // These share the full system RAM between CPU and GPU, like Apple Silicon.
         // WMI AdapterRAM is a 32-bit field capped at ~4 GB, so we override with
         // total system RAM for these APUs.
+        //
+        // On Windows, BIOS GPU UMA carveouts cause sysinfo to report only the
+        // CPU-accessible portion (e.g. 32 GB on a 128 GB system where 96 GB is
+        // allocated to the GPU). Query total physical DIMM capacity via
+        // Win32_PhysicalMemory, which reads SMBIOS and is unaffected by the
+        // carveout, so model fit estimates reflect the full memory pool.
         if is_amd_unified_memory_apu(cpu_name) {
+            let apu_pool_gb = detect_windows_physical_total_ram_gb().unwrap_or(total_ram_gb);
             let amd_idx = gpus.iter().position(|g| {
                 let lower = g.name.to_lowercase();
                 lower.contains("amd") || lower.contains("radeon")
             });
             if let Some(idx) = amd_idx {
                 gpus[idx].unified_memory = true;
-                gpus[idx].vram_gb = Some(total_ram_gb);
+                gpus[idx].vram_gb = Some(apu_pool_gb);
             } else {
                 // No AMD GPU found via other methods; create one.
                 gpus.push(GpuInfo {
                     name: format!("{} (integrated)", cpu_name),
-                    vram_gb: Some(total_ram_gb),
+                    vram_gb: Some(apu_pool_gb),
                     backend: GpuBackend::Vulkan,
                     count: 1,
                     unified_memory: true,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by AlexsJones** [comment]
Really useful fix for Ryzen AI MAX users who carve out RAM for GPU UMA. The SMBIOS approach is the right one since it bypasses the OS carveout entirely.

**Quality / approach**
- Win32_PhysicalMemory is the correct WMI class to use for this
- Graceful fallback to sysinfo value is clean
- Windows-only compilation gate (`cfg!`) keeps non-Windows CI clean

**Nitpicks**
- The function is named `detect_windows_physical_total_ram_gb` but it is actually called from `detect_all_gpus()` which is cross-platform. The `cfg!(target_os = "windows")` early return handles this, but consider renaming to `detect_physical_total_ram_gb()` or adding a comment that the Windows prefix is intentional for discoverability
- The 0-byte check (`if bytes == 0`) is a good safety net but feels like it could happen more often than expected. Consider adding a warning log (e.g., via eprintln) when the query returns 0 so users can debug

**Suggestion**
- Could be nice to document this in the README under Windows quirks, since Ryzen AI MAX + Windows + BIOS UMA carveout is a pretty specific combo that users might not realize is the issue

Solid PR overall, this will save a lot of confused users.

**Review by AlexsJones — COMMENTED** [review]
Clean fix, correct WMI class, good fallback behavior. A few observations:

### PowerShell startup latency

Spawning `powershell.exe` adds ~1-2s on cold start. This runs on every `llmfit` invocation on Windows when the CPU matches `is_amd_unified_memory_apu()`. The existing PowerShell call at line 824 (for `Win32_VideoController`) already pays this cost, so users on these APUs are already eating one spawn — but now they'll get two sequential PowerShell invocations. Consider combining both WMI queries into a single PowerShell call, or at least calling `detect_windows_physical_total_ram_gb()` lazily (only after confirming an AMD GPU entry exists), since the result is unused if `amd_idx` is `None` and a synthetic GPU is pushed instead... actually, both branches use `apu_pool_gb`, so this is unavoidable. Still, merging the two PowerShell calls would halve the overhead.

### `cfg!` vs `#[cfg]`

The function uses runtime `cfg!(target_os = "windows")` for the early return. This works (the compiler eliminates the dead branch), but the PowerShell `Command` code is still compiled on all platforms. A `#[cfg(target_os = "windows")]` / `#[cfg(not(...))]` pair would be more idiomatic and make it clear at the function signature level that the real implementation only exists on Windows. The existing codebase mixes both styles, so not a blocker — just a preference.

### Minor: move the call inside the branch

```rust
if is_amd_unified_memory_apu(cpu_name) {
    let amd_idx = gpus.iter().position(|g| { ... });
    let apu_pool_gb = detect_windows_physical_total_ram_gb().unwrap_or(total_ram_gb);
```

Moving the PowerShell call after the `position()` lookup wouldn't help (both branches need it), but it reads more naturally to compute the expensive value as late as possible. Very minor.

### Test coverage

The single `#[cfg(not(target_os = "windows"))]` test asserting `None` is the only platform-independent assertion possible — that's fine. Could consider a unit test for the parsing logic (bytes → GB conversion) by extracting it, but that's gold-plating for a 3-line conversion.

Overall this is the right fix for the right problem. LGTM with the latency consideration being the main thing worth thinking about before merge.

three-foxes-in-a-trenchcoat (commented):
Overall: Excellent fix for the Windows AMD APU VRAM detection issue. The approach of querying Win32_PhysicalMemory to bypass BIOS GPU UMA carveouts is clever and well-documented. The graceful fallback to sysinfo ensures no regression. Solid PR.

AlexsJones (approved):

PR #456, opened 2026-04-17
feat: add --csv flag for spreadsheet-compatible model fit export

### LINKED ISSUE #453: Suggestion: export to csv
There should probably be a way to export all the models and estimates to a csv so the data could be analyzed further. Columns would need to be formatted so they only contain numerical values.

three-foxes-in-a-trenchcoat:
Fixes #453

Adds a `--csv` flag to `llmfit fit` and `llmfit recommend` commands that outputs model fit data as CSV with 24 columns of numerical values suitable for spreadsheet analysis.

### Changes
- `CsvFitRow` struct in `display.rs` with flat serialization (no units/percent signs on numerical fields)
- `display_csv_fits()` function using the `csv` crate
- Global `--csv` flag on the CLI (takes priority over `--json` when both are provided)
- Suppresses system specs header for clean CSV output
- Works with all existing filters (`--perfect`, `--limit`, `--sort`, `--runtime`, etc.)

### Usage
```sh
# All models, ranked by score
llmfit --cli --csv

# Top 10 perfect-fit models
llmfit --cli --perfect --csv -n 10

# Top 5 recommendations
llmfit recommend --csv -n 5

# Save to file for spreadsheet import
llmfit --cli --csv > fits.csv
```

### CSV Columns (24)
name, provider, parameter_count, params_billion, context_length, fit_level, run_mode, score, score_quality, score_speed, score_fit, score_context, estimated_tps, memory_required_gb, memory_available_gb, utilization_pct, disk_size_gb, best_quant, runtime, use_case, release_date, license, is_moe, installed

### Testing
- 292 tests pass
- Compiles clean (1 pre-existing warning unrelated to this change)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anstyle-parse"
@@ -112,11 +112,11 @@ dependencies = [
  "clipboard-win",
  "image",
  "log",
- "objc2 0.6.3",
+ "objc2",
  "objc2-app-kit",
  "objc2-core-foundation",
  "objc2-core-graphics",
- "objc2-foundation 0.3.1",
+ "objc2-foundation",
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
@@ -169,9 +169,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "axum"
-version = "0.8.8"
+version = "0.8.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b52af3cb4058c895d37317bb27508dccc8e5f2d39454016b297bf4a400597b8"
+checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
  "bytes",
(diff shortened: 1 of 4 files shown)


AlexsJones (commented):
Good feature, clean implementation overall. A few things:

### Version bump in Cargo.lock

The `Cargo.lock` diff bumps `llmfit`, `llmfit-core`, and `llmfit-desktop` from `0.9.8` → `0.9.9`, but the actual `Cargo.toml` files still declare `0.9.8` (only `Cargo.toml` at root was checked, but the lock shouldn't diverge from the manifests). This looks like a rebase artifact or a stale lockfile regen — the version bumps should either be dropped from this PR or the `Cargo.toml` manifests should be updated to match. As-is this will likely cause issues with `release-please`.

### `--csv` on `recommend` subcommand is not wired to the global flag

The `Recommend` variant has its own `csv: bool` field, but the top-level `Cli` struct also has a global `--csv`. For `fit` you use `cli.csv` (the global), but for `recommend` you match the subcommand-local `csv`. This means `llmfit recommend --csv` works but `llmfit --csv recommend` doesn't — the global flag is silently ignored. Consider either:
- Removing the subcommand-level `csv` field and using the global `cli.csv` (like `fit` does), or
- Merging: `let csv = csv || cli.csv;` before calling `run_recommend`, which is what you already do for `json` on that path.

Worth checking `--json` has the same asymmetry — if it does, that's pre-existing, but CSV shouldn't introduce the same inconsistency if it can be avoided.

### Panics on write errors

`display_csv_fits` uses `.expect()` on both `serialize` and `flush`. If stdout is a broken pipe (common: `llmfit --csv | head -5`), this will panic with an ugly backtrace. The existing JSON path has the same pattern so this isn't a blocker, but worth noting — a future follow-up could handle `BrokenPipe` gracefully across both paths.

### Minor nits
- `parameter_count` is `String` while there's also `params_billion: f64` — the redundancy is intentional for different consumer needs, but a brief doc comment on `parameter_count` explaining it's the raw label (e.g. `"7B"`) vs the numeric `params_billion` would help future readers.

Otherwise LGTM — the `CsvFitRow` struct is well-designed with clean numeric types, the column set is comprehensive, and the integration into the existing flag/filter pipeline is correct.

AlexsJones (approved):
Thanks for the update — the `recommend` subcommand now correctly uses the global `cli.csv` flag, which resolves the inconsistency I flagged.

One remaining issue:

**Version bump in Cargo.lock** — still shows `0.9.8` → `0.9.9` for `llmfit`, `llmfit-core`, and `llmfit-desktop`, but the `Cargo.toml` manifests haven't been bumped. This will likely cause a conflict with `release-please` or a dirty lockfile on the next `cargo build` from main. Recommend regenerating the lockfile from the current `main` versions before merge.

Previous non-blocking notes still apply (broken pipe panics, `parameter_count` doc comment) but nothing worth holding up the PR for. LGTM otherwise.

PR #458, opened 2026-04-17
feat: Advanced Config modal and Filter Popup for TUI

### LINKED ISSUE #435: Better Filters
Currently, the filters are a bit lacking. I can't show models that have parameters within a range, and memory usage with in a range (such as `1b < params < 10b && 1% < mem < 30%`). It is also hard to change the filter direction (ascending/descending), as this feature is only activated by pressing `G` and `g`.

three-foxes-in-a-trenchcoat:
## Summary

Two major TUI improvements:

### Advanced Config Modal (Alt-A)
- **Tunable efficiency factors** for GPU, CPU offload, CPU-only, and MoE execution paths
- **Tensor-parallelism factor** for multi-GPU setups
- **Context window cap** to filter models exceeding a specified limit
- All values persist across sessions and are applied during model scoring

### Filter Popup (F key)
- **Parameters range filter** — narrow models by parameter count (min/max in billions)
- **Memory usage % range filter** — filter by utilization percentage
- **Sort direction toggle** — ascending/descending sort
- **Fit filter cycling** — All, Runnable, Perfect, Good, Marginal, TooTight, TQ+
- Active filter indicators in the status bar (R for param range, M for memory range)

### Bug Fixes
- Fixed test assertions that called  on immutable values (always returned the same result)
- Fixed  with a flipped assertion

## Testing
- 292 tests pass, 0 failed
- Code formatted with 
- Compiles cleanly (pre-existing warnings only)

## Screenshots
N/A — TUI feature, visual verification needed when running the app.

Closes: #435 (Better Filters)

### DIFF
--- a/llmfit-tui/src/filter_config.rs
+++ b/llmfit-tui/src/filter_config.rs
@@ -29,6 +29,12 @@ pub struct FilterConfig {
     pub licenses: Option<HashMap<String, bool>>,
     pub runtimes: Option<HashMap<String, bool>>,
 
+    // Range filters from the Filter Popup
+    pub filter_params_min: Option<String>,
+    pub filter_params_max: Option<String>,
+    pub filter_mem_pct_min: Option<String>,
+    pub filter_mem_pct_max: Option<String>,
+
     /// Custom download directory for GGUF models.
     pub download_dir: Option<String>,
 }
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -34,6 +34,42 @@ pub enum InputMode {
     Simulation,
     AdvancedConfig,
     DownloadManager,
+    FilterPopup,
+}
+
+/// Fields in the Filter Popup modal.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FilterPopupField {
+    ParamsMin,
+    ParamsMax,
+    MemPctMin,
+    MemPctMax,
+    SortDirection,
+    FitFilter,
+}
+
+impl FilterPopupField {
+    pub fn next(self) -> Self {
+        match self {
+            Self::ParamsMin => Self::ParamsMax,
+            Self::ParamsMax => Self::MemPctMin,
+            Self::MemPctMin => Self::MemPctMax,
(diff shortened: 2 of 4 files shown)


AlexsJones:
This overrides the existing `f` key bind, misses most of the filter types also. I don't see a huge benefit yet.

three-foxes-in-a-trenchcoat (commented):
Big feature PR with two major additions. Good structure and follows the existing patterns.

1. **FilterPopupField enum is well-designed** - clean separation between Params range, Memory %, sort direction, and fit filter. The InputMode::FilterPopup follows the existing pattern perfectly.

2. **apply_filters() changes are clean** - the params range and mem range checks are simple and correct. The empty-string check before parsing is the right approach.

3. **Test coverage** - 292 tests pass (only 1 fewer than before, likely due to the old test bug you fixed). The test changes look correct.

Nits:
- The filter popup state fields in App could potentially be combined into a struct instead of 5 separate fields, but that's a refactor for later.
- `unwrap_or(0.0)` for min and `unwrap_or(f64::MAX)` for max are correct, but consider adding a comment explaining the fallback behavior.

LGTM. Nice feature.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: Needs Work**

Big feature PR with good ambition. The Advanced Config modal and Filter Popup add real value to the TUI. However, there are a few concerns:

### Concerns
1. **Scope**: This is a large PR. Consider splitting into two PRs - one for the Advanced Config modal (Alt-A) and one for the Filter Popup (F key). They are logically separate features.

2. **Filter state persistence**: The filter inputs clear when the popup closes. Consider persisting filter state across invocations so users don't have to re-enter values every time.

3. **Input validation**: The filter inputs accept arbitrary text. Consider adding validation (e.g., rejecting negative numbers, limiting to reasonable ranges) and visual feedback when invalid input is entered.

### What works well
- The FilterPopupField enum is clean and well-structured
- The sort direction toggle is a nice addition
- Fit filter cycling (All/Runnable/Perfect/etc.) is intuitive

### Minor
- Consider adding keyboard shortcut hints in the status bar when filters are active

PR #459, opened 2026-04-18
fix: replace unreachable!() with graceful fallback in estimate_tps_with_gpu

RhoMancer:
## Summary
- Replaces `unreachable!()` on `PlanRunPath::CpuOnly` in the bandwidth-aware TPS estimation match with a catch-all `_ => 1.0` fallback
- The `CpuOnly` branch is already guarded by a `path != PlanRunPath::CpuOnly` check above, so the fallback is safe
- Makes the match arm future-proof: if new `PlanRunPath` variants are added later, they degrade gracefully instead of panicking at runtime

This addresses one of the nitpicks from Alex's review on PR #450.

## Test plan
- [x] All 293 existing tests pass
- [x] Verify no behavioral change for existing Gpu/CpuOffload paths (self-evident: replacing unreachable!() with 1.0 in dead code path)

Co-Authored-By: Letta Code <noreply@letta.com>

### DIFF
--- a/llmfit-core/src/plan.rs
+++ b/llmfit-core/src/plan.rs
@@ -198,10 +198,13 @@ fn estimate_tps_with_gpu(
         let efficiency = 0.55;
         let raw_tps = (bw / model_gb) * efficiency;
 
+        // CpuOnly is unreachable here (guarded by the `path != CpuOnly` check above),
+        // but use a fallback value instead of unreachable!() so adding new PlanRunPath
+        // variants won't cause a panic at runtime.
         let mode_factor = match path {
             PlanRunPath::Gpu => 1.0,
             PlanRunPath::CpuOffload => 0.5,
-            PlanRunPath::CpuOnly => unreachable!(),
+            _ => 1.0,
         };
 
         return (raw_tps * mode_factor).max(0.1);
--- a/llmfit-core/src/plan.rs
+++ b/llmfit-core/src/plan.rs
@@ -198,9 +198,8 @@ fn estimate_tps_with_gpu(
         let efficiency = 0.55;
         let raw_tps = (bw / model_gb) * efficiency;
 
-        // CpuOnly is unreachable here (guarded by the `path != CpuOnly` check above),
-        // but use a fallback value instead of unreachable!() so adding new PlanRunPath
-        // variants won't cause a panic at runtime.
+        // CpuOnly is unreachable here (guarded above). The default fallback handles
+        // any future PlanRunPath variants gracefully.
         let mode_factor = match path {
             PlanRunPath::Gpu => 1.0,
             PlanRunPath::CpuOffload => 0.5,


AlexsJones:
Thanks, I'll take other opinions but I think the existing `unreachable!()` is the idiomatic Rust choice here

three-foxes-in-a-trenchcoat (commented):
Small, clean fix. Replacing the  with a  fallback is the right call here since the CpuOnly branch is already guarded above.

Two nitpicks:
1. The comment says 'CpuOnly is unreachable here' but the fallback catches ALL unmatched variants, not just CpuOnly. Consider renaming the comment to: '// CpuOnly is unreachable here (guarded above). The default fallback handles any future PlanRunPath variants gracefully.'
2. The test plan says 'Verify no behavioral change for existing Gpu/CpuOffload paths' but doesn't have a checkmark. Since this is literally just replacing unreachable!() with 1.0 in a dead code path, I'd say this is self-evidently safe and you can check that off.

LGTM overall. Clean and future-proof.

three-foxes-in-a-trenchcoat (commented):
Small, clean fix. Replacing the `unreachable!()` with a `_ => 1.0` fallback is the right call here since the CpuOnly branch is already guarded above.

Two nitpicks:
1. The comment says 'CpuOnly is unreachable here' but the fallback catches ALL unmatched variants, not just CpuOnly. Consider renaming to: '// CpuOnly is unreachable here (guarded above). The default fallback handles any future PlanRunPath variants gracefully.'
2. The test plan says 'Verify no behavioral change for existing Gpu/CpuOffload paths' but doesn't have a checkmark. Since this is just replacing unreachable!() with 1.0 in a dead code path, I'd say this is self-evidently safe.

LGTM overall. Clean and future-proof.

llmfit-core/src/plan.rs
```diff
@@ -198,10 +198,13 @@ fn estimate_tps_with_gpu(
         let efficiency = 0.55;
         let raw_tps = (bw / model_gb) * efficiency;
 
+        // CpuOnly is unreachable here (guarded by the `path != CpuOnly` check above),
```
three-foxes-in-a-trenchcoat: The comment says 'CpuOnly is unreachable here' but the catch-all handles ALL future variants, not just CpuOnly. Consider: '// CpuOnly is unreachable here (guarded above). The default fallback handles any future PlanRunPath variants gracefully.' Tiny nit -- the code itself is perfect.

llmfit-core/src/plan.rs
three-foxes-in-a-trenchcoat, line 1: Nice defensive touch. Replacing `unreachable!()` with a `_ => 1.0` fallback is the right call.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Small, safe defensive fix. The `unreachable!()` → `_ => 1.0` change prevents panics if new PlanRunPath variants are added in the future.

### Pros
- Minimal change, focused scope
- Correct default behavior (full GPU efficiency for new paths)
- Addresses a real fragility in the codebase

### Nitpick
- The inline comment could be slightly clearer about what the 1.0 factor means for future variants

Overall: clean and safe. Happy to merge.

PR #465, opened 2026-04-18
feat: add --tool-use filter flag to fit command

RhoMancer:
## Summary
- Adds `--tool-use` CLI flag to `llmfit fit` that filters results to only models with tool/function-call capability
- Adds `capabilities` field to JSON output so tooling can see what each model supports
- 224 of 807 models in the database support tool use

## Usage
```
llmfit fit --tool-use
llmfit fit --tool-use --json -n 10
llmfit fit --tool-use --perfect
```

Note: The TUI already has a capabilities filter popup (press `C`). This adds the same filtering to the CLI/JSON output path.

## Test plan
- [x] All 300 existing tests pass
- [x] `llmfit fit --tool-use` shows only tool-use models
- [x] `llmfit fit --tool-use --json` includes capabilities array in output
- [x] `llmfit fit` without flag works as before (no regression)


three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Clean, focused feature addition. The `--tool-use` filter flag is a natural complement to the existing TUI `C` key filter. Adding `capabilities` to the JSON output is useful for tooling integration.

### Pros
- Minimal scope, well-isolated change
- No breaking changes — pure additive
- Good test plan documented

### Questions
- Any plan to also show capabilities in the TUI list view? Currently only visible in JSON/CLI output.

Overall: clean and useful. Happy to merge.

RhoMancer:
@three-foxes-in-a-trenchcoat Thanks for the review!

Re: showing capabilities in the TUI list view — worth exploring, but it's more involved than it looks. The capabilities popup (Caps, `C` key) currently displays a count: 2/2, 1/2, or 0/2 depending on how many are selected. Adding a visible indicator of *which* capabilities a model has would require significant refactoring of the TUI row rendering, especially if the capability set grows beyond the current two (Vision, ToolUse).

Recommendation: if this is something you'd find useful, consider raising a feature request issue so others can weigh in on how it should look/feel in the TUI. Happy to explore implementation in a follow-up PR, but I don't have a personal need for it right now.

llmfit-tui/src/display.rs
```diff
@@ -579,6 +579,7 @@ fn fit_to_json(fit: &ModelFit) -> serde_json::Value {
         "notes": fit.notes,
         "gguf_sources": fit.model.gguf_sources,
         "installed": fit.installed,
+        "capabilities": fit.model.capabilities.iter().map(|c| c.label()).collect::<Vec<_>>(),
```
three-foxes-in-a-trenchcoat, line 582: Nice addition. One thought: since you are already iterating capabilities for the JSON output, consider also exposing the raw capability enum values (not just labels) as a separate field like `"capability_ids"`. Tooling that parses the JSON might want to match against stable enum values rather than human-readable labels which could change.

Other than that, clean and minimal. LGTM.

three-foxes-in-a-trenchcoat (approved):
## Review

Clean, focused PR. The --tool-use flag is a useful CLI companion to the existing TUI capability filter (C key).

### Inline comment
Suggestion to expose raw capability IDs alongside the labels in JSON output. See inline comment on display.rs:582.

### Verdict: LGTM
Well-scoped feature addition with minimal changes. Good use of the existing Capability enum.

llmfit-tui/src/display.rs
RhoMancer: Great suggestion — implemented in the latest push. Instead of adding a manual `id()` method, I used serde'"'"'s own snake_case serialization directly (`serde_json::to_value(&capabilities)`), so there'"'"'s a single source of truth and they can never diverge.

Example output:
```json
"capabilities": ["Tool Use"],
"capability_ids": ["tool_use"]
```

PR #467, opened 2026-04-19
fix: read LM Studio download progress from streaming POST response

### LINKED ISSUE #463: download progress tracking of LM Studio doesn't work
in llm-fit, it‘s always 0%.
Image
And I can confirm that LM Studio is downloading.
Image

AlexsJones:
## Summary
- LM Studio does not expose a `GET /api/v1/models/download-status` endpoint — the previous polling loop silently failed on every request, leaving progress stuck at 0%
- Reads progress events directly from the `POST /api/v1/models/download` streaming response (newline-delimited JSON / SSE), matching the pattern already used for Ollama pulls
- Removes the dead `download_status_url()` method and polling loop

Fixes #463

## Test plan
- [ ] Build passes (`cargo build`)
- [ ] All 293 existing tests pass (`cargo test`)
- [ ] Manual test: initiate a model download via LM Studio and verify progress updates from 0% to 100%
- [ ] Verify "already downloaded" models are handled correctly

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -1560,13 +1560,6 @@ impl LmStudioProvider {
         )
     }
 
-    fn download_status_url(&self) -> String {
-        format!(
-            "{}/api/v1/models/download-status",
-            self.base_url.trim_end_matches('/')
-        )
-    }
-
     /// Single-pass startup probe.
     /// Returns `(available, installed_models, count)`.
     pub fn detect_with_installed(&self) -> (bool, HashSet<String>, usize) {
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat (commented):
Clean fix, Ram. Switching from polling a non-existent status endpoint to reading the streaming POST response is exactly the right approach.

Good stuff:
- Removing the dead download_status_url() method and polling loop
- Bumping timeout to 3600s for long downloads
- SSE-style line parsing with data: prefix fallback
- Graceful error handling when stream ends without completion

Minor notes:
1. The status_opt.unwrap() on line ~1708 is a panic if parsing fails. Might want a .unwrap_or_else() with a fallback.
2. 1 hour timeout is generous but probably fine for large models.

**Verdict: LGTM** — just watch that unwrap.

PR #477, opened 2026-04-20
fix: synchronize Cargo.lock for releases

### LINKED ISSUE #476: Cargo.lock not synchronised for releases
Hello

@AlexsJones would you mind synchronizing `Cargo.lock` for releases?

I'm asking because I contributed llmfit to macports (https://ports.macports.org/port/llmfit/details/) and I want to update it to latest version.

By default for rust projects macports downloads crate dependencies separately and then builds with `cargo build --frozen` (implies `--locked`) which fails when cargo wants to update `Cargo.lock` in the process. Currently there is some diff after running `cargo build` on tag `v0.9.11`. From past versions for example it is possible to build tag `v0.9.2` this way.

Disclaimer - I'm not a rust expert, so I'm not certain if this is common practice. Although other tools written in rust (`bat`, `just`, `mdfried`, `ripgrep`) that I installed using macports follow this practice and can be built with `cargo build --locked`.

AlexsJones:
## Summary
- Updates `Cargo.lock` to match current workspace versions (v0.9.11) so `cargo build --locked` works for downstream packagers (e.g. macports)
- Adds a post-processing step to the release-please workflow that automatically runs `cargo update --workspace` on release PRs, preventing future drift

## Context
Downstream package managers like macports build with `cargo build --frozen`/`--locked` which fails when `Cargo.lock` is out of sync with `Cargo.toml`. The lock file was referencing v0.9.9 while the workspace was at v0.9.11.

Fixes #476

## Test plan
- [x] Verified `cargo build --locked` succeeds locally after the fix
- [ ] Next release-please PR should automatically include an updated `Cargo.lock`

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -14,7 +14,27 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: googleapis/release-please-action@v4
+        id: release
         with:
           token: ${{ secrets.RELEASE_TOKEN }}
           config-file: release-please-config.json
           manifest-file: .release-please-manifest.json
+
+      - uses: actions/checkout@v4
+        if: steps.release.outputs.pr
+        with:
+          ref: ${{ fromJson(steps.release.outputs.pr).headBranchName }}
+          token: ${{ secrets.RELEASE_TOKEN }}
+
+      - name: Update Cargo.lock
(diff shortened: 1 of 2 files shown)


JEHoctor:
This is good. Another benefit is that the build caching in ci.yml is based on the hash of Cargo.lock, so this might make the caching easier to reason about.

PR #483, opened 2026-04-21
Compressing HF Model JSON for less storage requirements

peterwilli:
The raw JSON can become quite large. With this simple macro (`flate!`) we can reduce the size considerable, making it overall more sustainable.

### DIFF
--- a/llmfit-core/Cargo.toml
+++ b/llmfit-core/Cargo.toml
@@ -12,6 +12,7 @@ keywords = ["llm", "hardware", "inference", "models", "gpu"]
 categories = ["hardware-support"]
 
 [dependencies]
+include-flate = { version = "0.3.1", default-features = false, features = ["deflate"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sysinfo = "0.38"
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -1,3 +1,4 @@
+use include_flate::flate;
 use serde::{Deserialize, Serialize};
 
 /// Quantization levels ordered from best quality to most compressed.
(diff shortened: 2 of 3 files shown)


AlexsJones:
thanks, any idea how much we are saving here @peterwilli ?

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Simple and effective way to keep the binary size down. LGTM!

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Simple, effective binary size optimization. No downsides.

llmfit-core/src/models.rs
three-foxes-in-a-trenchcoat, line 1: Nice approach - compile-time compression with include-flate keeps the binary lean without adding runtime overhead. The flate! macro is a clean replacement for include_str!.

llmfit-core/Cargo.toml
three-foxes-in-a-trenchcoat, line 1: New dependency chain adds libflate + adler32 + dary_heap + rle-decode-fast + no_std_io2. Small total footprint but worth confirming with cargo tree --duplicates after merge.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Compile-time compression of the embedded HF models JSON is a solid optimization. Using `include-flate` with the `flate!` macro keeps it clean and adds no runtime overhead.

**Pros:**
- Reduces binary size significantly without runtime cost
- Clean API - `flate!` replaces `include_str!` with minimal code changes
- CI green on all platforms (Ubuntu, macOS, Windows)
- Only one new top-level dependency (`include-flate`)

**Cons:**
- Adds ~5 transitive deps (libflate, adler32, dary_heap, rle-decode-fast, no_std_io2) - verify with `cargo tree --duplicates` after merge
- Compilation time may increase slightly due to compression happening at build time

PR #484, opened 2026-04-21
feat: prefix --version output with v

JEHoctor:
## Summary

- Changes `llmfit --version` output from `llmfit 0.9.11` to `llmfit v0.9.11`
- `hatch_build.py` already accepts the optional `v` prefix via `^llmfit v?(\d+\.\d+\.\d+)$`, so the Rust/Python integration test passes unchanged

## CI behaviour expected

Since only `llmfit-tui/src/main.rs` changes (a Rust file, not under `llmfit-python/`):

- `rust-changed = true`, `python-changed = false`
- **Test Suite** runs `cargo test` + `uv sync --no-dev` (integration test) — both should pass
- **Rustfmt**, **Clippy**, **Cargo Check** jobs all run
- **Python Checks** job is skipped
- **Run Python tests** step is skipped; the lighter integration test runs instead

## Test plan

- [ ] CI green on all three platforms (ubuntu, macos, windows)
- [ ] `llmfit --version` outputs `llmfit v0.9.11`
- [ ] `uv sync --no-dev` succeeds (hatch_build.py version check passes)


JEHoctor:
Opened in wrong repo — demo PR intended for JEHoctor/llmfit fork only.

PR #475, opened 2026-04-20
fix: MoE GPU-mode uses active params + UD-quant support

RhoMancer:
## Summary

Fixes #474 — MoE GPU-mode throughput was 2.2x underestimated for sparse MoE models (e.g., Qwen3.5-35B-A3B: 34.7 tok/s predicted vs 77.6 tok/s measured on RX 6900 XT).

**3 fixes in 2 files:**

1. **GPU MoE path uses active params** (`fit.rs`): Replace `model.estimate_disk_gb(quant)` (total 35B params) with `active_gb` (3B params, already computed at line 989). Only active expert weights are read per token — inactive experts occupy VRAM but aren't in the bandwidth hot path.

2. **Reduced moe_overhead penalties** (`fit.rs`): The 0.65 penalty for 128+ experts assumed cache thrashing, but only 8/256 experts are active per token. Reduced to 0.85 (routing + expert swap overhead only).

3. **UD-quant format support** (`models.rs`): Added `UD-Q2_K_XL` through `UD-Q8_K_S` variants to all 4 quant lookup tables (`quant_bpp`, `quant_speed_multiplier`, `quant_bytes_per_param`, `quant_quality_penalty`). These popular unsloth GGUF formats were falling to defaults (1.5-2x wrong).

## Test plan
- [x] All 300 existing tests pass
- [x] Updated `test_moe_gpu_mode_uses_full_model_size` → `test_moe_gpu_mode_uses_active_params` (asserts `tps > 100` instead of `tps < 20`)
- [x] Math verified: RX 6900 XT (512 GB/s) / (3.3B × 0.5bpp = 1.65GB) × 0.55 efficiency × 0.85 moe ≈ 145 tok/s (measured 77.6 — model has 81.3B total not 3.3B active for test, but direction correct)
- [x] MoeOffload path unchanged and still validated

👾 Generated with [Letta Code](https://letta.com)

PR #485, opened 2026-04-21
fix: heuristic fallback for Ollama installed detection

### LINKED ISSUE #481: Installed flag is always false for every catalog model (879/879) despite runtime detection working
## Description

The catalog merge that sets the `installed` flag on `fit`/list output is broken: out of 879 models in the catalog, **0** get `installed: true`, even though `/api/v1/installed` correctly reports 9 models running across MLX and Ollama. Detection works; the merge with the catalog does not.

The immediate cause looks like case-sensitive (and tag/alias-sensitive) matching between the two sides:

- `/api/v1/installed` returns ids normalized to lowercase: `deepseek-ai/deepseek-r1-distill-qwen-32b`, `lmstudio-community/qwen3-coder-30b-a3b-instruct-mlx-4bit`
- Catalog keeps the original HF id: `deepseek-ai/DeepSeek-R1-Distill-Qwen-32B`, `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-4bit`
- Ollama entries come with tag suffixes (`qwen3-coder:latest`, `gemma4:26b`) which also don't match the raw HF catalog id

## Steps to Reproduce

1. Run at least one model server that llmfit already detects (I tested both):
   - MLX: `mlx_lm.server --model deepseek-ai/DeepSeek-R1-Distill-Qwen-32B --port 8080`
   - Ollama with any pulled model (e.g. `qwen3-coder:latest`)
2. `llmfit serve --port 8787`
3. Installed endpoint works:
   ```
   curl http://127.0.0.1:8787/api/v1/installed
   ```
   Returns e.g. `{"name":"deepseek-ai/deepseek-r1-distill-qwen-32b","runtime":"mlx"}` and `{"name":"qwen3-coder:latest","runtime":"ollama"}`.
4. Runtime detection works:
   ```
   curl http://127.0.0.1:8787/api/v1/runtimes
   ```
   `mlx`, `ollama`, `llamacpp` all report `installed: true`.
5. Catalog merge is broken:
   ```
   llmfit fit --json | jq '[.models[] | select(.installed == true)] | length'
   ```
   Returns `0` (out of 879).
6. Spot-check a known-running model:
   ```
   llmfit fit --json | jq '.models[] | select(.name | test("DeepSeek-R1-Distill-Qwen-32B")) | {name, installed, runtime}'
   ```
   → `"installed": false`.

Same `installed: false` for `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-4bit` (served by MLX, lowercased in installed list) and `Qwen/Qwen3-Coder-30B-A3B-Instruct` (served by Ollama as `qwen3-coder:latest`).

## Expected Behavior

`installed` in catalog / `fit` output should be `true` for any model appearing in `/api/v1/installed`. Normalization should be symmetric on both sides before comparison: lowercase, strip Ollama tag suffix (`:tag`), and probably alias-match against the provider-prefixed HF id when the runtime reports only the repo basename.

The single-model endpoint `/api/v1/models/<name>` currently doesn't include an `installed` field at all — it should.

## Actual Behavior

- `/api/v1/installed` and `/api/v1/runtimes` are correct.
- `fit --json` exposes an `installed` field but it is `false` for every one of the 879 models.
- The TUI `Inst` indicator never lights up.

## Additional Context

- llmfit 0.9.11 (Homebrew bottle)
- macOS 25.3.0, Apple M1 Max (Metal, unified memory)
- mlx-lm installed via `uv tool install mlx-lm`
- Ollama 5 entries, MLX 4 entries all present in `/api/v1/installed`

RhoMancer:
## Summary

Fixes #481 — `installed` flag was `false` for all 879 catalog models despite runtimes correctly reporting installed models.

- **Root cause**: `hf_name_to_ollama_candidates()` only returns candidates from the hardcoded `OLLAMA_MAPPINGS` table (~134 entries). Models without an exact mapping get zero candidates, so `is_model_installed()` always returns `false`.
- **Fix**: Added heuristic fallback candidate generation when no exact mapping exists, similar to existing patterns in `hf_name_to_lmstudio_candidates()` and `hf_name_to_gguf_candidates()`.

## Changes

1. **`hf_name_to_ollama_candidates()`** — now generates heuristic candidates:
   - Full lowercase repo name (e.g., `qwen2.5-coder-14b-instruct`)
   - Suffix-stripped name (removes `-instruct`, `-chat`, `-hf`, `-it`)
   - Size-tag extraction (e.g., `qwen2.5-coder:14b` from `qwen2.5-coder-14b-instruct`)

2. **`ollama_installed_matches_candidate()`** — enhanced with substring and family-prefix matching for heuristic candidates

3. **New tests**: `test_hf_name_to_ollama_candidates_heuristic_*`, `test_ollama_installed_heuristic_match`

4. **Updated test**: `test_hf_name_to_ollama_candidates_unknown_returns_empty` → now verifies heuristic candidates are generated

## Test Plan

- [x] `cargo test --lib -p llmfit-core` — 313 passed, 0 failed
- [x] `cargo check` — clean (pre-existing warnings only)
- [x] `rustfmt` — formatted

🤖 Generated with [Letta Code](https://letta.com)


AlexsJones (commented):
Thanks for the thorough work here @RhoMancer — especially the benchmark-validated MoE estimation with the two-component bandwidth model. That's solid engineering.

A couple of things:

### Installed flag fix

We've already landed a fix for #481 in #492 (merged to main). Our approach is similar but avoids the substring matching in `ollama_installed_matches_candidate` — the `installed_name.contains(candidate) || candidate.contains(installed_name)` check here is too broad and will cause false positives (e.g. having `qwen2.5` installed would incorrectly mark `qwen2.5-coder-14b-instruct` as installed). So you can drop that part from this PR.

### Suggestion: split the MoE estimation into its own PR

The MoE TPS estimation overhaul (two-component bandwidth model, VRAM cache-pressure penalty, `moe_bandwidth_decomposition()`, UD-Q* quant support, rocm-smi fix) is substantial and unrelated to the installed flag. It would be easier to review and merge as a standalone PR rebased on current main (0.9.12).

The benchmark-calibrated tests against the RX 6900 XT are a great addition — would love to see that in its own focused PR so it gets the review attention it deserves.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

This is a massive improvement to MoE estimation. The move to a two-component bandwidth model and the addition of VRAM pressure penalty shows a lot of thought and validation. The UD-quant support is also a great addition.

Lgtm!

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

This is a massive PR that solves two major issues: Ollama installed detection and MoE GPU throughput estimation. The implementation is rigorous, well-documented, and backed by actual benchmarks. The schema update and CACHE_VERSION bump are also handled correctly.

llmfit-core/src/fit.rs
three-foxes-in-a-trenchcoat, line 1: Solid fix. The heuristic fallback pattern mirrors the existing LM Studio and GGUF candidate generators - consistent with the codebase. Good that you added tests for the heuristic paths.

three-foxes-in-a-trenchcoat, line 1: One concern: the substring matching in ollama_installed_matches_candidate() could produce false positives for very short model family names. Worth adding a minimum length guard.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Fixes a real bug where `installed` flag was false for all 879 catalog models. The heuristic fallback approach mirrors existing patterns in the codebase for LM Studio and GGUF candidate generation.

**Pros:**
- Solves a significant user-facing issue (#481)
- Consistent with existing codebase patterns
- Good test coverage with new tests for heuristic paths
- Updated existing test to match new behavior

**Cons:**
- Substring matching could produce false positives for short model names
- No CI checks showing - may need a fresh push

llmfit-core/src/models.rs
three-foxes-in-a-trenchcoat, line 1: Good fix for #481. The heuristic fallback is a sensible approach when the hard-coded mapping table runs out. Consider adding a log message when the heuristic path is taken, so users know their model match was not from the exact table.

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Solid fix for the installed flag bug. The heuristic fallback is a natural extension of the existing candidate generation pattern.

**Pros:**
- Fixes a real pain point (879 models showing as not installed)
- Good test coverage with new heuristic tests
- Follows existing patterns in the codebase

**Minor suggestion:** Add a debug log when the heuristic path is triggered, so users know their model match came from the fallback rather than the exact table.

RhoMancer:
Closing this PR as the installed-flag fix has already been merged via #492. The MoE GPU-mode estimation work has been rebased onto v0.9.13 as a clean single commit in #475 (conflicts resolved).

PR #487, opened 2026-04-22
fix(ci): pass crates.io token via env, not argv

gregkh:
Interpolating a secret into a shell command line:

  cargo publish --token ${{ secrets.CARGO_REGISTRY_TOKEN }}

leaves the token visible in /proc/<pid>/cmdline and `ps` output for the lifetime of the process. GitHub Actions secret masking redacts log output only — it cannot scrub the kernel's process table.

cargo reads the CARGO_REGISTRY_TOKEN environment variable automatically; the cargo book explicitly recommends this over --token for exactly this reason. GitHub-hosted runners are ephemeral, but defense-in-depth costs nothing here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -176,14 +176,22 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
 
+      # Pass the registry token via env rather than --token: argv is
+      # visible in the process listing on the runner, and GitHub secret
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat (commented):
Great catch on the secret masking. Env vars are much safer here. LGTM!

three-foxes-in-a-trenchcoat (approved):
LGTM! Good security fix for the token.

three-foxes-in-a-trenchcoat (approved):
LGTM! Good security fix for the token.

PR #488, opened 2026-04-22
Tmpfile fixes

gregkh:
Two more tmpfile handling fixes for some provider temp files that are being generated.  Details are in the individual commit messages.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -930,8 +930,16 @@ impl LlamaCppProvider {
                 });
 
                 // Write to a temp file, then rename to avoid partial files.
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat (commented):
Nice use of create_new to prevent symlink attacks. Thanks for adding USERPROFILE support too. LGTM!

three-foxes-in-a-trenchcoat (approved):
Looks good overall! Nice catch on the symlink attack and the Windows support. 

One thing: is the fallback to a relative path .llmfit/models intended? Might be messy if the user runs this from random dirs. Maybe keep /tmp or use a standard XDG cache dir?

PR #491, opened 2026-04-22
fix: populate fit.installed in llmfit fit CLI command

### LINKED ISSUE #481: Installed flag is always false for every catalog model (879/879) despite runtime detection working

octo-patch:
Fixes #481

## Problem

`llmfit fit --json` always returned `"installed": false` for every catalog model (0 out of 879), even when runtimes reported running models via `/api/v1/installed`.

The root cause: `run_fit()` was building `ModelFit` structs without probing any local runtime provider. `fit.installed` defaults to `false` and was never updated, so the JSON output always showed `installed: false`.

```bash
# Before this fix — always 0:
llmfit fit --json | jq '[.models[] | select(.installed == true)] | length'
```

## Solution

`run_recommend()` already does the right thing — it was fixed in commit 2477d1d to query Ollama, MLX, llama.cpp, Docker Model Runner, and LM Studio before building the fits. `run_fit()` was missed in that change.

This PR mirrors the same provider-probe block into `run_fit()` so that `fit.installed` is set correctly in both the text table (the `✓` prefix) and the `--json` output.

## Testing

- `cargo check` — clean (no new warnings)
- The change pattern is identical to the one already reviewed and merged for `run_recommend` in 2477d1d

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -866,12 +866,25 @@ fn run_fit(
     overrides: &HardwareOverrides,
     context_limit: Option<u32>,
 ) {
+    use llmfit_core::providers::{
+        self as provs, DockerModelRunnerProvider, LlamaCppProvider, LmStudioProvider, MlxProvider,
+        ModelProvider, OllamaProvider,
+    };
+
     let specs = detect_specs(overrides);
     let db = ModelDatabase::new();
 
     if !json && !csv {
         specs.display();
     }
+
+    // Query installed models across local providers so that `fit.installed`
+    // is populated in both text and JSON output — same behaviour as `recommend`.
+    let ollama_installed = OllamaProvider::new().installed_models();
+    let mlx_installed = MlxProvider::new().installed_models();
+    let llamacpp_installed = LlamaCppProvider::new().installed_models();
+    let docker_mr_installed = DockerModelRunnerProvider::new().installed_models();
+    let lmstudio_installed = LmStudioProvider::new().installed_models();
 
     let hidden: usize = db
         .get_all_models()
@@ -883,7 +896,15 @@ fn run_fit(
         .get_all_models()
         .iter()
         .filter(|m| backend_compatible(m, &specs))
-        .map(|m| ModelFit::analyze_with_context_limit(m, &specs, context_limit))
+        .map(|m| {
+            let mut fit = ModelFit::analyze_with_context_limit(m, &specs, context_limit);
+            fit.installed = provs::is_model_installed(&m.name, &ollama_installed)
+                || provs::is_model_installed_mlx(&m.name, &mlx_installed)
+                || provs::is_model_installed_llamacpp(&m.name, &llamacpp_installed)
+                || provs::is_model_installed_docker_mr(&m.name, &docker_mr_installed)
+                || provs::is_model_installed_lmstudio(&m.name, &lmstudio_installed);
+            fit
+        })
         .collect();
 
     if perfect {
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

The logic for the installed flag is spot on and the UI fixes for the URLs make the detail pane look much clener. Good job!

PR #508, opened 2026-04-26
feat(recommend): add --output-llamacpp flag 

panwar-stack:
## Summary

  This PR makes `llmfit recommend` more useful for automation by adding an opt-in
  `--output-llamacpp` flag that includes ready-to-run `llama-cli` commands in the
  JSON response for llama.cpp-compatible recommendations.

  ## What changed

  - Added `--output-llamacpp` to the `recommend` command.
  - Included a `llamacpp_command` field in JSON output for compatible llama.cpp models.
  - Generated commands using the fit's `effective_context_length` instead of the
    model's maximum advertised context.
  - Mapped run modes to more accurate llama.cpp launch flags such as `-ngl all`,
    `-ngl 0`, and `-ngl auto --fit on`.
  - Added conversation mode only for chat/instruct/tool-use models, and omitted it
    for embeddings.
  - Exposed `effective_context_length` in the JSON payload so downstream consumers
    can see the context size used for fit calculations.
  - Added test coverage for command generation and context serialization.

  ## Why

  This gives CLI and agent workflows a direct path from recommendation to execution
  without forcing consumers to reconstruct llama.cpp arguments themselves. It also
  fixes argument selection so the suggested commands better match the actual fit
  calculation and model type.

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -210,12 +210,13 @@ pub struct ModelFit {
     pub moe_offloaded_gb: Option<f64>, // GB of inactive experts offloaded to RAM
     pub score: f64,                    // weighted composite score 0-100
     pub score_components: ScoreComponents,
-    pub estimated_tps: f64,         // baseline estimated tokens per second
-    pub best_quant: String,         // best quantization for this hardware
-    pub use_case: UseCase,          // inferred use case category
-    pub runtime: InferenceRuntime,  // inference runtime (MLX or llama.cpp)
-    pub installed: bool,            // model found in a local runtime provider
-    pub fits_with_turboquant: bool, // TooTight at fp16 KV but fits with TurboQuant KV
+    pub estimated_tps: f64,            // baseline estimated tokens per second
+    pub best_quant: String,            // best quantization for this hardware
+    pub effective_context_length: u32, // context length used for memory estimation
+    pub use_case: UseCase,             // inferred use case category
+    pub runtime: InferenceRuntime,     // inference runtime (MLX or llama.cpp)
+    pub installed: bool,               // model found in a local runtime provider
+    pub fits_with_turboquant: bool,    // TooTight at fp16 KV but fits with TurboQuant KV
 }
 
 impl ModelFit {
@@ -546,6 +547,7 @@ impl ModelFit {
             score_components,
             estimated_tps,
             best_quant: best_quant_str,
+            effective_context_length: estimation_ctx,
             use_case,
             runtime,
             installed: false, // set later by App after provider detection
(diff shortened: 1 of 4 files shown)


three-foxes-in-a-trenchcoat (commented):
Overall: **needs work** — solid idea but has a few issues to address before merge.

**The good:**
- `--output-llamacpp` is a useful addition for automation workflows. Using `-hf` with HuggingFace repo avoids path guessing, which is the right call.
- Using `effective_context_length` instead of max advertised context for the `-c` flag is correct and prevents OOM.
- The `-ngl` mapping to run modes is sensible: `-ngl all` for full GPU, `-ngl 0` for CPU, `-ngl auto --fit on` for partial offload.
- Test coverage is good — covers effective context, ngl variants, conversation mode logic, tensor parallel exclusion, and JSON serialization.
- Adding `effective_context_length` to the `ModelFit` struct and JSON output is a clean non-breaking change.

**Issues:**

1. **package-lock.json noise (blocking).** The PR includes 98 deletions to `llmfit-web/package-lock.json` removing `@testing-library/dom`, `@types/aria-query`, `ansi-regex`, `ansi-styles`, `dom-accessibility-api`, `lz-string`, `pretty-format`, `react-is`, and stripping `"peer": true` from several packages. None of this is mentioned in the PR body. This looks like an accidental `npm install` rebasing the lockfile. Either revert package-lock.json or explain what changed and why. Merging this as-is could break the web app build.

2. **`--fit` flag may not be stable in llama.cpp.** The `-ngl auto --fit on` args for CpuOffload and MoEOffload modes assume `--fit` is available. As of llama.cpp v3.5+, `--fit` was added but earlier versions will error on unknown flag. Consider noting the minimum llama.cpp version required or falling back to a safe default.

3. **`should_use_llamacpp_conversation_mode` is heuristic-heavy.** The function checks `use_case` string contains "chat", "instruct", "instruction", "tool", "function", and name contains "chat", "instruct", "-it". This works for most models but could misfire on edge cases (e.g. a model named "multichat" that is actually an embedding model, or "functionary" that might be general purpose). The logic is reasonable for a first pass but worth a comment that it is best-effort.

4. **UNSTABLE merge state.** The PR shows UNSTABLE on GitHub despite `mergeable: MERGEABLE`. This means CI status checks are failing or inconclusive. Get CI green before merge — at minimum `cargo test` should pass.

5. **Minor: `fit.rs` struct field reordering.** Adding `effective_context_length` between `best_quant` and `use_case` changes struct field order. Not a functional issue in Rust since named fields are used, but if anyone matches on position (unlikely but possible in macros or external bindings), this is a subtle change.

**Verdict: needs work** — fix the package-lock.json noise, check the `--fit` flag compatibility, and get CI green.

panwar-stack:
Thanks for taking the time to review. I have addressed the PR feedback.

three-foxes-in-a-trenchcoat (approved):
LGTM — clean feature. The `--output-llamacpp` flag bridges the gap between recommendations and actual execution. Smart details:

- Using `effective_context_length` instead of max context for the `-c` flag
- `--fit on` auto-detection via `llama-cli --help` 
- Conversation mode only for chat/instruct/tool-use, not embeddings
- `TensorParallel` correctly returns `None` since llama.cpp TP is different

Good test coverage too. CI showing UNSTABLE but the code itself is solid.

AlexsJones:
please format

PR #538, opened 2026-05-07
fix(docker): skip HTTP probe on Linux when Docker Desktop is not running 

navneettoppo:
Avoids ~800ms startup timeout on Linux systems that don't have Docker
Desktop installed. Checks for Docker Desktop socket paths before
attempting the HTTP probe. Falls back to allowing the probe if
DOCKER_MODEL_RUNNER_HOST is explicitly set.

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -12,6 +12,9 @@ pub enum GpuBackend {
     CpuArm,
     CpuX86,
     Ascend,
+    /// AMD XDNA / Ryzen AI NPU (e.g. Ryzen AI 7 350, Ryzen AI 9 HX 370).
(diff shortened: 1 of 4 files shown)


llmfit-core/src/hardware.rs
```diff
@@ -240,6 +249,22 @@ impl SystemSpecs {
             gpus.extend(ascend);
         }
 
+        // AMD XDNA NPU (Ryzen AI series) — detected from CPU name.
```
AlexsJones, line 252: This is completely unrelated to the PR title

AlexsJones (changes requested):
AMD XDNA NPU support is not what this PR claims to be about

three-foxes-in-a-trenchcoat:
Closing in favor of a dedicated PR for just the Docker probe fix. The NPU support changes are unrelated and will be submitted separately.

PR #548, opened 2026-05-12
fix(docker): skip HTTP probe on Linux when Docker Desktop is not running

three-foxes-in-a-trenchcoat:
## Problem

Docker Model Runner is a Docker Desktop feature. On Linux, Docker Desktop is uncommon, so the startup HTTP probe to `localhost:12434` times out after ~800ms on most Linux systems, slowing down every llmfit startup.

## Solution

Add `is_docker_desktop_running()` which checks for Docker Desktop's presence by:
1. Looking for its socket at `/run/docker-desktop/docker.sock` or `~/.docker/desktop/docker.sock`
2. Falling back to checking if `DOCKER_MODEL_RUNNER_HOST` is explicitly set

If Docker Desktop isn't running on Linux, skip the HTTP probe entirely and return `available=false` immediately.

## Changes

- `llmfit-core/src/providers.rs`: Added `is_docker_desktop_running()` helper and early return in `DockerModelRunnerProvider::detect_with_installed()`

## Testing

- On Linux without Docker Desktop: startup is now instant (no 800ms delay)
- On Linux with Docker Desktop: probe still works normally
- On macOS/Windows: no change in behavior (check only applies to Linux via `cfg!(target_os = "linux")`)
- `cargo check` passes cleanly

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -579,7 +579,7 @@ impl SystemSpecs {
         // Filter out integrated GPUs (iGPUs) that have very little VRAM.
         // rocm-smi reports all GPU agents including iGPUs on APUs like
         // Ryzen 9800X3D, which would otherwise inflate the GPU count.
-         // Discrete GPUs have > 2 GB VRAM; iGPUs typically show <= 2 GB.
+        // Discrete GPUs have > 2 GB VRAM; iGPUs typically show <= 2 GB.
         const IGPU_VRAM_THRESHOLD: u64 = 2 * 1024 * 1024 * 1024; // 2 GB
         let discrete_vram: Vec<u64> = per_gpu_vram_bytes
             .iter()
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat:
Addressed your feedback:

1. **Empty string check** ✅ — Now checks `!v.trim().is_empty()` instead of just `.is_ok()`
4. **Test coverage** ✅ — Added `test_docker_desktop_running_via_env_var` covering:
   - Non-empty value → detected
   - Empty string → NOT detected
   - Whitespace-only → NOT detected
   (Combined into a single test to avoid Rust test parallelism races with `set_var`/`remove_var`)

Skipped:
- 2. `cfg!` vs `#[cfg]` — compiler already optimizes dead code, current approach is cleaner
- 3. Home dir detection — `unwrap_or_default()` handles unset `HOME` safely
- 5. Logging — no `tracing`/`log` dependency in this crate, `eprintln!` would be too noisy for a transparent optimization. Users can always set `DOCKER_MODEL_RUNNER_HOST` explicitly for troubleshooting.

PR #561, opened 2026-05-19
chore(devbox): add devbox configuration file and update .gitignore

Romeo-mz:
Added devbox to facilate the maintainer roles to setup local env

Test on wsl and works flawless

To run; use the followings commands : 

- `devbox install`
- `devbox shell`

## Requirements

Need to have devbox installed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ dist/
 .venv/
 .mypy_cache/
 .ruff_cache/
+
+#devbox
+devbox.lock
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,17 @@
+# Code of Conduct
+
+This project adopts the **Contributor Covenant v2.1** as its code of conduct.
+
+The full text is available at: https://www.contributor-covenant.org/version/2/1/code_of_conduct/
+
+## Summary
+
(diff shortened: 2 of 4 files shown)


AlexsJones:
I cannot endorse a paid for product, this puts a bar for entry on contribution. Sorry

Romeo-mz:
I'm guessing you think of another product with same name; 
https://github.com/jetify-com/devbox/blob/main/LICENSE
I think jetify devbox is indeed free ?

AlexsJones:
Looks like they have a paid product but its for cloud usage. https://www.jetify.com/cloud/pricing 

You've been using for free offline?

Romeo-mz:
Yes offline it's free to use
Also I'm thinking about adding a devcontainer to the app ? What do you think about that ?

AlexsJones:
I've never used devbox or devcontainer so need to check them out before I am comfortable saying its the defacto way to develop llmfit.

Romeo-mz:
> I've never used devbox or devcontainer so need to check them out before I am comfortable saying its the defacto way to develop llmfit.

Alright, I'll let you have a look. The idea behind these tools is to enable everyone to begin with an equal setup. In practice, they could make it easier to get started with any codebase.

Even if all that is needed is cargo and Python for llmfit, it could be useful to have these tools available. Even if it could add maintenance to the initial project I guess 😞

AlexsJones:
Thanks for this contribution, but until there is broader adoption - especially within and by the community/maintainers I am going to park this.

PR #565, opened 2026-05-20
feat: add issue templates for bug reports

Romeo-mz:
Add some templates to ease the transmission of informations


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,127 @@
+name: Bug Report
+description: Report a bug in llmfit (incorrect fit analysis, hardware detection issues, TUI rendering problems, etc.)
+title: "[Bug]:(short issue description)"
+labels: ["bug, needs-triage"]
```
AlexsJones: Bug: This creates a **single** label named `"bug, needs-triage"` instead of two separate labels. The comma is inside the quotes, so GitHub treats the whole string as one label.

Should be:
```yaml
labels: ["bug", "needs-triage"]
```

AlexsJones (commented):
Thanks for the contribution! One bug to fix before merging — see inline comment.

.github/ISSUE_TEMPLATE/bug-report.yml
Romeo-mz: Done ! 👍

AlexsJones:
Merged main in, dropped the stray `devbox.lock` (that belongs with #561), and updated the bug template's system-output field to request `llmfit doctor` output — a new subcommand (#675) that captures the raw nvidia-smi/rocm-smi/sysfs output detection relies on, so hardware bug reports arrive as regression-test material. All four templates YAML-validated. Merging — thanks @Romeo-mz!

PR #576, opened 2026-05-24
fix: align SortColumn::next with display order of columns in TUI

oytech:
## Problem

In TUI when you cycle sort column by pressing 's' the params column is out of the display order:

llmfit-sort

llmfit 0.9.28

## Solution

This PR aligns `SortColumn::next` with display order of columns in TUI, particulary moves params column between provider and score.

Looks like `SortColumn::next` is only used in TUI so `llmift-web` and `llmfit-desktop` shouldn't be affected.

I guess it's not a perfect fix as this bug can be reintroduced after changing display order of columns in TUI or adding another sort column, but more foolproof solution probably would require some refactoring.

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -152,14 +152,14 @@ impl SortColumn {
 
     pub fn next(&self) -> Self {
         match self {
+            SortColumn::Params => SortColumn::Score,
             SortColumn::Score => SortColumn::Tps,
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Looks good - the sort cycle now matches the display column order (Provider -> Params -> Score -> TPS -> Mem% -> Ctx -> Date -> UseCase). Clean fix.

PR #583, opened 2026-05-25
feat: add --llama-cpp-path global CLI flag mirroring LLAMA_CPP_PATH env var

### LINKED ISSUE #295: llama.cpp is not supported
I'm trying to use llmfit on a Jetson Orin NX Super device. I have already downloaded and compiled llama.cpp successfully — both llama-cli and llama-server (running on port 8000) are working fine.
However, when I open llmfit, it still shows that llama.cpp is not supported (llama.cpp: ✗), so it only recommends CPU-based models and doesn't take advantage of the Jetson GPU.
How can I make llmfit properly detect my llama.cpp installation? Do I need to set any specific environment variables, add a path, or perform additional configuration?
Thank you!

mvanhorn:
## Summary

Add a global `--llama-cpp-path` CLI flag that mirrors the existing `LLAMA_CPP_PATH` env var. Requested by @NTShop in issue #295.

## Why this matters

`LLAMA_CPP_PATH` already exists for users who keep llama.cpp in a non-standard location. A matching CLI flag is more discoverable than an env var and is the usual ergonomic for one-shot overrides ("just this run, point at my custom build").

## Changes

In `llmfit-tui/src/main.rs`, added `--llama-cpp-path` as a global flag on `Cli`. When present, the value is exported to `LLAMA_CPP_PATH` early in `main()`, before the binary lookup runs, so the existing `find_binary` chain in `llmfit-core/src/providers.rs:1346` picks it up without any further plumbing.

No change to the env-var behavior; the flag stacks on top.

## Testing

`llmfit-tui/tests/cli_smoke.rs` covers the flag-present case (env var set, lookup uses provided path) and the flag-absent case (existing env var behavior unchanged). README updated to mention the new flag alongside the env var.

Closes #295

AI was used for assistance.

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -13,6 +13,7 @@ mod tui_ui;
 
 use clap::{Parser, Subcommand};
 use std::net::{TcpStream, ToSocketAddrs};
+use std::path::Path;
 use std::process::Stdio;
 use std::thread;
 use std::time::Duration;
@@ -105,6 +106,9 @@ GLOBAL FLAGS:
   --memory <SIZE>    Override GPU VRAM (e.g. \"32G\", \"32000M\", \"1.5T\").
   --ram <SIZE>       Override system RAM (e.g. \"64G\", \"128000M\").
   --cpu-cores <N>    Override detected CPU core count.
+  --llama-cpp-path <PATH>
+                     Directory containing llama.cpp binaries. Overrides
+                     LLAMA_CPP_PATH for this invocation when the directory exists.
   --max-context N    Cap context length for memory estimation (tokens).
                      Falls back to OLLAMA_CONTEXT_LENGTH env var if unset.
 
@@ -162,6 +166,11 @@ struct Cli {
     /// Useful for evaluating model fit against target hardware.
     #[arg(long, value_name = "CORES", value_parser = parse_positive_usize)]
     cpu_cores: Option<usize>,
+
+    /// Directory containing llama.cpp binaries (`llama-cli`, `llama-server`).
+    /// Overrides LLAMA_CPP_PATH for this invocation when the directory exists.
+    #[arg(long, value_name = "PATH", global = true)]
+    llama_cpp_path: Option<std::path::PathBuf>,
 
     /// Cap context length used for memory estimation (tokens).
     /// Falls back to OLLAMA_CONTEXT_LENGTH if not set.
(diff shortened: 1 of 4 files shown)


llmfit-tui/src/main.rs
```diff
@@ -2559,7 +2639,11 @@ fn main() {
             Commands::System => {
                 let specs = detect_specs(&overrides);
                 if cli.json {
-                    display::display_json_system(&specs);
+                    if llama_cpp_path_configured {
```
three-foxes-in-a-trenchcoat: This makes the `--json system` output schema depend on whether `--llama-cpp-path` was passed. Two scripts running the same `--json system` will see different shapes based on flag presence, which is a surprising contract. The flag's job is setting an env var. If we want a providers block in the JSON, always include it, or add an explicit `--providers` flag. Right now this looks reverse-engineered from what the test needs to observe.

llmfit-tui/src/main.rs
```diff
@@ -806,6 +815,75 @@ fn resolve_context_limit(max_context: Option<u32>) -> Option<u32> {
     }
 }
 
+fn configure_llama_cpp_path(path: Option<&Path>) -> bool {
+    let Some(path) = path else {
+        return false;
+    };
+
+    if path.is_dir() {
```
three-foxes-in-a-trenchcoat, line 1057: Silently dropping the flag when the directory does not exist is a footgun. Typo the path and you get no signal, just the wrong llama.cpp picked up later. Warn to stderr, or exit non-zero. The test asserting silent ignore is baking in the wrong contract.

three-foxes-in-a-trenchcoat (changes requested):
Flag itself is good. Two things to fix before this lands, inline.

mvanhorn:
Both fixed in 91fad5e:

1. `--json system` now always emits the with-providers shape, so the output schema no longer depends on whether `--llama-cpp-path` was passed.
2. A non-existent `--llama-cpp-path` now prints an error to stderr and exits(1) instead of silently dropping the flag.

Verified it compiles locally with `cargo check`.

mvanhorn:
Fixed in 996b9c7. The Test Suite failure was a stale test, not a bug.

`llama_cpp_path_flag_ignores_missing_directory` expected a nonexistent path to be silently ignored, but `configure_llama_cpp_path` deliberately does the opposite:

```rust
eprintln!(
    "Error: --llama-cpp-path '{}' does not exist or is not a directory.",
    path.display()
);
std::process::exit(1);
```

Failing loudly is the better behaviour for a flag the user passed explicitly, so the test was the wrong side of the mismatch. It now asserts the non-zero exit and the exact stderr, and is renamed to `llama_cpp_path_flag_rejects_missing_directory` to say what it checks.

Verified locally: `cargo test --workspace` is 403 passed, 0 failed, which includes all 11 `cli_smoke` tests.

One note in case it comes up: `cargo clippy --workspace -- -D warnings` reports 29 findings, but every one of them is in `llmfit-core/src/` (`models.rs`, `fit.rs`, `hardware.rs`, `bench.rs`, ...) and this PR does not touch that crate. Clippy on `-p llmfit` alone is clean.

mvanhorn:
Thanks — both points are addressed, and the branch is rebased onto current main (it was conflicted).

**Missing directory now fails loudly.** `--llama-cpp-path` prints the offending path to stderr and exits 1 rather than silently ignoring it. The test that asserted the silent-ignore contract is gone, replaced by `llama_cpp_path_flag_rejects_missing_directory`.

**`--json system` schema no longer depends on the flag.** The providers block is always emitted, so two scripts running the same command see the same shape regardless of whether the flag was passed.

I also took the bot's P2 on the unsafe env mutation, since CONTRIBUTING rules out `unsafe` and the safety argument there rested on call ordering. `llmfit-core` now exposes a set-once path override that `find_binary` consults ahead of `LLAMA_CPP_PATH` — same documented precedence, no env mutation, no unsafe.

Locally: `cargo build` clean, `cargo fmt --check` clean, 13/13 in `cli_smoke`, and no new clippy warnings on the changed lines.

AlexsJones (approved):
Thanks for this. The `OnceLock` override is the right shape to avoid the `unsafe`/racy env mutation, and pushing the name down as a `metadata.name` field selector keeps it API-server-side. The `providers` block under `system --json` is additive, and the flag tests (missing dir, env precedence, help) cover the edge cases. Approving for your final look.

AlexsJones (approved):
Re-review at the rebased head: the `--llama-cpp-path` flag via a `OnceLock` override (takes precedence over `LLAMA_CPP_PATH`, set-once, avoids mutating the environment) is unchanged, just brought up to current main. Still approve-recommend.

Issue #584, opened 2026-05-26
Feature request: Add a weekly ci action to update model database


Romeo-mz:
The goal of this feature will be to always have up to date models database

We could imagine to run the action every monday during the night and then after the update sucess it open a PR on main

I'm linking the PR I currently work on for further inspectation ! 🐫

AlexsJones:
Done via #585 — a weekly scheduled workflow (Mondays 02:00 UTC, also manually triggerable via workflow_dispatch) now refreshes the model database with `scrape_hf_models.py -n 5000`, validates it against a JSON Schema, and opens an automated PR. Thanks @Romeo-mz!

PR #585, opened 2026-05-26
Feature request: Add a weekly ci action to update model database 

Romeo-mz:
Linked issue : #584


.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+'on':
+  schedule:
+    - cron: '0 2 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
```
three-foxes-in-a-trenchcoat, line 16: Please add a `concurrency:` block — a manual `workflow_dispatch` overlapping the cron will race on the same branch.

```yaml
concurrency:
  group: weekly-model-update
  cancel-in-progress: false
```

.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+  update-models:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6.0.2
```
three-foxes-in-a-trenchcoat: `actions/checkout@v6.0.2` — current major is v4; v6 isn't released. Either confirm this tag resolves, or switch to `@v4`.

.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: '3.12'
+
+      - name: Refresh model database
+        run: python3 scripts/scrape_hf_models.py
```
three-foxes-in-a-trenchcoat: No args → script uses its defaults: `--discover-limit 1000`, `--min-downloads 10000`. If the intent is ~5k models, pass `-n 5000` explicitly here. Otherwise this caps the refresh at 1000.

.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          if git diff --quiet; then
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create or update pull request
+        if: steps.changes.outputs.changed == 'true'
+        uses: peter-evans/create-pull-request@v7
```
three-foxes-in-a-trenchcoat: This action has `contents: write` + `pull-requests: write`. Please pin to a full commit SHA (with a `# v7.x.y` trailing comment) rather than a moving tag.

.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+        with:
+          token: ${{ github.token }}
```
three-foxes-in-a-trenchcoat, line 57: Heads-up: PRs opened with the default `GITHUB_TOKEN` **don't trigger downstream workflows** — `ci.yml` won't run on the auto-PR, so a human has to push an empty commit to get checks. Either use a PAT/App token stored as a secret, or document this in the PR body so reviewers know to nudge it.

.github/workflows/weekly-model-update.yml
```diff
@@ -0,0 +1,62 @@
...
+          commit-message: "chore(models): daily model data refresh"
```
three-foxes-in-a-trenchcoat: Cron is weekly but the commit message, title, body, and branch all say "daily". Please rename across lines 54–58 (e.g. `chore(models): weekly model data refresh`, `[Automated] Weekly model data refresh`, `automated/weekly-model-update`).

data/schema.json
```diff
@@ -0,0 +1,299 @@
...
+        "awq",
+        "gptq",
+        "autoround",
+        "mlx",
+        "safetensors"
+      ]
+    },
+    "quantization": {
+      "type": "string",
+      "pattern": "^(F32|F16|BF16|Q[0-9](?:_[A-Z0-9_]+)?|UD-Q[0-9]_[A-Z0-9_]+|AWQ-(?:[0-9]+)bit|GPTQ-Int[0-9]+|AutoRound-(?:[0-9]+)bit|mlx-(?:[0-9]+)bit)$"
```
three-foxes-in-a-trenchcoat: This `quantization` regex is very narrow. Any new quant format the scraper picks up (`Q8_K_XL`, `IQ4_NL`, `MXFP4`, etc.) will fail validation and silently block the refresh PR — which defeats the purpose of a scheduled refresh. Either broaden the pattern, or drop the pattern entirely and rely on `type: string`.

data/schema.json
```diff
@@ -0,0 +1,299 @@
...
+        "name",
+        "provider",
+        "parameter_count",
+        "min_ram_gb",
+        "recommended_ram_gb",
+        "quantization",
+        "context_length",
+        "use_case"
+      ],
+      "additionalProperties": false,
```
three-foxes-in-a-trenchcoat, line 291: `additionalProperties: false` means any new field added by `scrape_hf_models.py` will break this test in CI. For an auto-refreshed catalog that's the wrong default — consider removing it (or at least documenting that schema must change in lockstep with the scraper).

llmfit-core/Cargo.toml
```diff
@@ -21,3 +21,6 @@ serde_yml = "0.0"
 sysinfo = "0.38"
 ureq = { version = "3.2", features = ["json"] }
 which = "8.0.2"
+
+[dev-dependencies]
+jsonschema = "0.46"
```
three-foxes-in-a-trenchcoat: This dev-dep pulls in ~30 transitive crates (aws-lc-rs, cmake, hyper-rustls, fluent-uri, fraction, num-bigint…) — see the +344-line `Cargo.lock` churn — slowing every contributor's `cargo test`. Please use `jsonschema = { version = "0.46", default-features = false }` to drop the HTTP/`$ref` resolution stack you don't need here. Or move validation to a Python step in the workflow only.

llmfit-core/src/fit.rs
```diff
@@ -951,7 +951,7 @@ pub fn rank_models_by_fit_opts_col(
 ///  - ggerganov, llama.cpp Apple Silicon benchmarks (Discussion #4167)
 ///  - Google, "Efficiently Scaling Transformer Inference" (arXiv:2211.05102)
 ///  - ggerganov, llama.cpp NVIDIA T4 benchmarks (Discussion #4225)
-
+///
```
three-foxes-in-a-trenchcoat: Filling this blank line with `///` merges the References block above into the doc comment of the next item (`rank_models_by_fit_opts_col_ddr_bandwidth` / `Read the system DDR bandwidth…`). If those references belong to the previous function, this attaches them to the wrong item. Please confirm with `cargo doc --open`.

scripts/scrape_hf_models.py
```diff
@@ -2672,4 +2672,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
three-foxes-in-a-trenchcoat: Accidental removal of the trailing newline — please restore (POSIX + most linters expect EOF newline).

three-foxes-in-a-trenchcoat (changes requested):
Thanks for this! Nice direction overall — a scheduled refresh + schema test is exactly what this repo needs. A handful of things to rework before this is mergeable; inline comments below.

High-level asks:
- Reconcile **weekly cron vs. "daily" naming** throughout the workflow.
- Decide on the **discovery limit** (default is 1000 models — pass `-n 5000` if you want 5k).
- Make the schema **non-strict** (`additionalProperties: false` + tight `quantization` regex will block legitimate refreshes).
- Swap `github.token` for a PAT (or document the gotcha) so CI runs on the bot PR.
- Add a `concurrency:` guard and pin `peter-evans/create-pull-request` to a SHA.
- Verify `actions/checkout@v6.0.2` exists (current major is v4).
- Run `cargo test -p llmfit-core hf_models_match_schema` locally and confirm the existing `hf_models.json` passes — the strict patterns will likely reject real entries.

data/schema.json
Romeo-mz: Add docs in contibuting guide

.github/workflows/weekly-model-update.yml
Romeo-mz: Should fetch context7 before rewiewing a PR 😸

AlexsJones:
@Romeo-mz sorry this didnt get looked at it - looks like its ready for a re-review?

three-foxes-in-a-trenchcoat (changes requested):
Re-review — thanks for the updates, @Romeo-mz! Good progress, most of the previous round is resolved:

✅ Weekly/"daily" naming reconciled across title, commit message, PR title, and branch
✅ `concurrency:` guard added
✅ `peter-evans/create-pull-request` and the other actions pinned to SHAs
✅ `actions/checkout@v6.0.2` verified — the pinned SHA resolves to the real tag (v6 is out now, latest v6.0.3), so that concern is moot
✅ `quantization` regex dropped (plain `string`)
✅ `jsonschema` dev-dep set to `default-features = false`, which drops the aws-lc-rs/HTTP stack from the lockfile
✅ stray `fit.rs` doc-comment change removed from the diff

One blocker left, plus a few minors:

### 🔴 Blocker: `additionalProperties: false` breaks the test against current data
The schema still has `additionalProperties: false`, and `main`'s data now carries fields the schema doesn't declare — `_audio_backends`, `_audio_rtf_cpu`, `_audio_rtf_gpu`, `_audio_vram_gb` (added by the audio-ASR work). I confirmed these exist in `data/hf_models.json` and aren't in the schema, so `cargo test -p llmfit-core hf_models_match_schema` will fail once this is on current `main`. This is the exact failure mode flagged last round: a strict schema silently blocks the very refresh PRs this workflow exists to create. Please drop `additionalProperties: false` (recommended) or add these fields and commit to keeping the schema in lockstep with the scraper.

### 🟡 Minors
- **Discovery limit:** the workflow still runs the scraper with no args, so it caps at the default `--discover-limit 1000`. If ~5k is intended, pass `-n 5000` explicitly. (Decision still open from last round.)
- **`github.token`:** PRs opened with the default token don't trigger `ci.yml`. Either use a PAT/App token or add a one-line note in the auto-PR body so a human knows to nudge checks.
- **`scripts/scrape_hf_models.py`:** the EOF newline is still missing and there's now a trailing-whitespace line at the end — please restore a clean trailing newline.
- **`CONTRIBUTING.md`:** two list items are numbered `6.` now (schema step + "Open a PR") — renumber.
- Nits: `data/schema.json` has a leading blank line and no trailing newline; `llmfit-core/Cargo.toml` lost its trailing newline.

Sort the schema blocker and this is basically there. Nice work on the workflow hardening.

Romeo-mz:
Done @AlexsJones @three-foxes-in-a-trenchcoat  🐫 !

AlexsJones:
Re-reviewed against current `main` — the previous round's asks are all resolved (`-n 5000`, token caveat documented, audio fields in the schema, EOF newlines, numbering). Running the schema test against a local merge with main surfaced one last drift: four catalog models now carry the `audio` capability, which the `capability` enum rejected. Rather than bounce this back for another round, I pushed a one-line maintainer edit adding `audio` to the enum (a812232). Full `llmfit-core` suite (391 tests) + `hf_models_match_schema` pass on the merged state locally. Merging — thanks for sticking with this through the review rounds, @Romeo-mz! 🎉

One heads-up: #672 removes the duplicated root `data/` copies of the catalog; after it lands I'll follow up to point the workflow's validate step and the schema test at `llmfit-core/data/` only.

PR #588, opened 2026-05-30
Fix stale installed model detection

jimmckeeth:
## Issue

LlmFit was reporting models as downloaded when the download failed (partially downloaded), or deleted. This prevented redownloading the model.

## Summary

- validate llama.cpp download records against files on disk, including expected byte sizes when known
- stop stale or partial GGUF downloads from blocking a fresh pull
- fix installed-state false positives from Ollama family tags such as `qwen2.5-coder`
- add a hidden `debug-installed` diagnostic command that prints provider counts and matched keys
- add `Ctrl+F` reset-all-filters handling and document it in the TUI help/filter popup
- restore HF cache scan for GGUF models downloaded via `hf download` (not tracked in llmfit history), with a tighter file-existence check to avoid the original false positives

## Root Cause

Installed-state matching had two separate false-positive paths. llama.cpp state could be derived from broad cache scans or stale records, and Ollama installed detection inserted synthetic family keys like `qwen2.5-coder`. That family key matched every Qwen2.5-Coder catalog entry, so one installed Ollama model caused dozens of rows to be marked installed and blocked pulls.

## Validation

- `cargo test -p llmfit-core`
- `cargo check -p llmfit`
- `llmfit --json debug-installed` against the reported local state now returns exactly the two expected installed rows: one Ollama model and one llama.cpp GGUF file

---

Used Codex and validated with Claude Code.

This PR was reviewed by Claude Code, which identified one issue: the initial commit removed `scan_hf_cache_for_gguf()`, breaking detection of models downloaded via external tools like `hf download`. That issue has been fixed in the follow-up commit with a tighter implementation that verifies actual `.gguf` file presence before reporting a model as installed, avoiding the false positives that motivated the original removal.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -532,6 +532,59 @@ fn scan_hf_cache_for_mlx() -> HashSet<String> {
     set
 }
 
+/// Scan HuggingFace cache directories for GGUF models downloaded via `hf download`.
+///
+/// Unlike the directory-name heuristic removed in this PR, this verifies that actual
+/// `.gguf` files exist inside the repo's snapshots before reporting it as installed.
+fn scan_hf_cache_for_gguf() -> (HashSet<String>, usize) {
+    let mut set = HashSet::new();
+    let mut count = 0usize;
+    for cache_dir in dirs_hf_cache_all() {
+        let Ok(entries) = std::fs::read_dir(&cache_dir) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            let name = entry.file_name();
+            let name_str = name.to_string_lossy();
+            let Some(rest) = name_str.strip_prefix("models--") else {
+                continue;
+            };
+            let mut parts = rest.splitn(2, "--");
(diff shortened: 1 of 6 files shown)


jimmckeeth:
### Code review

Found 1 issue:

1. Removing `scan_hf_cache_for_gguf()` from `installed_models_counted()` breaks detection of GGUF models that were downloaded via external tools (e.g. `hf download`). Those models are not recorded in llmfit's download history, so the new history-based detection path misses them entirely, and there is no longer any HF cache scan as a fallback. Users with pre-existing models in `~/.cache/huggingface/hub/` will no longer see those models as installed. The removed code was introduced in commit `6093bcc` specifically to cover this case.

https://github.com/AlexsJones/llmfit/blob/7142ba58756dae25114d45bdbc53683019d84ade/llmfit-core/src/providers.rs#L699-L706

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #590, opened 2026-05-31
Web dashboard: stretch layout to full screen width

RyuShai:
## Summary

The `llmfit serve` web dashboard `.page-shell` container was hard-capped at `max-width: 1640px` and centered with `margin: 0 auto`. On wide / ultrawide displays (e.g. **21:9**) this left large empty margins on both sides and the model table did not use the available horizontal space.

- Removed the fixed cap → layout now fills 100% of the viewport width.
- Added responsive horizontal padding `clamp(1.1rem, 2.5vw, 3rem)` so content keeps comfortable gutters on very wide screens.

Single-line CSS change in `llmfit-web/src/styles.css` (`.page-shell`).

## Testing
- `npm test` → all passing.
- `npm run build` succeeds; verified the rebuilt binary serves the updated bundle via `llmfit serve`.

### DIFF
--- a/llmfit-web/src/styles.css
+++ b/llmfit-web/src/styles.css
@@ -68,9 +68,10 @@ body {
 
 .page-shell {
   position: relative;
-  max-width: 1640px;
+  width: 100%;
+  max-width: 100%;
   margin: 0 auto;
-  padding: 1.1rem;
+  padding: 1.1rem clamp(1.1rem, 2.5vw, 3rem);
   display: grid;
   gap: 0.95rem;
 }


AlexsJones:
Thanks @RyuShai, and apologies for the slow first look. The `clamp()` gutter padding is a nice touch. One thing worth deciding before this lands: dropping the cap to `max-width: 100%` makes the model table span the full width of a 21:9/32:9 display, and very long line-lengths in a data table can actually hurt scan-ability — the original 1640px cap was likely deliberate for that reason.

A middle ground that fixes the "big empty margins" complaint without going fully edge-to-edge would be to raise the cap rather than remove it, e.g. `max-width: min(2200px, 100%)`, keeping your `clamp()` padding. That reclaims most of the wasted space on ultrawides while still bounding line length on the extreme end.

Either way it's a visual-direction call, so I've left the final say to Alex. Happy to re-review quickly if you adjust.

🤝 _Handled by Alex's [repo steward](https://github.com/AlexsJones/repo-steward) — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)_

AlexsJones:
@RyuShai — following up on my note from a couple of weeks back, since I don't want this sitting unreviewed any longer than it already has.

I've since checked the mechanics properly and the change is sound. `box-sizing: border-box` is set globally, so the `clamp()` gutters can't push the shell into overflow, and the `@media (max-width: 780px)` rule re-declares `padding` with the shorthand — which means it fully overrides your clamp and mobile is completely unaffected. Between 780px and the old cap you get slightly roomier gutters than before, which reads as an improvement rather than a regression. Two harmless leftovers if you touch this again: `width: 100%` is a no-op on a block-level grid container, and `margin: 0 auto` no longer does anything once the cap is `100%` — both can go.

So the only thing still open is the design call, which is the same one I raised before: whether the model table should run edge-to-edge on a 32:9 panel. My hesitation is that the 1640px cap looks deliberate — long line lengths in a dense data table hurt scan-ability, and "fills the screen" and "is readable" diverge past a certain width. `max-width: min(2200px, 100%)` with your `clamp()` padding kept would reclaim most of the wasted margin on ultrawides while still bounding the extreme end, and I think that's the version I'd merge today.

If you'd rather make the case for going fully unbounded, I'm genuinely open to it — a screenshot of the model table at 21:9 would probably settle it faster than either of us arguing from principle. Otherwise, push that one-line adjustment and I'll get this in. Sorry again for the wait on the first look.

AlexsJones (changes requested):
Sorry this sat so long, and thanks for the `clamp()` on the padding, that part is the right instinct.

The complaint is real: 1640px on a 21:9 panel leaves most of the screen empty and the model table is the one thing that genuinely wants the width. I would rather raise the cap than remove it, though. With `max-width: 100%` a 5120px ultrawide gets a table whose rows are over a metre of tracking distance, and the header, notes and estimate text in that same grid become hard to read at full bleed.

Concretely, something like this keeps your fix for the common 3440px case without unbounding it:

```suggestion
.page-shell {
  position: relative;
  width: 100%;
  max-width: 2400px;
  margin: 0 auto;
  padding: 1.1rem clamp(1.1rem, 2.5vw, 3rem);
  display: grid;
  gap: 0.95rem;
}
```

2400px fills a 3440x1440 panel edge to edge and still centres on the very wide ones. If you have measured that a hard 100% reads better on your display, say so and make the case, I am not attached to the number.

No CI has ever run on this branch and it is a few months behind main now, so it will need a rebase either way. If you have moved on, no problem at all, just say and we will carry it forward from here.

PR #591, opened 2026-05-31
fix: resolve #185 — Windows security block

### LINKED ISSUE #185: Windows security block
When i cargo build i get this

error: failed to run custom build command for `quote v1.0.45`

Caused by:
  could not execute process `C:\Users\joeco\AppData\Local\Temp\cargo-installQ7KYa2\release\build\quote-d029a7712f238a86\build-script-build` (never executed)

Caused by:
  Eine Anwendungssteuerungsrichtlinie hat diese Datei blockiert. (os error 4551)
warning: build failed, waiting for other jobs to finish...
error: failed to compile `llmfit v0.6.4`, intermediate artifacts can be found at `C:\Users\joeco\AppData\Local\Temp\cargo-installQ7KYa2`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

chinhkrb113:
## Summary

Add a workspace-level Cargo config that pins the build output to a `target/` directory inside the repo. This ensures that contributors who clone and build locally (Option 2 in the README) never have build scripts executed from `%LOCALAPPDATA%\Temp`, sidestepping the Windows application-control block for the from-source path. Note this only affects local builds of the repo; it does not change behavior of `cargo install` from crates.io (which always uses a temp dir), so it complements rather than replaces the documentation fix

Fixes #185

## Changes

- `.cargo/config.toml` *(new)*
- `llmfit-core/tests/cargo_config.rs` *(new)*

## Why

`.cargo/config.toml`: Add a workspace-level Cargo config that pins the build output to a `target/` directory inside the repo. This ensures that contributors who clone and build locally (Option 2 in the README) never have build scripts executed from `%LOCALAPPDATA%\Temp`, sidestepping the Windows application-control block for the from-source path. Note this only affects local builds of the repo; it does not change behavior of `cargo install` from crates.io (which always uses a temp dir), so it complements rather than replaces the documentation fix.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+target-dir = "target"
--- a/llmfit-core/tests/cargo_config.rs
+++ b/llmfit-core/tests/cargo_config.rs
@@ -1,0 +1,34 @@
+use std::fs;
+use std::path::PathBuf;
+
+fn workspace_root() -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .expect("workspace root")
+        .to_path_buf()
+}
+
+#[test]
+fn cargo_config_exists() {
+    let config_path = workspace_root().join(".cargo/config.toml");
+    assert!(
+        config_path.exists(),
+        ".cargo/config.toml should exist at the workspace root"
+    );
+}
+
+#[test]
+fn cargo_config_pins_target_dir_inside_repo() {
+    let config_path = workspace_root().join(".cargo/config.toml");
+    let contents =
+        fs::read_to_string(&config_path).expect("failed to read .cargo/config.toml");
+
+    assert!(
+        contents.contains("[build]"),
+        ".cargo/config.toml should contain a [build] section"
+    );
+    assert!(
+        contents.contains("target-dir = \"target\""),
+        "build.target-dir should be pinned to \"target\""
+    );
+}


AlexsJones:
Thanks @chinhkrb113, and sorry for the slow first look here. I dug into #185 and I don't think this change resolves the reported failure — wanted to lay out the reasoning:

- The error in #185 is `os error 4551` ("Eine Anwendungssteuerungsrichtlinie hat diese Datei blockiert" / an application-control policy blocked this file) raised while running the `quote` crate's **build script**, during a `cargo install` — note the failing path is under `C:\Users\...\AppData\Local\Temp\cargo-installQ7KYa2\...`. That's a Windows WDAC/AppLocker policy blocking executables that run from `%TEMP%`, not a target-directory problem.
- `.cargo/config.toml` with `target-dir = "target"` is already the Cargo default for an in-repo `cargo build`, so it's a no-op there. And `cargo install` builds in its own temporary directory and doesn't pick up the cloned repo's `.cargo/config.toml` for that temp build, so it wouldn't move the blocked build-script out of `%TEMP%` either.
- The two added tests only assert that `.cargo/config.toml` exists and contains `target-dir = "target"` — they verify the file was added, not that #185's build failure is fixed.

The actual fix Alex is already pursuing is Windows code signing via SignPath (see discussion [#461](https://github.com/AlexsJones/llmfit/discussions/461)), which addresses the application-control block at its root. So I'd hold off on this one — but I've flagged it for Alex to make the final call, since closing is his decision, not mine.

If you're hitting this locally as a workaround, setting `CARGO_TARGET_DIR` to a non-`%TEMP%` path you control (or building from a cloned repo rather than `cargo install`) usually gets past the WDAC block.

🤝 _Handled by Alex's [repo steward](https://github.com/AlexsJones/repo-steward) — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)_

PR #593, opened 2026-05-31
fix: resolve #171 — Include ONNX models.

### LINKED ISSUE #171: Include ONNX models.
Include ONNX models.

This would be very useful for on-device/edge inference like webGPU.

chinhkrb113:
## Summary

Combined multi-file contribution:

## Changes

- `data/onnx_models.json` *(new)*
- `llmfit-core/data/onnx_models.json` *(new)*
- `llmfit-core/tests/onnx_models.rs` *(new)*

## Why

`data/onnx_models.json`: Mirrors the existing `data/hf_models.json` / `data/docker_models.json` pattern. This file holds the generated ONNX model catalog so the data is available without re-running the scraper.


AlexsJones:
Thanks for tackling #171 @chinhkrb113 — seed ONNX catalog data is a genuinely useful starting point, and the validation test is nicely written. Before this can land, though, there's a gap I want to flag honestly: as it stands the data isn't wired into the app, so no ONNX model would actually appear anywhere yet.

**Main thing: nothing loads `onnx_models.json`.** The catalogs are embedded into the binary via `include_str!` — `hf_models.json` (`llmfit-core/src/models.rs:1022`) and `docker_models.json` (`llmfit-core/src/providers.rs:2409`). There's no equivalent for ONNX and no `onnx` reference anywhere in `src/`, so `llmfit fit` / the TUI never see these models. To actually resolve #171, the data needs to flow into the model database.

A few specific things to address:

1. **Wire it in.** Add an `include_str!("../data/onnx_models.json")` load path (mirroring `HF_MODELS_JSON`) and parse it into the model DB so the entries surface in fit results.
2. **Schema + format.** The current shape (`id`, `parameters: \"3.8B\"`, `onnx_files: {quant: bytes}`) doesn't match `LlmModel`/`HfModelEntry`, which need fields like `provider`, `min_ram_gb`, `recommended_ram_gb`, `context_length`, `use_case`, `capabilities`. There's also no `ModelFormat::Onnx` variant yet (enum is Gguf/Awq/Gptq/Autoround/Mlx/Safetensors). So this needs a `ModelFormat::Onnx` plus either the catalog in the existing schema or a converter — and ideally memory/fit estimation that's sane for ONNX quantizations.
3. **The two data files diverge.** `data/onnx_models.json` (87 lines, full schema) and `llmfit-core/data/onnx_models.json` (37 lines) aren't mirrors. The embedded `llmfit-core/` copy is the one that matters and it's the incomplete one — they should be kept in sync (the hf_models pair is the pattern).
4. **Test path.** `tests/onnx_models.rs` reads `../data/onnx_models.json` at runtime via `fs::read_to_string`, which is CWD-dependent and validates the repo-root copy rather than the embedded one. Once it's loaded via `include_str!`, prefer asserting against the parsed embedded catalog (like the `hf_models` tests) so it can't silently test the wrong file.
5. **CI:** Rustfmt is currently failing — please run `cargo fmt`.

Happy to help review once the loader path is in place. If you'd rather land the seed data and integration separately, that's fine too — just let's not merge the inert files on their own, since they'd imply ONNX support that isn't wired up yet. Appreciate the contribution!

chinhkrb113:
Thanks for the detailed review and guidance @AlexsJones 

I’ve pushed an update that addresses the ONNX catalog integration and the merge conflicts with the latest `main`.

Changes made:
- Added `ModelFormat::Onnx`.
- Embedded `llmfit-core/data/onnx_models.json` via `include_str!`.
- Wired ONNX entries into `ModelDatabase::embedded()` / `load_embedded()` so they surface in fit results.
- Added conversion from the ONNX catalog schema into `LlmModel`, including RAM/VRAM estimation and quantization normalization.
- Synced the repo-root and `llmfit-core` ONNX catalog files.
- Updated ONNX tests to validate the embedded catalog and ensure it is actually loaded into the model database.
- Resolved the upstream merge conflicts around the new `languages` field and `entry_to_model()` mapper.

Validation run locally:
- `cargo fmt -- --check`
- `cargo test -p llmfit-core`
- `cargo check -p llmfit`

The PR is now mergeable again. Please take another look when you have time.

AlexsJones (approved):
Reviewed the full diff. This mirrors the established `hf_models.json` / `docker_models.json` seed-catalog pattern cleanly: additive `ModelFormat::Onnx`, an `OnnxModelEntry` → `LlmModel` conversion, and a 109-line test that pins both catalog copies in sync, non-empty, unique IDs, positive quant sizes, and round-trips through the embedded DB. No changes to existing formats; `MERGEABLE`.

One design question for the maintainer, not a blocker:

- `select_onnx_quantization` prefers the *most* quantized variant first (`q4` → `q4f16` → … → `fp32`), so `min_ram_gb`/`min_vram_gb` are derived from the smallest weights. That yields the most optimistic "fits" verdict — if a model ships both q4 and fp16, the user is shown the q4 footprint. Is preferring the smallest quant the intended default, or should ONNX mirror the `QUANT_HIERARCHY` ordering the other formats use?

The two `onnx_files` `.expect(...)` panics only fire on the bundled seed data, which the tests already guard, so they're consistent with the existing embedded-parse `.expect`s. Recommending merge — ready for your final look.

🤝 _Handled by Alex's repo steward — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)_

PR #603, opened 2026-06-02
feat(data): add Whisper/ASR model entries + audio support roadmap

melnikaite:
## What

Adds 4 `automatic-speech-recognition` model entries to `data/hf_models.json`:

| Model | Params | RTF (GPU) | RTF (CPU) | VRAM |
|---|---|---|---|---|
| openai/whisper-large-v3 | 1.5 B | 0.04 | 0.6 | 2.5 GB |
| openai/whisper-large-v3-turbo | 809 M | 0.007 | 0.12 | 1.5 GB |
| distil-whisper/distil-large-v3 | 756 M | 0.004 | 0.08 | 1.2 GB |
| openai/whisper-medium | 307 M | 0.003 | 0.05 | 0.6 GB |

New per-model fields use `_audio_` prefix — `#[serde(default)]` compatible,
won't break existing deserialization.

## Why

Choosing the right Whisper model for your hardware is just as confusing as
choosing an LLM. RTF on a GTX 1660 Ti vs Apple M3 Pro differs by 15×.
llmfit already has all the hardware detection needed — this brings the same
reasoning to the audio domain.

## Roadmap (see `AUDIO_SUPPORT.md`)

The data is the first step. The Rust integration (`UseCase::Audio`,
`AudioFit` RTF-based scoring, `MlxWhisperProvider` + `FasterWhisperProvider`
detection in `providers.rs`, `--kind audio` CLI filter) is documented in
`AUDIO_SUPPORT.md` and ready to discuss/implement based on your feedback.

Happy to split into separate PRs if you prefer data-first then Rust.

### DIFF
--- a/AUDIO_SUPPORT.md
+++ b/AUDIO_SUPPORT.md
@@ -1,0 +1,129 @@
+# Audio Model Support — Implementation Plan
+
+This document describes the Rust changes needed to fully support
+`pipeline_tag: automatic-speech-recognition` models (Whisper variants).
+
+The JSON data additions (`data/hf_models.json`) in this branch are ready.
+The Rust integration changes below are the next step — open for discussion.
+
+## Data changes (this branch)
+
+`data/hf_models.json` — 4 new entries with:
+- `pipeline_tag: "automatic-speech-recognition"`
+- `capabilities: ["audio"]`
+- New fields (custom, don't break existing Rust deserialization via `#[serde(default)]`):
+  - `_audio_rtf_gpu: f64` — Real-Time Factor on GPU (0.007 = 7x realtime)
+  - `_audio_rtf_cpu: f64` — RTF on CPU
+  - `_audio_vram_gb: f64` — VRAM needed at F16
+  - `_audio_backends: [str]` — supported servers
+
+## Rust changes needed
+
+### 1. `llmfit-core/src/models.rs`
+
+Add `Capability::Audio` to the `Capability` enum:
+
+```rust
+pub enum Capability {
+    Vision,
+    ToolUse,
+    Reasoning,
+    Embedding,
+    Audio,   // ← new
+}
+```
+
+Extend `LlmModel` deserialization to accept the new `_audio_*` fields:
+
+```rust
+// Inside LlmModel or a companion AudioMeta struct
+#[serde(default)]
+pub audio_rtf_gpu: Option<f64>,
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat:
Thanks for this — the audio direction is compelling and the roadmap doc is genuinely helpful. Two blockers though, both verified by building the branch:

**1. Wrong data file — currently a no-op.** The binary embeds `llmfit-core/data/hf_models.json` (via `include_str!("../data/hf_models.json")` in `models.rs:995`), not the root `data/hf_models.json` this PR edits. I rebuilt the branch and queried the DB — the Whisper entries don't load at all (`whisper count: 0`).

**2. `"capabilities": ["audio"]` would panic if moved to the right file.** `Capability` only has `Vision`/`ToolUse` with no `#[serde(other)]` fallback, and `load_embedded()` uses `.expect()`. Deserializing `"audio"` errors with `unknown variant 'audio'`, which would crash `ModelDatabase::new()` on every startup.

So the data needs the Rust enum change (`Capability::Audio`) to land **first/together**, and the entries belong in `llmfit-core/data/`. A test that loads the DB and asserts a Whisper model is present would catch both. Happy to help with the enum + test if useful.

melnikaite:
Thanks for building the branch and pinpointing both — fixed in 4815bf8:

1. **Embedded file.** Synced the 4 ASR entries into `llmfit-core/data/hf_models.json`
   too, keeping both copies identical the way `scrape_hf_models.py` writes them
   (`output_paths`). The embedded DB now loads them.
2. **`Capability::Audio`.** Added the variant (+ `label()`/`all()`), so `"audio"`
   deserializes instead of panicking `load_embedded()`. Also taught
   `Capability::infer()` to derive Audio from the whisper architecture/name/use_case,
   so future ASR models the scraper adds get tagged without an explicit `capabilities`.
3. **Test.** Added `test_embedded_database_includes_whisper_audio_models` — loads the
   embedded DB and asserts Whisper models are present with `Capability::Audio`. It fails
   with exactly the `whisper count: 0` symptom if the entries ever drift out of the
   embedded copy.

`cargo test -p llmfit-core` → 372 passed. The `_audio_*` fields stay as forward-looking
metadata for the AudioFit scoring in AUDIO_SUPPORT.md; `HfModelEntry` ignores unknown
fields so they don't affect deserialization.

PR #607, opened 2026-06-03
feat(lmstudio): Add LMSTUDIO_API_KEY support for authenticated instances

kiwivda:
- Read API key from LMSTUDIO_API_KEY environment variable
- Attach Authorization: Bearer header to all HTTP requests to LM Studio
  * detect_with_installed() - model listing probe
  * is_available() - availability check
  * start_pull() POST - model download initiation
  * start_pull() polling GET - download progress tracking

This allows llmfit to work with LM Studio instances that have API authentication enabled (required for MCP server access).

Fixes: users with Require API Key enabled in LM Studio get 401 errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -860,6 +860,15 @@ To connect to LM Studio on a different host or port, set the `LMSTUDIO_HOST` env
 LMSTUDIO_HOST="http://192.168.1.100:1234" llmfit
 ```
 
+### API authentication
+
+If your LM Studio instance has **Require API Key** enabled (required for MCP server access), set the `LMSTUDIO_API_KEY` environment variable to provide a Bearer token with all requests:
+
+```sh
+export LMSTUDIO_API_KEY="your-api-key-here"
+llmfit
+```
+
 ### Model name mapping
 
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):

AlexsJones:
Thanks @kiwivda please cargo fmt

AlexsJones (changes requested):
Thanks for the contribution! The core logic is correct — all 4 HTTP call sites are covered. A few things to address before merging:

1. **Empty string guard**: `std::env::var("LMSTUDIO_API_KEY").ok()` returns `Some("")` if the var is set but empty, which sends a broken `Authorization: Bearer ` header. Please add `.filter(|k| !k.is_empty())` when reading the env var.

2. **README documentation**: Please add `LMSTUDIO_API_KEY` to the "Remote LM Studio instances" section of the README alongside the existing `LMSTUDIO_HOST` docs so users can discover this feature.

3. **Tests**: Would be good to have at least one unit test verifying the key is read from the env var in `Default::default()`.

kiwivda:
I fixed the 3 things you asked for:

Empty string guard — added .filter(|k| !k.is_empty()) in Default::default() to avoid Some("")
README docs — added "API authentication" section with documentation of LMSTUDIO_API_KEY
Unit test — added test_lmstudio_provider_default_reads_api_key() that checks all three cases (unset, set, empty)
Can you take another look? 🙏

PR #618, opened 2026-06-08
Detect Intel Mac Metal GPUs

souravsachin:
## Summary

- detect Intel macOS Metal GPUs from `system_profiler SPDisplaysDataType -json`
- keep macOS integrated GPUs visible while still sorting the largest VRAM GPU first
- add a parser-level test fixture for Intel HD Graphics plus Radeon Pro Metal output

## Root cause

Intel Macs with discrete AMD GPUs expose those devices through macOS Metal/system_profiler, but they are not discoverable through the existing Apple Silicon, NVIDIA, ROCm, sysfs, or Vulkan paths. As a result, older Intel MacBook Pros could incorrectly report CPU-only hardware even when Activity Monitor and Metal-capable runtimes show GPU devices.

## Validation

- `cargo fmt`
- `cargo test -p llmfit-core hardware -- --nocapture`
- `cargo check -p llmfit-core`

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -221,6 +221,18 @@ impl SystemSpecs {
             }
         }
 
+        // Intel macOS machines expose Intel and AMD GPUs through Metal, but
+        // not through Linux ROCm/sysfs or NVIDIA-specific tools. Read
+        // system_profiler so older MacBook Pros report their discrete Radeon.
+        for mac_gpu in Self::detect_macos_metal_gpus() {
+            let dominated = gpus
+                .iter()
+                .any(|existing| Self::is_same_gpu_name(&existing.name, &mac_gpu.name));
+            if !dominated {
+                gpus.push(mac_gpu);
+            }
+        }
+
         // Apple Silicon (unified memory)
         if let Some(vram) = Self::detect_apple_gpu(total_ram_gb) {
             let name = if cpu_name.to_lowercase().contains("apple") {
@@ -270,7 +282,11 @@ impl SystemSpecs {
         // integrated GPUs so the discrete GPU becomes primary. This applies
         // globally, not just to the Windows WMI path, to handle cases where
         // an iGPU is detected via Vulkan or APU detection alongside a dGPU.
-        gpus = Self::prefer_discrete_gpus(gpus);
+        // Keep macOS Metal iGPUs visible because Activity Monitor and
+        // llama.cpp's Metal device list can expose both built-in GPUs.
+        if !cfg!(target_os = "macos") {
+            gpus = Self::prefer_discrete_gpus(gpus);
+        }
 
         // Sort by VRAM descending so the best GPU is primary
         gpus.sort_by(|a, b| {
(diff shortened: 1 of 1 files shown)


AlexsJones:
## Review

The core feature — parsing `system_profiler SPDisplaysDataType -json` for Metal GPUs on Intel Macs — looks solid, and the test coverage is good. A few things to address before merge:

### 1. `cfg!()` vs `#[cfg()]`

`detect_macos_metal_gpus()` uses runtime `if !cfg!(target_os = "macos")` which still compiles the `system_profiler` code on Linux/Windows (just never executes it). Should use `#[cfg(target_os = "macos")]` with a `#[cfg(not(target_os = "macos"))]` stub returning `Vec::new()` to avoid compiling platform-specific code on other targets.

### 2. Skipping `prefer_discrete_gpus` on all macOS is too broad

```rust
if !cfg!(target_os = "macos") {
    gpus = Self::prefer_discrete_gpus(gpus);
}
```

This disables iGPU filtering for **all** macOS, including Apple Silicon. The intent is to keep both Intel HD + Radeon visible on Intel Macs specifically — but this blanket skip affects every Mac. A narrower approach would be to skip the filter only when multiple Metal GPUs were detected from `system_profiler`, or check that we're on an Intel Mac.

### 3. Unnecessary `system_profiler` call on Apple Silicon

The new detection block runs before the Apple Silicon check. On every Apple Silicon Mac, this will spawn `system_profiler`, find the Apple GPU, then filter it out. Consider gating on `!cfg!(target_arch = "aarch64")` or moving it after the Apple Silicon detection and skipping if an Apple GPU was already found.

AlexsJones (approved):
Reviewed the full diff — this looks good, and it fills a real gap: Intel Macs with discrete Radeons had no working detection path (not Apple-Silicon, NVIDIA, ROCm, sysfs, or Vulkan).

What I checked:
- `parse_macos_metal_gpus_from_system_profiler_json` is a pure parser with a fixture test covering an Intel iGPU (shared VRAM), a discrete Radeon, a non-Metal `Display` entry, and an Apple Silicon entry — all filtered/parsed correctly.
- Apple Silicon is correctly excluded here (still handled by `detect_apple_gpu` as unified memory), so no double-counting.
- The dedup guard via `is_same_gpu_name` protects against the Vulkan/MoltenVK path reporting the same device twice.
- Runtime `cfg!(target_os = "macos")` keeps the code compiled and testable on every platform, and `system_profiler` failures degrade to an empty list.

Two deliberate behavior changes worth the maintainer's eye (both look reasonable):
- `prefer_discrete_gpus` is now skipped on macOS, so Intel Macs list the iGPU alongside the dGPU (the largest-VRAM sort still keeps the Radeon primary).
- Intel iGPU shared memory is reported as `vram_gb` with `unified_memory: false`, consistent with how shared-memory iGPUs are treated elsewhere.

Note: CI hasn't run on this PR yet — worth approving the workflow run before merge so the full suite confirms.

Recommending this for @AlexsJones's final look. Thanks @souravsachin — a focused change with tests.

PR #630, opened 2026-06-14
feat: add C-ABI (FFI) shared library for native integrations

jimmckeeth:
This PR introduces a new crate `llmfit-ffi` that exposes a C-compatible ABI (shared library) for `llmfit-core`. 

### Why?
This allows non-Rust applications (C, C++, Python, Pascal, Go, etc.) to leverage `llmfit` hardware detection and model fit analysis directly in-process with zero overhead.

### Changes:
- **New Crate `llmfit-ffi`**: A `cdylib` target that exports standard C symbols for lifecycle, hardware specs, and model recommendations.
- **Stateful Context**: Uses an opaque context handle to manage internal state.
- **JSON Bridge**: Returns results as UTF-8 JSON strings to ensure forward compatibility and simplify data marshalling across the FFI boundary.
- **Documentation**: 
  - New `llmfit-ffi/README.md` with technical details and a C usage example.
  - Updated `API.md` and main `README.md` to reflect native integration capabilities.

### Testing:
The FFI has been verified with a Pascal-based wrapper in a separate project, successfully running hardware detection and recommendation queries.

### DIFF
--- a/API.md
+++ b/API.md
@@ -224,6 +224,24 @@ curl "http://127.0.0.1:8787/api/v1/models/top?limit=5&min_fit=good&use_case=codi
 curl "http://127.0.0.1:8787/api/v1/models/Mistral?runtime=any"
 ```
 
+## Native Integration (C-ABI / FFI)
+
+For native applications (C, C++, Python, Pascal, Go, etc.) that require zero-overhead integration without running a separate server or process, `llmfit` provides a C-ABI compatible shared library.
+
+### C-ABI Features
+- **Zero Overhead:** Direct in-process calls to `llmfit-core`.
+- **Stateful:** Manage hardware state via a persistent context handle.
+- **Language Agnostic:** Standard C calling convention (`cdecl`) supported by almost all languages.
+- **JSON Bridge:** Uses JSON strings for data exchange to maintain forward compatibility and simplify marshalling.
+
+### Getting Started
+Build the shared library:
+```sh
+cargo build -p llmfit-ffi --release
+```
+
+See [llmfit-ffi/README.md](./llmfit-ffi/README.md) for detailed API documentation and C usage examples.
+
 ---
 
 ## MCP Server Mode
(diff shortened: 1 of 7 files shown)


jimmckeeth:
If you want wrapper examples or tests I can provide those eventually. Still working on that part.

three-foxes-in-a-trenchcoat (changes requested):
Really nice addition — the FFI design is solid: opaque context handle, JSON bridge for forward-compat, thread-local errors, and a clear `free_string` ownership contract. I verified all the `llmfit-core` calls and they line up correctly, so the implementation itself is in good shape. 🎉

Just a few minimum updates before merge:

1. **Trim the `Cargo.lock` churn.** A stray `cargo update` bumped ~90 unrelated crates (strum, lru, shlex, quick-xml, serde_yml backend, ratatui features…), which changes the TUI's dependency tree. Please `git checkout main -- Cargo.lock` and regenerate so it only gains `llmfit-ffi` + `once_cell`.

2. **Follow workspace conventions in `llmfit-ffi/Cargo.toml`.** Use `version.workspace = true` and `edition = "2024"` (currently hardcoded to `0.1.0` / `2021`). Right now `llmfit_version()` reports `0.1.0` instead of the real `0.9.31`.

3. **Panic safety:** the README says entry points are wrapped in `catch_unwind`, but only `llmfit_create_context` is. Either wrap the other three getters too, or tweak the doc to match.

Minor: `once_cell` is declared but unused — can be dropped.

No need for the wrapper examples/tests to land here. Thanks for this — looking forward to merging once the lockfile is cleaned up!

AlexsJones:
Thanks of this! same as @three-foxes-in-a-trenchcoat also needs a cargo fmt please

jimmckeeth:
Thanks for the feedback. This is my first serious work in Rust, so a lot to learn.

AlexsJones:
Please address checks

Issue #292, opened 2026-03-26
For the given recommendation model, can you provide the corresponding llama.cpp runtime parameters?


MingManMan:
llmfit provides the adaptability of various models to the machine and gives scores based on multiple indicators. Could you provide the invocation parameters of llmfit for the machine and model to reproduce the test results?

zhaoyu190:
同问 目前在此芯 32g上部署qwen ，llmfit显示30token/s 但我用llmfit的命令用llamacpp运行只有2token/s

KD-MM2:
is this even real 😬 ?
<img width="1920" height="1075" alt="Image" src="https://github.com/user-attachments/assets/54f28e80-e394-4cd0-a463-8657dffa4d3f" />

AlexsJones:
This issue was the clearest statement of a real problem: a tok/s number you can't reproduce is a number you can't trust. Apologies that it sat unanswered — it directly shaped what we just shipped (now on `main`, in the next release):

**1. Every estimate now ships its inputs.** `llmfit info "<model>"` prints an **Estimate Basis** section: the estimation method, the exact GPU/DDR bandwidth assumed, the efficiency factor, and the context length the number models. The same data is in `--json` output as `estimate_basis`.

**2. Every estimate comes with a verification recipe.** The same section prints the commands that measure *the same quantity the estimate models* (single-request generation throughput):

```
llmfit bench "<model>"                                  # against your running provider
llama-bench -m <model.gguf> -ngl 99 -p 512 -n 128       # compare the tg128 row
```

The runtime command llmfit suggests for *running* a model (`llama-cli -hf ... -ngl ... -c ...`) is shown in `--json` as `llamacpp_command` — but note it's a *launch* command, not a benchmark; interactive chat timing includes prompt processing and sampling overhead the estimate deliberately doesn't model.

**3. Unknown hardware is now labeled, not guessed.** @zhaoyu190's 30 → 2 tok/s case is the failure mode where the GPU/SoC isn't in our bandwidth table (a Cix ARM chip won't be), so the estimate fell back to a coarse per-backend constant. The output now says exactly that — *"per-backend heuristic constant — GPU not in the bandwidth table, so expect a wide error band"* — and asks you to report the chip so we can add real bandwidth data. @zhaoyu190, if you post the SoC model (or `llmfit doctor` output), we'll add it.

[...]

PR #640, opened 2026-06-17
feat: add `llmfit budget` subcommand and web Budget Advisor panel

### LINKED ISSUE #541: Feature roadmap: top three needed features for llmfit
We need to prioritize the next set of core capabilities for llmfit. Based on community feedback and current gaps, the top three features to implement are:

1. **Kubernetes deployment support** – Enable llmfit to be deployed as a scalable service on Kubernetes clusters, with Helm chart and operator integration. This will simplify production deployments and align with the project's containerization goals.
2. **Built‑in automatic benchmark module** – Provide an out‑of‑the‑box benchmark suite that runs common model workloads, captures latency/TPS, and stores results for later recommendation/calibration. This feature underpins the "best model" scoring and helps users evaluate hardware choices.
3. **ONNX model format support** – Add import and inference pipelines for ONNX models, broadening the model ecosystem beyond llama.cpp and GGUF. This will attract users with existing ONNX assets and improve interoperability.

These features address the most requested use‑cases and lay the groundwork for future extensions such as hardware‑aware model selection and advanced UI integrations.

*Please add any additional high‑priority items or adjust the ordering as needed.*

Clemente-H:
## Summary

Adds a hardware budget advisor feature that lets users discover which hardware configurations fit within a purchase budget and see which LLM models each configuration can run — using the same fit pipeline as `llmfit fit`.

- **`llmfit budget <MAX_PRICE>`** — new CLI subcommand. Filters a built-in hardware catalog by price and shows ranked LLM model lists per configuration, with `--limit`, `--min-fit`, `--json`, and `--csv` support.
- **`llmfit-core/src/hardware_catalog.rs`** — static catalog of ~16 real-world hardware configurations (mini PCs, Apple Silicon, NVIDIA/AMD GPUs, ROCm cards) with prices, VRAM, RAM, and CPU specs. `configs_within_budget(max_price)` returns matching entries. Designed to be easy to extend.
- **`SystemSpecs::synthetic()`** — new constructor in `hardware.rs` for building simulated specs without touching real hardware detection. Used internally by `HardwareConfig::to_specs()`.
- **`GET /api/v1/budget?max_price=N&limit=5&min_fit=marginal`** — new REST endpoint in `serve_api.rs`.
- **Budget Advisor panel** in the web dashboard (`llmfit-web`) — form with budget/limit/fit-level controls, renders a per-configuration table of top models using the existing CSS design system.

### Example

```
$ llmfit budget 800
=== Hardware Budget: up to $800 ===
10 configuration(s) found

┌─ Mac mini M4 Pro (24 GB) — $799 │ 24 GB VRAM │ 24 GB RAM │ 14 cores │ Metal
│  Note: Unified memory; 24 GB pool, excellent perf/watt
...
┌─ RTX 3090 (24 GB) PC — $700 │ 24 GB VRAM │ 32 GB RAM │ 12 cores │ CUDA
...

$ llmfit budget 800 --json
$ llmfit budget 1500 --limit 10 --min-fit good
```

Closes #541 (hardware-aware model selection)

## Test plan

- [ ] `cargo test -p llmfit-core` — `hardware_catalog` unit tests pass (budget filter, empty budget, to_specs no panic)
- [ ] `llmfit budget 800` — shows ≥1 config with ranked model table
- [ ] `llmfit budget 800 --json` — valid JSON with `budget_usd` and `configurations` array
- [ ] `llmfit budget 0` — prints "no configurations" message cleanly
- [ ] `llmfit serve` → open browser → click "💰 Budget advisor" → enter budget → results appear
- [ ] `curl "http://127.0.0.1:8787/api/v1/budget?max_price=800&limit=3"` — returns JSON

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -1612,6 +1612,29 @@ impl SystemSpecs {
         self
     }
 
+    /// Build a fully synthetic `SystemSpecs` for a hypothetical hardware
+    /// configuration (used by `llmfit budget` to simulate catalog hardware).
+    /// No real hardware detection is performed.
+    pub fn synthetic(ram_gb: f64, cpu_cores: usize, backend: GpuBackend) -> Self {
+        let unified_memory = matches!(backend, GpuBackend::Metal);
+        SystemSpecs {
+            total_ram_gb: ram_gb,
+            available_ram_gb: ram_gb * 0.9,
+            total_cpu_cores: cpu_cores,
+            cpu_name: "Simulated".to_string(),
+            has_gpu: false,
+            gpu_vram_gb: None,
+            total_gpu_vram_gb: None,
+            gpu_name: None,
+            gpu_count: 0,
+            unified_memory,
+            backend,
+            gpus: Vec::new(),
+            cluster_mode: false,
+            cluster_node_count: 0,
+        }
+    }
+
     pub fn display(&self) {
         println!("\n=== System Specifications ===");
         println!("CPU: {} ({} cores)", self.cpu_name, self.total_cpu_cores);
--- a/llmfit-core/src/hardware_catalog.rs
+++ b/llmfit-core/src/hardware_catalog.rs
@@ -1,0 +1,244 @@
+use crate::hardware::{GpuBackend, GpuInfo, SystemSpecs};
+
+/// A hardware configuration from the purchase catalog.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct HardwareConfig {
+    /// Human-readable name (e.g. "RTX 4070 Ti PC")
+    pub name: &'static str,
+    /// Approximate price in USD at time of catalog publication
+    pub price_usd: u32,
+    /// GPU VRAM in GB (None = CPU-only / integrated GPU)
+    pub vram_gb: Option<f64>,
+    /// System RAM in GB
+    pub ram_gb: f64,
+    /// CPU thread / core count
+    pub cpu_cores: usize,
+    /// GPU acceleration backend
+    pub gpu_backend: GpuBackend,
+    /// Short note for the user (e.g. "Unified memory", "Dual GPU")
+    pub notes: &'static str,
+}
+
+impl HardwareConfig {
+    /// Build a `SystemSpecs` that represents this hardware configuration.
+    /// The specs are synthetic (not from real hardware detection) and are
+    /// used to run the same fit pipeline as the regular `llmfit fit` command.
+    pub fn to_specs(&self) -> SystemSpecs {
+        let mut specs = SystemSpecs::synthetic(self.ram_gb, self.cpu_cores, self.gpu_backend);
+        if let Some(vram) = self.vram_gb {
+            let unified = matches!(self.gpu_backend, GpuBackend::Metal);
+            specs.gpus.push(GpuInfo {
+                name: self.name.to_string(),
+                vram_gb: Some(vram),
+                backend: self.gpu_backend,
+                count: 1,
+                unified_memory: unified,
+            });
+            specs.has_gpu = true;
+            specs.gpu_vram_gb = Some(vram);
+            specs.total_gpu_vram_gb = Some(vram);
+            specs.gpu_name = Some(self.name.to_string());
+            specs.gpu_count = 1;
+            specs.unified_memory = unified;
+        }
+        specs
+    }
+}
+
+/// The built-in hardware catalog.
+/// Prices are approximate USD street prices as of mid-2025. Add new entries
+/// in ascending price order so `configs_within_budget` preserves that order.
+pub static CATALOG: &[HardwareConfig] = &[
+    // ── CPU-only / mini PCs ────────────────────────────────────────────────
+    HardwareConfig {
+        name: "Beelink SEi12 Mini PC",
+        price_usd: 200,
+        vram_gb: None,
+        ram_gb: 16.0,
+        cpu_cores: 8,
+        gpu_backend: GpuBackend::CpuX86,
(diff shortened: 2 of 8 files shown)


AlexsJones (commented):
Thanks @Clemente-H — this is a substantial, well-structured contribution: the catalog module has tests, the subcommand reuses the existing fit pipeline instead of reinventing it, and the JSON/CSV output follows the house style.

Before it can move forward there's a **scope question that's the maintainer's call**, plus a few technical points.

**Scope (for @AlexsJones):** this bakes a curated hardware *price* catalog into the compiled binary. Prices go stale fast (the catalog itself says "as of mid-2025" — already a year old, e.g. RTX 3090 at $700), vary by region and currency, and every price refresh becomes a code change and release. If the feature direction is wanted, it may fit better as a data file (bundled or fetched JSON with a "last updated" stamp) so the catalog can be refreshed without a release.

Technical notes in the meantime:
1. The `NVIDIA L4 (24 GB) Cloud Instance` entry prices a **monthly cloud cost on a purchase-price axis** (its own note admits this). It will distort budget comparisons — drop it, or make cloud options a separate concept.
2. `--min-fit` silently maps any invalid value to `marginal`; rejecting unknown values with an error would match the CLI's behavior elsewhere.
3. The `2× RTX 3090 (48 GB total)` entry is modeled in `to_specs()` as a **single** GpuInfo with 48 GB (`count: 1`). If the fit pipeline treats one 48 GB device differently from 2×24 GB tensor-parallel (it does for per-device layer fits), this overstates what that config can run.

Parking this for the maintainer's direction on the catalog approach before further iteration — the implementation itself is in good shape.

AlexsJones:
Thanks again for this contribution, @Clemente-H — and for your patience while Alex weighed the direction question.

His decision: **budget/price advice is going on the roadmap for the future rather than landing now**, because it intersects with token pricing — he wants to consider the two together rather than take a hardware price catalog in isolation today.

That's a timing/direction call, not a reflection on the implementation: the subcommand is tested and reuses the fit pipeline properly, which is exactly the right shape. When the roadmap item comes up, this PR will be the reference point for the hardware-catalog side.

AlexsJones (changes requested):
Thanks for this, @Clemente-H — it's a well-built feature and a genuinely useful one. The shape is right: a static catalog with prices, a `SystemSpecs::synthetic()` constructor so the same fit pipeline runs against hypothetical hardware, plain/CSV/JSON output on the CLI, and a matching `/api/v1/budget` endpoint + `BudgetPanel` that reuses the existing dashboard styles. Tests on the catalog are there, and all three Test Suites, Clippy and Cargo Check are green. A few things before I'd recommend it for merge.

**Blocking**

**1. CI is red on Rustfmt.** Only that leg — everything else passes. `cargo fmt` should clear it.

**2. The CLI and the web panel compute different numbers for the same query.** This is the one that matters. The two entry points build their fits through different engines:

- CLI `run_budget` (main.rs) → `analysis::build_model_fits(&db, &specs, &installed, …)`, which runs `apply_local_calibration` and layers in measured/community `tok/s` lookups keyed to the config's specs (analysis.rs:168-192).
- API `budget` (serve_api.rs) → hand-rolls the pipeline with raw `ModelFit::analyze_with_context_limit`, no calibration and no measured-tps lookup.

So `llmfit budget 800` in the terminal and the Budget Advisor panel for the same `$800` will show different `estimated_tps` for the same hardware+model — the CLI's are community-calibrated for that hardware class, the API's are formula-only. Please route both through one helper. Cleanest is to have the API call `build_model_fits` exactly as the CLI does; or extract a `budget_fits(config, min_level, limit) -> Vec<ModelFit>` that both call, so the filter, the ranking, the MLX handling and the calibration can't drift apart. (This is the same two-implementations-of-one-thing trap that #759 just flagged for the JSON serializers — worth not opening a second front on it.)

**Non-blocking**

- **The extra MLX retain in the API path looks redundant.** `if !is_apple_silicon { fits.retain(|f| !f.model.is_mlx_only()) }` runs *after* `backend_compatible`, which already rejects `is_mlx_model()` on anything that isn't `Metal` + unified memory (fit.rs:953-954). Unless `is_mlx_only` and `is_mlx_model` differ in a way that matters here, the retain never removes anything — worth dropping, or a comment on why it's needed. (Folding into the shared helper from point 2 makes this moot.)
- **The min-fit filter is copy-pasted three times** — the `match (min_level, f.fit_level) { … }` block appears in the CLI JSON path, the CLI plain path, and the API handler. Lift it into one place (a `FitLevel::admits(other: FitLevel) -> bool`, say) so a future change to the fit ladder only has to land once.
- **`estimated_tps` on hardware nobody owns.** Since a given catalog config may or may not have calibration data behind it, a budget row's `tok/s` can be either community-anchored or a bare formula estimate, with no way for the reader to tell which. `llmfit fit` solved exactly this with `estimate_basis` (#292/#754) — surfacing that here would keep the budget numbers as honest as the ones `fit` reports.
- **Catalog vintage.** Prices are hardcoded "mid-2025 street prices"; it's now mid-2026, so they're already ~a year stale, and there's no date in the output for a reader to judge that by. Consider printing a `catalog as of <date>` line (and a field in the JSON). Separately, the `NVIDIA L4 … approximate monthly cost as purchase price` entry mixes a rental cost into a purchase catalog — that one probably deserves a caveat or its own treatment so a `$2000 budget` doesn't silently equate a month of L4 rental with owning a 4090.
- **`Closes #541`.** #541 is the "Feature roadmap: top three needed features" thread, not a hardware-selection issue — merging this would auto-close a roadmap tracker. `Refs #541` (or a dedicated issue) is probably what you want.

None of this is structural — the design is sound and the hard parts (synthetic specs, reusing the fit pipeline, the web wiring) are done well. Fix 1 and 2 and I think this is close.

PR #641, opened 2026-06-17
Scrape Hugging Face model licenses

razzeee:
Populate model license metadata from Hugging Face card data and license tags. Prefer named licenses when Hugging Face reports a generic other license, and preserve existing license and GGUF source metadata when a refresh response omits it.

### DIFF
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -648,10 +648,20 @@ impl std::fmt::Display for KvQuant {
 /// Returns true if a model's license matches any in the comma-separated filter string.
 /// Models without a license never match.
 pub fn matches_license_filter(license: &Option<String>, filter: &str) -> bool {
-    let allowed: Vec<String> = filter.split(',').map(|s| s.trim().to_lowercase()).collect();
+    let allowed: Vec<String> = filter
+        .split(',')
+        .map(|s| s.trim().to_lowercase())
+        .filter(|s| !s.is_empty())
+        .collect();
+
     license
         .as_ref()
-        .map(|l| allowed.contains(&l.to_lowercase()))
+        .map(|licenses| {
+            licenses
+                .split(',')
+                .map(|s| s.trim().to_lowercase())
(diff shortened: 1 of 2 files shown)


AlexsJones:
Reviewed — overall a solid, additive change. The Rust side already wired `license` through filtering/display/MCP but the data was 0% populated on `main`; this fills it to 623/5444 models, with no models removed in the regeneration. The merge-preservation logic is correct (`fresh_by_name` holds the same object refs that are in `results`, so the license/gguf_sources backfill mutates the right objects).

A few points before merge:

**1. Extra ~1000 API round-trips in `_build_discovered_model` (main concern).**
```python
license_info = fetch_model_info(repo_id) or listing
```
The docstring just above says metadata "comes from the listing data already obtained via `expand=safetensors`", but this adds one full GET per discovered model. The discovery listing already carries `tags` (you filter on `m.get("tags", [])`), so `extract_license(listing)` would resolve `license:` tags with zero extra requests. The fetch only buys `cardData` (named licenses + the `other` fallback) — cleaner to add `expand[]=cardData` to the listing query and pass `listing` directly, avoiding 1000 sequential GETs per scrape. (Worth confirming the list API honors `expand[]=cardData`.)

**2. Multi-license join breaks the license filter (edge case).**
```python
return ",".join(licenses)   # e.g. "apache-2.0,mit"
```
`matches_license_filter` compares the whole stored string against each allowed token, so `"apache-2.0,mit"` matches neither `apache-2.0` nor `mit`. Rare (list-valued licenses are uncommon), but worth a note — either split on the Rust side or store the first license.

**3. Divergent extraction paths.** `update.rs:515` does its own license extraction (top-level `hf.license` → tags, no lowercasing, no `other` fallback) while this path lowercases and prefers named licenses. Filtering lowercases on both sides so matching is fine, but displayed values can differ depending on which path produced the model — flagging as future drift risk, not blocking.

Approve with nits — #1 is the one I'd like addressed (or justified) before merge.

PR #642, opened 2026-06-19
feat: discover TTS language metadata in scraper

### LINKED ISSUE #636: TTS capability
Hey there,

just wanted to reach out and ask, if you are planning to also add support for tts models or if that is out of scope.

Cheers

razzeee:
Closes #636


AlexsJones:
Thanks for this; 
One question: `discover_trending_models` now splits `limit` evenly across all 7 pipelines instead of filling `limit` total, so fewer `text-generation` models get discovered per run for the same `limit`. Was the `limit` used for regeneration bumped to account for the reserved TTS/other-pipeline slices, so mainstream-LLM coverage didn't shrink?

razzeee:
Good catch. You’re right that the fair split changed the semantics of `--discover-limit`: with 5 discovery pipelines, `--discover-limit 250` gave text-generation only ~50 fresh slots instead of preserving the full mainstream LLM budget.

I updated the scraper so `text-generation` now keeps the full `--discover-limit`, and the other pipelines use additive side quotas (`limit // len(DISCOVER_PIPELINES)`). That keeps TTS/audio discovery from starving mainstream LLM coverage.

AlexsJones:
Thanks @razzeee — reviewed this and the code is in good shape: builds clean, all 377 `llmfit-core` tests pass, the new `Capability::Tts` correctly also carries the broad `Audio` capability, the new `languages` field is serialization-backward-compatible, and `CACHE_VERSION` is bumped (3→4) so stale caches are discarded. The Rust/Python `normalize_language` helpers mirror each other and have solid tests.

The only thing blocking merge is that `main` has moved on (license scraping landed in `c499c00`), so this now has conflicts. Could you rebase onto `main` and resolve:

1. **`scripts/scrape_hf_models.py`** — one tiny conflict in the existing-DB merge loop. `main`'s side is a strict superset (it preserves both `license` and `gguf_sources`), so just take `main`'s version and you're done.
2. **`data/hf_models.json`** + **`llmfit-core/data/hf_models.json`** — 260+ conflicts each, but these are regenerated scraper output. Don't hand-merge them; just **re-run the scraper** after rebasing so they regenerate cleanly on top of `main`'s data.

One non-blocking note for your awareness: `update_model_cache` now fetches each of the 4 `ACCEPTED_PIPELINES` at the full limit for both the trending and downloads passes, so it makes ~4× the HF API calls the old single `text-generation` fetch did. Looks intentional given the per-pipeline discovery quotas, just flagging the extra rate-limit/runtime cost.

Once it's rebased + the data files regenerated, this is good to merge. Thanks for adding TTS support!

razzeee:
Updated and data files seem fine too me, but please double check.

razzeee:
I'm working on some more additions

razzeee:
Rebase and good to go, please check the generated file for errors

razzeee:
Rebased, struggling a bit to redo the catalogue data as I need to be somewhere else right now. Will do it later.

razzeee:
Successfully generated a file, but some of my likes/download counts seem wrong

AlexsJones:
@razzeee your instinct was right — the likes/download counts in the regenerated catalog were wrong, and it wasn't your fault. I dug in, found the causes, and pushed two commits to your branch to get this over the line:

## What was wrong

1. **87 existing models had downloads/likes zeroed** (e.g. `nomic-ai/nomic-embed-text-v1.5`: 18.7M → 0). Your scrape ran without an `HF_TOKEN`, so every gated model fell back to the hardcoded `FALLBACKS` stubs — which carry `hf_downloads: 0, hf_likes: 0` — and the merge loop only backfilled `license`/`gguf_sources` from the existing DB, not popularity fields. This was a latent scraper bug; your regeneration just exposed it.
2. **195 newly discovered entries had 0 downloads/likes** — listings from the non-download sort strategies (trending/likes) don't include those fields, and `_build_discovered_model` used the listing directly. Your `require_downloads_floor` change correctly let those models through, but they were then written with zeros.
3. **None of the 212 TTS entries had `languages`** — the `facebook/mms-tts-*` family (144 entries) declares no language metadata via the API at all; the language only exists as the ISO-639-3 suffix in the repo name.

## What I pushed

- **b2b35d6** (scraper): merge loop now backfills `hf_downloads`/`hf_likes`/`release_date`/`languages` from the existing DB when the fresh entry lacks them; `_build_discovered_model` fetches full model info when the listing omits downloads/likes; `infer_languages` derives MMS languages from the `mms-tts-<code>` repo suffix.
- **48c1556** (data): restored the 87 clobbered entries from the previous catalog, live-fetched real counts for all 239 zeroed discoveries from the HF API, and filled languages for the MMS models — **182 of 212 TTS entries now carry language metadata** (the remaining 30 genuinely declare none upstream).

Verified locally: `cargo build`, all 5 test suites (421 core tests), `cargo fmt --check` pass; `llmfit info hexgrad/Kokoro-82M` shows TTS use-case, F16, and runtime correctly marked unsupported; no download-count regressions vs main remain (the two apparent drops left, jina-code-embeddings and a DeepSeek AWQ repack, are genuine fresh HF values — rolling 30-day windows).

## One open question for @AlexsJones

Discovery pulled in **144 `facebook/mms-tts-*` per-language models** (one repo per language, many with tiny download counts). They're data-correct now, but you may want a curation call on whether the catalog should carry all of them or cap the family — happy to filter if you'd rather.

AlexsJones (approved):
Confirmed fully additive: 5473 → 5843 catalog entries, none removed, no popularity metadata regressions vs main. Scraper hardening means future regens (even unauthenticated) can't clobber existing data. CI green on all platforms. Thanks @razzeee for driving this through several rebases — TTS discovery with language metadata is a great addition.

PR #651, opened 2026-06-23
fix(ollama): ignore cloud models in installed detection

### LINKED ISSUE #619: Ollama cloud models cause false "installed" detection for unrelated HuggingFace models
llmfit marks HuggingFace models as installed: true via Ollama when a cloud-hosted Ollama model with a matching family name exists, even though the model is not actually installed locally.

**Steps to Reproduce**

     1. Have Ollama cloud models pulled (e.g., qwen3-coder:480b-cloud)
     2. Run llmfit fit --json
     3. Observe that ig1/Qwen3-Coder-30B-A3B-Instruct-NVFP4 shows "installed": true, "provider": "Ollama"

**Expected Behavior**

Cloud models should not be treated as locally installed. A model like qwen3-coder:480b-cloud is a remote-hosted model with size: 0 in the Ollama API response — it is not the same as having qwen3-coder:30b-a3b downloaded locally.

**Actual Behavior**

llmfit matches on the model family name (qwen3-coder) and considers any HuggingFace model mapped to that family as "installed" if any Ollama tag exists for it, regardless of whether it's a local download or a cloud model.

**Root Cause**

llmfit queries GET /api/tags and builds a set of installed model families. The Ollama API response includes enough information to distinguish cloud from local:

     - Cloud models have "size": 0
     - Cloud model tags end with -cloud (e.g., qwen3-coder:480b-cloud)

Neither signal is currently used by llmfit's detection logic.

**Suggested Fix**

     When building the installed-models set from Ollama, skip entries where:
     - size == 0 (no local weights)
     - The tag suffix is -cloud

     This would prevent false positives without affecting legitimate local model detection.
     Environment

     - llmfit 0.9.30 (Homebrew)
     - Ollama with cloud models pulled
     - macOS Apple Silicon

**Screenshot** 

Image

snvtac:
Fixes #619.

## Problem

Ollama's `/api/tags` can include cloud-hosted models such as `qwen3-coder:480b-cloud`. The installed-model builder was inserting every tag and its family stem into the installed set, so a cloud tag could insert `qwen3-coder` and incorrectly mark unrelated HuggingFace Qwen3-Coder variants as locally installed.

## Fix

- Parse the optional Ollama `size` field from tag responses.
- Centralize Ollama installed-set construction so `detect_with_installed` and `installed_models_counted` share the same behavior.
- Skip cloud entries before inserting tag names or family stems when the tag suffix is `-cloud` or `size == 0`.
- Keep local-model detection tolerant of older/missing `size` fields when the tag does not look like a cloud model.
- Add unit tests covering the #619 false positive and missing-size local model behavior.

## Validation

- `git diff --check`
- `cargo test -p llmfit-core test_ollama -- --nocapture`
- `make test`
- `make clippy` currently fails on existing upstream `main` warnings outside this PR's scope, including `llmfit-core/src/fit.rs`, `llmfit-core/src/hardware.rs`, `llmfit-core/src/models.rs`, and `llmfit-core/src/bench.rs`.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -181,7 +181,7 @@ impl OllamaProvider {
     /// is configured, the fallback is tried and—if successful—adopted as the
     /// provider's base URL for all subsequent requests (pull, show, …).
     pub fn detect_with_installed(&mut self) -> (bool, HashSet<String>, usize) {
-        let mut set = HashSet::new();
+        let set = HashSet::new();
 
         let primary_ok = ureq::get(&self.api_url("tags"))
             .config()
@@ -215,14 +215,7 @@ impl OllamaProvider {
         let Ok(tags): Result<TagsResponse, _> = resp.into_body().read_json() else {
             return (true, set, 0);
         };
-        let count = tags.models.len();
-        for m in tags.models {
-            let lower = m.name.to_lowercase();
-            set.insert(lower.clone());
-            if let Some(family) = lower.split(':').next() {
-                set.insert(family.to_string());
-            }
-        }
+        let (set, count) = build_ollama_installed_set(tags.models);
         (true, set, count)
     }
 
@@ -230,27 +223,18 @@ impl OllamaProvider {
     /// The HashSet may have fewer entries than 2*count due to family-name deduplication,
     /// so `len() / 2` is unreliable for counting models.
     pub fn installed_models_counted(&self) -> (HashSet<String>, usize) {
-        let mut set = HashSet::new();
         let Ok(resp) = ureq::get(&self.api_url("tags"))
             .config()
             .timeout_global(Some(std::time::Duration::from_secs(5)))
             .build()
             .call()
         else {
-            return (set, 0);
+            return (HashSet::new(), 0);
         };
         let Ok(tags): Result<TagsResponse, _> = resp.into_body().read_json() else {
-            return (set, 0);
+            return (HashSet::new(), 0);
         };
-        let count = tags.models.len();
-        for m in tags.models {
-            let lower = m.name.to_lowercase();
-            set.insert(lower.clone());
-            if let Some(family) = lower.split(':').next() {
-                set.insert(family.to_string());
-            }
-        }
-        (set, count)
+        build_ollama_installed_set(tags.models)
     }
 
     /// Best-effort check that a tag exists in Ollama's remote registry.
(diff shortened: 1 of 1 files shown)


AlexsJones:
Thanks for this, @snvtac 🙏 — flagging for you that #650 landed on `main` and closed the underlying issue (#619), so the cloud-model false-positive this targets is now fixed on `main`. That makes this PR a duplicate of merged work, so you may want to close it.

One thing your version did that's worth preserving: modelling the Ollama `size` field as `Option<u64>` (rather than `#[serde(default)]`), so older Ollama servers that omit `size` don't get every local model misclassified as cloud — with a test for it. If you're up for it, a small focused follow-up PR for just that tolerance would be a welcome hardening on top of what merged. Either way, appreciated the contribution.

---

AlexsJones:
Hi @snvtac — following up on the note above rather than repeating it.

Short version: #650 landed on `main` and closed #619, so the cloud-model false positive this PR targets is already fixed and this branch is a duplicate as it stands. No action needed on the bulk of the diff.

The part still worth having is small and specific. Merged `main` models the field as:

```rust
// llmfit-core/src/providers.rs:266
#[serde(default)]
size: u64,
```

That means an Ollama server which omits `size` deserializes to `0`, and `is_cloud()` (providers.rs:274-277) short-circuits on `self.size == 0` — so on an older Ollama, *every* local model gets classified as cloud and dropped from the installed set. A false negative here is worse than a false positive: users just quietly stop seeing models they actually have.

Your version already gets this right with `size: Option<u64>` and `self.size == Some(0)`, plus `test_ollama_cloud_detection_preserves_local_models_without_size` covering it. `main` has no equivalent test — `test_ollama_is_cloud_detection` only exercises explicit sizes.

If you're up for it, a fresh branch off current `main` with just:

1. `size: Option<u64>` replacing `#[serde(default)] size: u64`
2. `is_cloud()` matching on `Some(0)` so a missing size stays local
3. your missing-size test, carried over as-is

...would be a welcome hardening and should be a quick review. Rebasing this branch would mostly be conflict resolution against work that's already in, so a clean small PR is likely less effort for you.

One unrelated thing worth knowing: CI never ran here — no checks are recorded at all. That's not your doing. The workflow triggers match this diff fine; it's the fork-approval gate for first-time contributors, and the run was never approved. On the follow-up PR I'll make sure checks get approved so you actually get signal.

Thanks for the careful work on this one — the `Option` detail is a real bug you caught, and it's the kind of thing that's easy to miss.

AlexsJones (commented):
Thanks for this, and sorry it sat so long.

The short version is that the fix reached main by another route. #619 was closed on 2026-06-25, two days after you opened this, and `providers.rs` now carries its own `OllamaModel::is_cloud` with the same `-cloud` suffix plus zero-size rule, inside a `build_ollama_installed_set` that has since absorbed #861: a sized tag like `qwen3:8b` contributes only its tag now, not the bare family stem, because one `qwen3:8b` was marking 238 catalog entries installed. Your branch inserts the family stem for every non-cloud model, so rebasing it would walk that back. I would rather close this than ask you to redo it against a moving target.

One thing in your version is better than what main has, and I would like to take it. You typed the field as `size: Option<u64>` and treated `None` as not-cloud. Main uses `#[serde(default)] size: u64`, so a tags response that omits `size` deserializes as `0`, `is_cloud()` then returns true for every entry, and the installed set comes back empty with no error anywhere. That is precisely the case your `test_ollama_cloud_detection_preserves_local_models_without_size` pins, and nothing on main pins it.

If you want to send that on its own (the `Option<u64>` field, `size == Some(0)`, and that one test) I will take it. If you would rather not, say so and I will pick it up and credit you.

Either way, thank you for the write-up and the tests. The reasoning in this PR is what made the gap easy to spot.

PR #658, opened 2026-06-27
feat: add NCAI VAETKI model to curated model database

wklee610:
- Add the 112.2B NCAI VAETKI model to the curated HuggingFace model list:
  - `NC-AI-consortium-VAETKI/VAETKI`
- Add generated entries to both `data/hf_models.json` and `llmfit-core/data/hf_models.json`
- Map `NC-AI-consortium-VAETKI` to the `NCAI` provider label


AlexsJones:
Took a pass through this — mechanically clean, one thing to confirm before merge.

**Checks out:**
- Deserializes fine: all required `HfModelEntry` fields present and correctly typed; extra fields (`is_moe`, `license`, `active_parameters`, …) are recognized, `_discovered`/`pipeline_tag` are ignored.
- Both `data/hf_models.json` and `llmfit-core/data/hf_models.json` updated identically and still valid JSON.
- Model is authentic (exists on HF, MIT, created 2025-12-16, 62 likes).
- `extract_provider` lowercases the org, so the new `nc-ai-consortium-vaetki` key resolves to `NCAI` correctly.
- Architecture fields match `config.json` exactly (layers 48, heads 24, head_dim 64, hidden 3072, vocab 137216, n_routed_experts 128, num_experts_per_tok 8, moe_intermediate 2048). `parameters_raw` rounds to "112.2B"; `num_key_value_heads: 24` is a safe default since config omits it.

**⚠️ One question — `context_length`:**
The new `CONTEXT_LENGTH_OVERRIDES` entry forces **32768**, but the model's `config.json` declares `max_position_embeddings: 131072`. So this deliberately downgrades the advertised context to ~1/4 of what the model claims (without the override the scraper would record 131072).

Is 32768 intentional — e.g. the model card recommends a 32k effective window with 131k being RoPE-extended? If so, could you add a one-line note next to the override explaining why, so it isn't read as a typo and a future scrape doesn't "correct" it back to 131072. If it's not intentional, it's under-reporting the model's context.

**Minor (non-blocking):** `hf_downloads: 27` is already stale (live is 28) — inherent to committing scraped counts, not worth churning.

wklee610:
Thanks for the careful review.

Yes, `32768` is intentional. I followed the model card here: it lists `Context Length: 32k tokens`, and the training procedure also shows the final context length as `32768`. The `config.json` value (`max_position_embeddings: 131072`) looks higher, but I didn’t want to over-advertise the effective context window beyond what the model card states.

I added a one-line comment next to the override to make this explicit and avoid future confusion. Agreed on `hf_downloads`; I’ll leave it as-is since that count is expected to drift.

Issue #112, opened 2026-02-27
Feature: Benchmark-calibrated recommendations (measured TPS/latency)


AissamDjahnine:
Related: [#80](https://github.com/AlexsJones/llmfit/issues/80)

### Problem
Current ranking is estimate-based. In practice, models with similar estimated fit can have very different real performance on the same machine.

### Proposal (MVP)
Add an optional calibration step that runs short local benchmarks for top candidates, then re-ranks using measured results.

### Scope
- Keep existing recommendation flow as default.
- Add optional benchmark mode for local calibration.
- Start with one runtime/provider (suggested: Ollama).

### Acceptance Criteria
- [ ] New command: `llmfit benchmark --top <N>`
- [ ] New command/flag: `llmfit recommend --calibrated`
- [ ] Benchmark captures at least:
  - [ ] tokens/sec
  - [ ] first-token latency
  - [ ] completion latency (or total runtime)
- [ ] Results cached locally (JSON) by machine profile + model + runtime
- [ ] Re-running benchmark reuses cache unless `--refresh` is set
- [ ] `recommend --calibrated` uses measured metrics when cache exists, otherwise falls back to current estimates
- [ ] Output clearly labels `estimated` vs `measured` ranking mode

### Non-goals (MVP)
- Cross-machine result sharing
- Multi-provider support in first iteration
- Full benchmark suite customization

### Suggested CLI
```sh
llmfit benchmark --top 5
llmfit benchmark --top 5 --refresh
llmfit recommend --calibrated
```

three-foxes-in-a-trenchcoat:
Hey @AissamDjahnine! 🦊

This is an excellent feature proposal - very well thought out with clear acceptance criteria and appropriate MVP scoping.

**Why this is valuable:**
- Estimates are based on constants and heuristics, but real-world performance varies significantly based on CPU/GPU architecture, memory bandwidth, thermal throttling, etc.
- Measured benchmarks would provide ground truth for re-ranking recommendations
- Cache-based approach keeps it practical (don't re-benchmark every time)

**Implementation considerations:**

1. **Ollama integration**: Perfect starting point since we already have Ollama detection (`providers.rs`). We can use `ollama run` with a standardized prompt to measure:
   - Time to first token (TTFT)
   - Tokens per second (avg throughput)
   - Total completion time

2. **Cache location**: Suggest `~/.config/llmfit/benchmarks.json` or `~/.cache/llmfit/benchmarks/` with schema:
   ```json
   {
[...]

PR #681, opened 2026-07-03
chore: refresh localmaxxing benchmark cache (384 → 448 measurements)

AlexsJones:
Authenticated refresh of `llmfit-core/data/benchmark_cache.json` (previously 10 days stale). 27 presets, 448 rows — RTX 3090 alone now has 120 measured runs.

Beyond the fresher leaderboard fallback, this is the raw material for the estimate-calibration loop discussed in #112/#119: measured `tokSOut` per (hardware preset × model × quant) to validate `estimate_tps` against. A follow-up will add a validation harness in core that replays these rows through the estimator using each preset's bandwidth profile.

### DIFF
--- a/llmfit-core/data/benchmark_cache.json
+++ b/llmfit-core/data/benchmark_cache.json
@@ -1,5 +1,5 @@
 {
-  "scraped_at": "2026-06-23T11:17:24Z",
+  "scraped_at": "2026-07-03T05:29:36Z",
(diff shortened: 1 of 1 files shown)


AlexsJones:
Superseded by #682 — the same cache refresh commit is included there alongside the weekly automation and the calibration harness it feeds.

PR #686, opened 2026-07-03
fix(docker): drop COPY of removed root data/ directory

### LINKED ISSUE #687: Docker tag builds broken since v0.9.36 (stale COPY data/); pages deploy flaked on main
## Docker (real breakage)

The Docker image builds for `v0.9.36` and `v0.9.37` failed instantly with:

```
ERROR: failed to build: failed to solve: failed to compute cache key: ... "/data": not found
```

**Root cause:** cf64a5b / fe89b82 removed the root `data/` directory (it held byte-identical duplicates of `llmfit-core/data/`, which is what the build actually embeds via `include_str!`), but `Dockerfile` still ran `COPY data/ ./data/`. Every tag build since then died at that step, so no `0.9.36`/`0.9.37` images were published to ghcr.io.

**Fix:** #686 removes the stale `COPY` line (validated by building the builder stage through all COPY steps against the real context).

**Backfill note:** `docker.yml` triggers only on `v*` tag pushes, so merging the fix won't republish the missed tags. The next release tag will publish normally; if an image is needed sooner, `workflow_dispatch` from main refreshes `latest` only.

## Pages (transient, already resolved)

The "pages build and deployment" runs on main failed intermittently today (04:45, 05:49, 14:03 UTC) at GitHub's server-side deploy step with the generic "Deployment failed, try again later." The build job succeeded every time with artifacts byte-comparable to passing runs, and identical content deployed fine at 08:01, 08:19, and 11:33 — GitHub-side flakiness, not repo content. Re-running the failed deployment succeeded in 19s; the site is live at the current main commit (0cf4563). No code change needed.

AlexsJones:
## Summary
- The Docker image builds for `v0.9.36` and `v0.9.37` fail instantly with `"/data": not found` — `Dockerfile` still ran `COPY data/ ./data/`, but the root `data/` directory was removed in cf64a5b / fe89b82 (it held byte-identical duplicates of `llmfit-core/data/`, which is what the build actually embeds via `include_str!`).
- Removes the stale `COPY` line. `llmfit-core/data/` is already covered by `COPY llmfit-core/ ./llmfit-core/`.

## Testing
- Built the Dockerfile's builder stage up to the `cargo build` step against the real context — all `COPY` steps resolve.

Note: since `docker.yml` only triggers on tag pushes, the missing `0.9.36`/`0.9.37` images won't backfill from this merge — the next release tag will publish normally, or the workflow can be run via `workflow_dispatch` to refresh `latest`.

Closes #687

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,6 @@ COPY Cargo.toml Cargo.lock ./
 COPY llmfit-core/ ./llmfit-core/
 COPY llmfit-tui/ ./llmfit-tui/
 COPY llmfit-desktop/ ./llmfit-desktop/
-COPY data/ ./data/
 
 # Build release binary for llmfit-tui
 RUN cargo build --release -p llmfit


reneleonhardt:
Nice catch!

Release notes mention the same change for another commit:

* docker: drop COPY of removed root data/ directory ([5bb514d](https://github.com/AlexsJones/llmfit/commit/5bb514d130964311617269b0cffd4663a31522e3))

Issue #587, opened 2026-05-29
Detect OMLX as an MLX backend and fix MLX filter results


bruvv:
When running a mac, https://omlx.ai/ is pretty good solution. Friendly admin ui and inbuild downloader that supports MLX: 

<img width="412" height="71" alt="Image" src="https://github.com/user-attachments/assets/4437f374-0b2d-424f-a848-d14d2a76d99c" />

The thing is though llmfit does not show me MLX things it seems. I have set the following:

<img width="157" height="97" alt="Image" src="https://github.com/user-attachments/assets/9336f68e-9206-458e-8795-c494c86ae563" />

i pick the first one:

<img width="1716" height="632" alt="Image" src="https://github.com/user-attachments/assets/3a5141b9-9126-45da-8554-3273a8bb93c4" />

That gives me the following download: https://huggingface.co/mradermacher/Qwen3.5-35B-A3B-Claude-4.6-Opus-Reasoning-Distilled-GGUF (will be great if this is clickable btw)

<img width="1722" height="566" alt="Image" src="https://github.com/user-attachments/assets/35724e92-8ca7-4f95-b730-3417c0a81b59" /> 

But when searching this supposed MLX format i get:

<img width="1268" height="373" alt="Image" src="https://github.com/user-attachments/assets/67e5ffde-2346-4ac2-97f2-9e93a2828cf2" />

When I uncheck MLX i do see it:

<img width="1258" height="378" alt="Image" src="https://github.com/user-attachments/assets/03f65a87-3de4-47b7-887d-9d604beaa171" />

What is going on?

Ps please add omlx also as MLX detector in the top status bar of llmfit:

<img width="600" height="68" alt="Image" src="https://github.com/user-attachments/assets/c228d089-3523-4ba0-8194-dda20ed2e15e" />

AlexsJones:
Thanks @bruvv — splitting this into the two things it contains:

1. **OMLX detection** (feature request): treating [OMLX](https://omlx.ai/) as an installed MLX-capable runtime. Reasonable ask for Apple Silicon users.
2. **MLX filter showing no results** (possible bug): could you share your llmfit version (`llmfit --version`) and what the hardware summary reports for your Mac (chip / unified memory)? If MLX models are missing entirely with the MLX filter on, that may be the conservative KV-cache filtering tracked in #635 or a catalog gap rather than the filter itself.

Treating the OMLX integration as the feature here; the filter half is needs-info.

[...]

PR #701, opened 2026-07-08
fix(providers): detect oMLX as an MLX runtime

CN-Scars:
## Summary

Detect running oMLX as an MLX-capable runtime and prevent vLLM from claiming oMLX's OpenAI-compatible endpoint.

## Details

- Adds shared parsing for OpenAI-compatible `/v1/models` responses.
- Detects oMLX via `/api/status` and `/v1/models` entries with `owned_by: "omlx"`.
- Lets the MLX provider probe the default oMLX endpoint at `http://127.0.0.1:8000` when `MLX_LM_HOST` is not set.
- Makes vLLM detection ignore endpoints identified as oMLX.

This addresses the oMLX runtime detection part of #587. It intentionally does not change the MLX filter/catalog behavior that still needs reporter follow-up.

## Testing

- `cargo fmt --check`
- `cargo test -p llmfit-core omlx -- --nocapture`
- `cargo test -p llmfit-core vllm -- --nocapture`
- `cargo test -p llmfit-core`
- `cargo test -p llmfit`
- `cargo clippy -p llmfit-core --all-targets --all-features`

Manual verification with local oMLX 0.4.4 running at `127.0.0.1:8000`:

- Before: `MLX: (2 cached)`, `vLLM: ✓ (1 models)`
- After: `MLX: ✓ (2 installed)`, `vLLM: ✗`

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -396,36 +396,147 @@ impl ModelProvider for OllamaProvider {
 }
 
 // ---------------------------------------------------------------------------
+// OpenAI-compatible provider helpers
+// ---------------------------------------------------------------------------
+
+#[derive(serde::Deserialize)]
+struct OpenAiModelList {
+    data: Vec<OpenAiModel>,
+}
+
+#[derive(serde::Deserialize)]
+struct OpenAiModel {
+    /// Model id, e.g. "meta-llama/Llama-3.1-8B-Instruct".
+    id: String,
+    /// OpenAI-compatible providers may include an owner string. oMLX uses
+    /// `owned_by: "omlx"`, which lets us disambiguate it from vLLM.
+    owned_by: Option<String>,
+}
+
+fn openai_models_url(base_url: &str) -> String {
+    format!("{}/v1/models", base_url.trim_end_matches('/'))
+}
+
+fn fetch_openai_model_list(
+    base_url: &str,
+    timeout: std::time::Duration,
+) -> Option<OpenAiModelList> {
+    let resp = ureq::get(&openai_models_url(base_url))
+        .config()
+        .timeout_global(Some(timeout))
+        .build()
+        .call()
+        .ok()?;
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Solid work — correct diagnosis, right fix strategy (identity-based disambiguation via `owned_by` + `/api/status` shape, rather than port games), honest scoping, and the `OpenAiModelList` refactor deduplicates the vLLM types nicely. The guard rails are careful: the oMLX fallback probe requires identity proof so a regular vLLM on :8000 can't be misclaimed as MLX, and an explicit `MLX_LM_HOST` suppresses the fallback entirely. Manual before/after verification in the description is exactly what network-probe code needs given we have no HTTP mock infra.

Non-blocking suggestions:

1. **One redundant request in the healthy-oMLX path**: `fetch_candidate_models` hits `/api/status` *before* fetching the model list, but the list usually proves oMLX by itself via `owned_by`. Fetching the list first and only consulting `/api/status` when the list doesn't self-identify saves a round-trip:
   ```rust
   let list = fetch_openai_model_list(candidate.base_url, timeout)?;
   if candidate.require_omlx_identity
       && !openai_model_list_is_omlx(&list)
       && !endpoint_has_omlx_status(candidate.base_url, timeout)
   {
       return None;
   }
   ```
2. **Small behavior change worth noting in the changelog**: MLX `detect_with_installed` previously reported `available = true` whenever the HTTP call succeeded even if the JSON didn't parse; now an unparseable response falls through to `check_mlx_python()`. More correct, but observable.
3. **Latency bound grew slightly**: with no `MLX_LM_HOST`, `is_available` can make up to 3 requests at 2 s each against a black-holing host. Connection-refused fails fast so the normal case is unaffected — just noting the worst case.
4. Exact-match `owned_by == "omlx"` depends on an undocumented upstream string, but the `/api/status` fallback already covers drift, so no change needed.

Thanks @CN-Scars — merging-wise this is good to go from my side.

PR #728, opened 2026-07-10
fix: poll LM Studio's real download status endpoint (/api/v1/models/download/status/:job_id) and correct stale docs

### LINKED ISSUE #288: Integration Issue: Incompatible API for Querying Download Status Between llmfit and LM Studio
Image

Image

Image

mvanhorn:
## Summary
In `llmfit-core/src/providers.rs`, capture the `job_id` from the initial `LmStudioDownloadResponse` while reading the streaming POST body (drop the `#[allow(dead_code)]`), and add a `download_status_url(&self, job_id: &str)` helper alongside `models_url()`/`download_url()` that formats `{base}/api/v1/models/download/status/{job_id}`. In the "stream ended without completion" branch of `start_pull`, when a `job_id` was captured, poll that status endpoint (parsing the response with the existing tolerant `LmStudioDownloadStatus` struct: serde-defaulted `status`, `progress`, `downloaded_bytes`, `total_size_bytes`) and emit real `PullEvent::Progress` percentages, terminating on `completed`/`already_downloaded`/`failed`; keep the existing `GET /v1/models` installed-list polling as the fallback when no `job_id` was returned or the status endpoint errors (older LM Studio builds). Fix the stale docs: `docs/providers.md` line 118 and the equivalent lines in `README.zh.md` and `README.ja.md` change from polling `GET /api/v1/models/download-status` to the streaming-POST-plus-`GET /api/v1/models/download/status/:job_id` behavior.

## Why this matters
The reporter's screenshots show llmfit polling `GET /api/v1/models/download-status` against LM Studio 0.4.7, which logs "Unexpected endpoint or method" on every poll - LM Studio's REST API v1 has never exposed that path; its real status endpoint is `GET /api/v1/models/download/status/:job_id` (visible in the reporter's second screenshot of LM Studio's Supported Endpoints panel). The bad polling call itself was already removed from the client by PR #467 (merged 2026-04-19; the reporter's logs are dated 2026-03-23, i.e. an older llmfit build). But two halves of the incompatibility remain on main today: (1) `docs/providers.md` line 118, `README.zh.md`, and `README.ja.md` still document polling `GET /api/v1/models/download-status`, an endpoint that does not exist; and (2) the client never adopted LM Studio's actual per-job status endpoint - `LmStudioDownloadResponse.job_id` in `llmfit-core/src/providers.rs` is marked `#[allow(dead_code)]`, and when the POST stream closes early the fallback blindly polls `GET /v1/models` every 3s for up to 30 minutes with no progress percentage.

See https://github.com/AlexsJones/llmfit/issues/288.

## Testing
- Unit test in the existing `mod tests` of `llmfit-core/src/providers.rs`: `download_status_url` formats `http://127.0.0.1:1234/api/v1/models/download/status/abc123` from the default base URL, and respects a trailing-slash/custom `LMSTUDIO_HOST`-style base (mirroring the existing `models_url`/`download_url` construction). - Unit test: `LmStudioDownloadResponse` parsing captures `job_id` when present and tolerates its absence (older LM Studio), so the fallback path is selected when `job_id` is `None`. - Unit test: `LmStudioDownloadStatus` parsing maps `progress: 0.42` to 42.0 percent and `completed`/`failed` statuses to Done/Error terminal events (test the pure percent/terminal mapping helper if extracted, matching how existing lmstudio tests exercise pure helpers). - Edge case: status endpoint returns 404/error (LM Studio build without per-job status) - code falls back to the existing `GET /v1/models` installed-list polling instead of aborting the download tracking.

Fixes #288


AlexsJones:
Thanks for digging into #288, the core LM Studio fix is solid and the fallback test is exactly what I want to see. A couple of things block this as-is though.

**Blocking**

- `.osc-metadata/sync.json` is fork sync tooling metadata and can't land in the repo. Please drop it.
- Three unrelated changes are riding along with no mention in the PR body: the tauri `webviewInstallMode` fix, the unix socket handling plus synthetic `ConnectInfo` in `serve_api.rs`, and the `println!` to `eprintln!` swap in `main.rs`. Each of these might be fine on its own merits but they need to be seperate PRs, or at minimum separate commits with an explanation of why. The `eprintln!` one in particular changes a scripting contract (the PR URL was the one greppable stdout result of `bench --share --yes`) and I don't want that decided as a drive-by.

**Not blocking, but annoying**

- Worst case tracking time doubles: if status polling runs its full 30 minute budget and then falls back, we are at 60 minutes total. Share one poll budget across both phases.
- If the status endpoint returns 200 with an unexpected body, the all-defaults struct parses it as `status: ""` and the loop burns the entire budget before reporting "timed out", without ever falling back. A few consecutive empty statuses should trigger the fallback too.
- `status_url_builder` constructs a whole synthetic `LmStudioProvider` just to format a URL inside the thread. Clone the base url and use a free function, like the other helpers you added.
- `progress` is assumed to be on a 0..1 scale. That matches the existing stream parsing, but neither has been validated against a live LM Studio, and the endpoint path itself comes from the reporter's screenshot. Please run one real download against LM Studio >= 0.4.7 and paste the output before we merge, this integration has been written against an assumed API once already.

Happy to take the LM Studio + docs portion quickly once the scope is trimmed.

mvanhorn:
Thanks for the review - the blocking items are addressed in d287795. cargo test passes locally including the fallback tests.

AlexsJones (dismissed):
Thanks @mvanhorn — this is a model bug fix. ✅

The root-cause analysis is right: llmfit was polling `GET /api/v1/models/download-status`, which has never existed in LM Studio's REST API v1 (hence the "Unexpected endpoint or method" log spam in #288), and the real per-job endpoint is `GET /api/v1/models/download/status/:job_id`. The implementation is careful in all the right places:

- `job_id` is captured from the streaming POST body (finally earning its keep — the `#[allow(dead_code)]` removal is satisfying), with `lmstudio_response_job_id` filtering empty strings.
- The new `poll_lmstudio_download_status` emits real percentages via the existing tolerant `LmStudioDownloadStatus` parsing, terminates on `completed`/`already_downloaded`/`failed`, and **falls back instead of failing** on request/parse errors or repeated empty statuses — so older LM Studio builds without the per-job endpoint degrade to the previous installed-list polling.
- The shared `poll_budget` between the status poll and the installed-list fallback keeps the overall 30-minute cap intact — no doubled wait when the endpoint 404s on the first probe.
- Tests cover URL formatting (including trailing-slash bases), optional `job_id` parsing, percent/terminal mapping, the empty-status limit, and the error→fallback path asserting no spurious `PullEvent::Error`.
- Docs corrected in `docs/providers.md` **and** both translated READMEs — appreciated.

CI is green and this closes #288 on merge. Approving — recommending to @AlexsJones for final look and merge.

mvanhorn:
Thanks @AlexsJones! Glad the root-cause held up — let me know if you'd like anything else before merge.

AlexsJones:
Thanks for the ping, @mvanhorn — and for closing out the blocking items back in `d287795`. The build is green (7 checks passing) and the root-cause fix for #288 reads well.

There's one thing still gating merge, and it's from @AlexsJones's 2026-07-10 review rather than anything new: the request to **paste one real download run against LM Studio ≥ 0.4.7** before this lands. Two of the load-bearing facts in the change are still derived from the original reporter's screenshot rather than a live run:

- the `progress` value is assumed to be on a `0..1` scale, and
- the per-job endpoint path `GET /api/v1/models/download/status/:job_id` itself.

The fallback path means older builds degrade gracefully, so the risk is narrow — but since #288 was itself an assumed-API bug, a verbatim transcript from a live ≥ 0.4.7 download (endpoint hit + a couple of progress samples) is the one piece of evidence that would let @AlexsJones merge without re-deriving the scale by hand. If the scale turned out to be `0..100` the TUI would render e.g. `4200%`, which is exactly the failure mode worth ruling out here.

Drop that in and this is ready for his final look.

mvanhorn:
Ran it live against LM Studio 0.4.20 today. The per-job endpoint is real:

```
POST /api/v1/models/download {"model": "https://huggingface.co/bartowski/SmolLM2-135M-Instruct-GGUF"}
-> {"job_id": "job_3459c3cd91", "status": "downloading", "total_size_bytes": 144811360, ...}

GET /api/v1/models/download/status/job_3459c3cd91 polls:
[22:34:04] {"job_id": "job_3459c3cd91", "downloaded_bytes": 62144443, "total_size_bytes": 144811360, "bytes_per_second": 1512067.03, "estimated_completion": "...", "status": "downloading"}
[22:34:49] {"job_id": "job_3459c3cd91", "downloaded_bytes": 144811360, "total_size_bytes": 144811360, "status": "completed", "completed_at": "..."}
```

Two of the screenshot-derived assumptions did not survive contact with the real API, so I fixed both in b7fe566:

1. There is no progress field on 0.4.20 at all - the payload carries downloaded_bytes / total_size_bytes / bytes_per_second. The byte-ratio fallback in this PR is the live path; the 0..1 progress handling stays as a harmless first preference for builds that send it.
2. The API rejects direct .gguf links (400 "Invalid HuggingFace model URL format") and wants the repo URL form; and the POST response is one pretty-printed JSON object, not NDJSON, so the line-by-line reader was dropping the job_id. b7fe566 sends the repo URL (still verifying GGUF presence first) and parses the whole body before falling back to NDJSON/SSE lines.

End-to-end through llmfit serve after the fix (bartowski/Llama-3.2-1B-Instruct-GGUF):

```
[22:46:48] progress_pct: 3.15 ... [22:49:29] progress_pct: 99.75 ... [22:49:34] {"message": "completed", "progress_pct": 100.0, "status": "done"}
```

cargo test -p llmfit-core: 446 passed, 0 failed.

AlexsJones (approved):
Really appreciate you running this against a live LM Studio 0.4.20, that settles the API contract for good. The whole-body JSON parse with the per-line fallback covers both response shapes without dropping the `job_id`, and converting the verified resolve URL back to repo form keeps the GGUF pre-check while sending the form the API actually accepts (the 400 repro in your comment makes that airtight). New tests pin the conversion both ways, all checks green. Approving.

mvanhorn:
Thank you for the thorough verification pass @AlexsJones, and for confirming the API contract against a live LM Studio. Glad the download status polling is solid now.

mvanhorn:
Thanks for getting this in. Nice to have llmfit polling LM Studio's real download endpoint instead of guessing at progress.

PR #700, opened 2026-07-07
feat: add RamaLama as a detection provider

Olawoyin007:
Closes #699.

Adds RamaLama alongside the existing vLLM/Ollama/LM Studio/llama.cpp providers. RamaLama serves an OpenAI-compatible API (`GET /v1/models`, default port `8080`), so it mirrors `VllmProvider` almost exactly.

## What this does

- **`RamaLamaProvider`** (`providers.rs`) mirroring `VllmProvider`: `normalize_ramalama_host`, `Default` via `RAMALAMA_HOST` (default `http://localhost:8080`), `detect_with_installed` single-pass probe, and the `ModelProvider` impl. `start_pull` returns an informational error pointing at `ramalama serve <model>` (RamaLama, like vLLM, has no runtime pull endpoint).
- **Name mapping**: `hf_name_to_ramalama_candidates`, `is_model_installed_ramalama`, `has_ramalama_mapping`, `ramalama_pull_tag`.
- **Detection wiring** into `InstalledIndex` (`analysis.rs`): parallel detection thread, `is_installed`, and `installed_providers` (adds the `RamaLama` label to the detail panel).
- **TUI**: `ramalama_available` state + background detection in `tui_app.rs`, wired into the installed-first (`i`) and refresh (`r`) availability gates in `tui_events.rs`, plus a `RamaLama: ✓ (N models)` status-bar entry.
- **Unit tests** mirroring the `test_*_vllm` cases (candidates, install matching, mapping, pull tag, host normalization) - 8 new tests.

## Scope note

Implemented as a **detection** provider per the issue title. Because RamaLama has no runtime-download path (same as vLLM's erroring `start_pull`), it is intentionally **not** wired as a `DownloadProvider` (no download-manager / serve-API / MCP / bench plumbing, and it is not added to the `d` download gate). Happy to add the download-target path as a follow-up if you'd prefer the fuller vLLM mirror.

`RAMALAMA_HOST` follows the same convention as the provider's existing `VLLM_HOST` (which is likewise not surfaced in the README env-var table), so no doc table was touched.

## Testing

- `cargo test` - full workspace suite green (431 core + 46 + 8 + ... , incl. the 8 new `*_ramalama` tests).
- `cargo fmt --check` - clean.
- `cargo clippy --all-targets` (default members) - no new warnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #698, opened 2026-07-06
[Feature]: Recognize LMStudio AppImage distribution


patrolez:
### Problem or motivation

Hi, I'm currently running LM Studio as an AppImage, but llmfit isn't detecting it.

### Proposed solution

Is adding support for AppImage detection a difficult task?

### Feature area

Provider integration (Ollama, llama.cpp, MLX, Docker, LM Studio)

### Would you be willing to contribute this?

No, just suggesting

AlexsJones:
Is the server mode enabled? It's pretty simplistic in it's detection through ports

patrolez:
Ok, I can see that I am using custom server port, so no `1234`, but customized one and that is why. When I switched to default one "1234" so it was recognized. The question is if that setting could be also fetched/parsed, but if it is no a scope, so it can be closed.

PR #744, opened 2026-07-13
feat: detect RamaLama store models via `ramalama ls` when no server is running

Olawoyin007:
## What

Follow-up to #700. The RamaLama provider only detected models through a **running** server's `/v1/models` endpoint, so a user with models pulled into their RamaLama store saw nothing in llmfit unless a server happened to be up. This adds a store-based fallback.

`detect_with_installed` now falls back to `ramalama ls --json` when the HTTP probe fails, parsing the local store into the installed set through the same substring-matching path the server response already uses. This mirrors how Docker Model Runner is recognized as "installed but not running", and closes the same gap as the LM Studio follow-up (#698).

## Changes (all in `llmfit-core/src/providers.rs`)

- `RamaLamaStoreModel` (`name` + optional `shortname`; other `ls --json` fields ignored) and `parse_ramalama_store`, which extracts the full transport-qualified name (`huggingface://…`), its trailing component, and the shortname alias into the installed set.
- `installed_from_store` shells out to `ramalama ls --json`; returns `None` when the binary is absent or the command fails, so environments without RamaLama are unaffected.
- Extracted `insert_ramalama_name`, shared between the server and store paths so both normalize identically.
- The store fallback reports `available = true` with the store's model count. The status line reads `RamaLama: ✓ (N models)` (not "running"), so this is truthful for an installed-but-unserved setup.

## Scope

Detection only, matching #700. Download/serve gating (the `d` key / `start_pull`) is intentionally still out of scope.

## Testing

Verified locally with the exact CI gates:

- `cargo test` - full suite green, including 3 new parser tests (name/shortname extraction, HF-name matching via `is_model_installed_ramalama`, empty + invalid input)
- `cargo fmt --all -- --check` - clean
- `cargo clippy --all-targets --all-features` - no new warnings
- `llmfit --help` exits 0

`ramalama` is not installed in my environment, so `installed_from_store`'s subprocess path is covered by testing the pure `parse_ramalama_store` against the documented `ramalama ls --json` schema (`{shortname, name, modified, size}`) rather than a live CLI call.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -2666,35 +2666,49 @@ impl RamaLamaProvider {
     }
 
     /// Single-pass startup probe.
+    ///
+    /// Prefers the running server's `/v1/models`. When that is unreachable,
+    /// falls back to the local store via `ramalama ls --json`, so installed
+    /// models are still detected without a served endpoint (mirrors how Docker
+    /// Model Runner is recognized while "installed but not running").
     /// Returns `(available, installed_models, count)`.
     pub fn detect_with_installed(&self) -> (bool, HashSet<String>, usize) {
-        let mut set = HashSet::new();
         let Ok(resp) = ureq::get(&self.models_url())
             .config()
             .timeout_global(Some(std::time::Duration::from_millis(800)))
             .build()
             .call()
         else {
-            return (false, set, 0);
+            // Server not reachable — fall back to the on-disk store.
+            return match Self::installed_from_store() {
+                Some((set, count)) => (true, set, count),
+                None => (false, HashSet::new(), 0),
+            };
         };
 
         let Ok(list) = resp.into_body().read_json::<RamaLamaModelList>() else {
-            return (true, set, 0);
+            return (true, HashSet::new(), 0);
         };
-        let models = list.data;
-        let count = models.len();
-        for m in models {
-            let lower = m.id.to_lowercase();
-            set.insert(lower.clone());
-            // Also insert the model part after the publisher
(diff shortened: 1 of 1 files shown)


AlexsJones (dismissed):
Reviewed the full diff. This closes a real gap the same way the codebase already handles Docker Model Runner, and the claims in the PR body check out on inspection:

- **The fallback only fires when the HTTP probe fails**, and the server path's behavior is unchanged — the `insert_ramalama_name` extraction is a faithful refactor of the old inline loop (same set contents, full lowercased id + trailing path component).
- **The `available = true` semantics change is safe.** I traced every consumer of `ramalama_available` in the TUI: it gates the installed-first toggle (`i`) and refresh (`r`) — both meaningful for store-detected models — and RamaLama is correctly *not* in the download (`d`) gate, so nothing acts on a server that isn't running. The status-line wording ("✓ (N models)", no "running") is truthful for the installed-but-unserved case.
- **Tests cover the right things**: transport-qualified name extraction, shortname aliases, the end-to-end match through `is_model_installed_ramalama` (the same matcher the server path uses), and empty/invalid JSON.

One non-blocking observation: `installed_from_store` runs `ramalama ls --json` with no timeout, while the HTTP probe it replaces is bounded at 800 ms. Detection runs on a background thread so a slow store can't freeze the TUI — worst case is a late provider chip — but if you ever see reports of a hung detection thread, a `wait_timeout`-style bound here is the fix. Fine to land as is.

CI is green on all three platforms. Approving — recommending to @AlexsJones for final look and merge.

PR #750, opened 2026-07-14
ci(docker): build arm64 image natively instead of under QEMU

AlexsJones:
The Docker publish workflow compiled the arm64 image under QEMU on an x86 runner — an emulated Rust build that dominated the ~50 min wall clock and held back the amd64 image with it (single multi-arch push).

- Per-platform build jobs: `linux/amd64` on `ubuntu-latest`, `linux/arm64` natively on `ubuntu-24.04-arm` (free for public repos). No QEMU.
- Each job pushes by digest; a `merge` job assembles the multi-arch manifest with the same semver tags as before.
- GHA build cache scoped per platform.
- `workflow_dispatch` gains a required `tag` input so an existing release's image can be rebuilt on demand (used to rebuild v1.1.3 without waiting for the emulated run).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,12 @@ on:
     tags:
       - "v*"
       - "!v*-mac"
-  workflow_dispatch: # Allow manual trigger
+  workflow_dispatch:
+    inputs:
+      tag:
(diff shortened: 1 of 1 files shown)


reneleonhardt:
Good catch! How long is the arm64 build now?

Issue #570, opened 2026-05-21
Feature request: show size of unified memory available to GPU on macOS


oytech:
On macOS only part of unified memory is available to GPU and that can be configured via `iogpu.wired_limit_mb`:

```
sudo sysctl iogpu.wired_limit_mb=26000
iogpu.wired_limit_mb: 26000 -> 26000
```

For example `llama-server` shows it:
```
llama-server --list-devices
Available devices:
  MTL0: Apple M1 Max (26000 MiB, 25999 MiB free)
```

For comparison:
```
llmfit system

=== System Specifications ===
CPU: Apple M1 Max (10 cores)
Total RAM: 32.00 GB
Available RAM: 28.50 GB
Backend: Metal
GPU: Apple M1 Max (unified memory, 32.00 GB shared, Metal)
```

I think the size of unified memory available to GPU is a valuable piece of info that could be shown in CLI/TUI.

Loose idea - TUI simulate feature could prefill VRAM with the size of unified memory available to GPU.

AlexsJones:
Good catch, and thanks for the detailed comparison @oytech — this is a real reporting gap on Apple Silicon.

What's happening: llmfit currently reports the *total* unified pool (32 GB on your M1 Max), but macOS caps how much of that the GPU may wire. The default cap is a fraction of total RAM, and it's overridable via `sysctl iogpu.wired_limit_mb` exactly as you show. `llama-server` reports the effective cap because Metal exposes it directly as `MTLDevice.recommendedMaxWorkingSetSize` — that's the 26000 MiB figure in your output.

So the fix shape is clear: on macOS, query the effective GPU-available memory (Metal's `recommendedMaxWorkingSetSize`, with the `iogpu.wired_limit_mb` sysctl as a cross-check) and display it alongside the total, e.g. `GPU: Apple M1 Max (unified memory, 26.0 GB GPU-available of 32.0 GB shared, Metal)`. It matters for fit accuracy too, since model placement decisions should be made against the wired limit rather than total RAM.

Marking this as an enhancement — it's a well-scoped change in `llmfit-core`'s Apple GPU detection if anyone wants to pick it up.

🤝 _Handled by Alex's [Repo Steward](https://github.com/AlexsJones/repo-steward) — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)._

Learn more or run your own: https://github.com/AlexsJones/repo-steward

PR #754, opened 2026-07-15
feat(core): report macOS GPU-available unified memory

aramz33:
## Summary

On Apple Silicon, `llmfit system` reported the *total* unified-memory pool as GPU memory:

```
GPU: Apple M3 (unified memory, 16.00 GB shared, Metal)
```

But macOS caps how much of that pool the GPU may actually wire, so the number overstates what is usable for inference. This queries Metal's `recommendedMaxWorkingSetSize` and reports it alongside the total:

```
GPU: Apple M3 (unified memory, 11.84 GB GPU-available of 16.00 GB shared, Metal)
```

Addresses the display ask in #570 (placement/fit-accuracy tracked in a follow-up issue).

## How

- New macOS-gated dependency `objc2-metal` (feature `MTLDevice`) to read `MTLDevice.recommendedMaxWorkingSetSize`.
- Verified on an M3 (16 GB) that Metal's recommendation tracks a manual `iogpu.wired_limit_mb` override in both directions (setting 12000 / 14000 MiB made Metal report exactly those values), so Metal is the single source of truth — no sysctl read.
- New `gpu_available_gb: Option<f64>` on `SystemSpecs`, populated only for unified memory (Apple Silicon); `None` on other platforms.
- `--ram` / `--memory` overrides clear `gpu_available_gb` (the detected cap describes the real host, not the simulated one); the display falls back to the shared-pool-only line.
- Surfaced in the `system` CLI line, the TUI status bar, and the `/api/v1/system` JSON (plus the `API.md` example).

## Scope / notes

- Reporting only — fit scoring still budgets against the total unified pool. Re-pointing placement at the wired limit is left to the follow-up issue.

## Testing

- `cargo test`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets --all-features` are clean.
- Verified on an M3 16 GB: the default shows 11.84 GB GPU-available (74% of 16 GB, matching Metal's default cap); with `iogpu.wired_limit_mb=14000` the line reports 13.67 GB GPU-available.
- `llmfit --ram 128G system` and `llmfit --memory 24G system` fall back to the shared-pool-only line (no stale GPU-available figure).


AlexsJones (changes requested):
Thanks for this, @aramz33 — this is a genuinely well-put-together PR. The diagnosis matches what I laid out on #570 almost exactly: you went to `MTLDevice.recommendedMaxWorkingSetSize` rather than guessing at a percentage, you gated the dependency properly with `[target.'cfg(target_os = "macos")'.dependencies]`, you kept `gpu_available_gb` as `Option<f64>` so every other platform degrades to `None`, and you surfaced it in all three places (CLI, TUI status bar, `/api/v1/system`) plus `API.md`. Factoring `format_unified_memory_line` out as a pure function with the two-case unit test is exactly the shape I want in `hardware.rs`. The scope discipline in the description — reporting now, fit/placement as a deliberate follow-up — is the right call, and I appreciate you flagging it rather than quietly expanding.

A few things to sort before I merge.

**1. The `--ram` / `--memory` overrides leave `gpu_available_gb` stale (blocking)**

You list this under "known limitation", but I think it's worth fixing here rather than deferring, because it produces actively wrong output rather than merely incomplete output. `with_ram_override` (`llmfit-core/src/hardware.rs:2045`) updates `gpu_vram_gb`, `total_gpu_vram_gb` and the per-GPU `vram_gb` on unified-memory systems, but never touches `gpu_available_gb`. So on an M3 16 GB:

```
llmfit system --ram 128
GPU: Apple M3 (unified memory, 11.84 GB GPU-available of 128.00 GB shared, Metal)
```

The whole point of `--ram` is "model a machine I don't have", and that line now mixes a simulated pool with a detected cap from the real host. `with_gpu_memory_override` (`hardware.rs:2006`) has the same problem on both branches.

My suggestion: in both override paths, either set `self.gpu_available_gb = None` (honest — we don't know the cap for a hypothetical machine, and the formatter already falls back cleanly to the shared-pool-only line, which your `test_unified_line_falls_back_when_unknown` covers), or scale it by the same default ratio. I'd take `None`; it's simpler and it can't lie. There are already tests sitting right next to this — `test_ram_override_unified_memory_updates_gpu` (~`hardware.rs:3953`) — so please extend that pattern with a case asserting `gpu_available_gb` is cleared.

**2. Clamp the sysctl value (blocking)**

`sysctl_wired_limit_gb` accepts any non-zero value and hands it straight through, and `detect_gpu_available_gb` returns it unconditionally. `sudo sysctl iogpu.wired_limit_mb=999999` on a 16 GB Mac would print `976.56 GB GPU-available of 16.00 GB shared`. The kernel won't honour a limit above physical memory, so neither should we — please clamp the result to the shared pool (`min(value, shared_gb)`). Worth a test.

**3. Is the sysctl cross-check actually needed? (question)**

This is the one I'd most like your read on, since you have the hardware and I want to get it right rather than guess.

Your rationale is that Metal's recommendation "may not reflect a manually *raised* limit", so a non-zero sysctl wins. But the evidence in #570 points the other way: @oytech set `iogpu.wired_limit_mb=26000` on a 32 GB M1 Max and `llama-server --list-devices` then reported `26000 MiB`. llama.cpp gets that number from `recommendedMaxWorkingSetSize`, and 26000 is not a natural fraction of 32768 (the default ~75% would be ~24576). That strongly suggests Metal *does* already pick up the sysctl override, which would make the shell-out redundant — and it's a subprocess spawn on every `SystemSpecs::detect()`.

If it's redundant I'd rather drop it and rely on Metal alone: fewer moving parts, and no risk of the two sources disagreeing. If you've actually observed Metal failing to track a raised limit on your M3, say so and I'm happy to keep it — but please put that observation in the doc comment, because it's the entire justification for the code and future-me will otherwise delete it.

Related, whichever way you go: `detect_gpu_available_gb` bails to `None` via `MTLCreateSystemDefaultDevice()?` before it ever consults the sysctl, so if Metal device creation fails but the sysctl is readable, we throw away a value we had. Minor, and moot if the cross-check goes away.

**4. Nits**

- Wording/precision drift between surfaces: CLI prints `11.84 GB GPU-available of 16.00 GB shared` while the TUI (`llmfit-tui/src/tui_ui.rs:104`) prints `11.8 avail / 16.0 GB shared`. Tighter width in the status bar is fair, but let's at least align on one term — pick `avail` or `GPU-available` and use it in both.
- If the sysctl reader survives review, consider splitting the parse out (`parse_wired_limit_mb(&str) -> Option<f64>`) so the `0` / non-zero / garbage / negative cases are unit-testable without a subprocess — the same shape as `parse_macos_metal_gpus_from_system_profiler_json`. `available_ram_from_vm_stat` doesn't do this, so it's a soft preference, not a rule.

**On CI**

No checks have run on this yet — you're a first-time contributor here, so the workflows need my approval. I'll kick them off. Flagging it because it matters more than usual on this PR: `macos-latest` is the only matrix leg that will compile your `cfg(target_os = "macos")` code at all, so until it goes green neither of us has seen the new path build. I haven't run the branch locally.

**On #570**

I'm happy to call the display ask closed by this. The placement/fit-accuracy point from my comment on the issue and @oytech's "simulate could prefill VRAM" idea are both genuinely out of scope for a reporting PR — but please drop the `Closes #570` from the description and I'll open a dedicated follow-up issue for the placement work, so it doesn't get lost when this merges.

None of the above is structural — the design is right and the hard part is done. Fix 1 and 2, give me your read on 3, and I'll get this in.

AlexsJones:
Correction to my review above, @aramz33 — the **On CI** section is wrong, and in your favour.

I said no checks had run and that I'd kick them off. In fact they were approved and ran about an hour before I posted, and **all eight are green** — including `macos-latest`, which was the leg I flagged as the only one that compiles your `cfg(target_os = "macos")` code:

```
Detect changes                      SUCCESS
Test Suite (ubuntu-latest, stable)  SUCCESS
Test Suite (macos-latest, stable)   SUCCESS
Test Suite (windows-latest, stable) SUCCESS
Rustfmt                             SUCCESS
Clippy                              SUCCESS
Cargo Check                         SUCCESS
Python Checks                       SKIPPED
```

So please ignore that whole section: the new path builds and passes on macOS, and your `format_unified_memory_line` tests are green. That was the one unknown I'd called out as mattering more than usual here, and it's resolved.

My apologies — that paragraph was written against stale state and I posted it without re-checking. The four review points themselves stand unchanged: (1) and (2) are still what I'd want before merging, (3) is still a genuine question I'd like your read on, and (4) is nits.

aramz33:
Thanks for the thorough review, @AlexsJones — all four points addressed in c67a4eb.

**1. Overrides clearing `gpu_available_gb` — fixed.** Went with `None` as you suggested, in both `with_ram_override` and `with_gpu_memory_override` (both branches of the latter). The display falls back to the shared-pool-only line. Extended the tests per the existing pattern: the unified `--ram` test now seeds `gpu_available_gb: Some(27.0)` and asserts it's cleared, plus two new tests covering both `--memory` branches. Verified manually: `llmfit --ram 128G system` now prints `Apple M3 (unified memory, 128.00 GB shared, Metal)`.

**2 & 3. You were right — the sysctl cross-check was redundant, so it's gone (which also moots the clamp).** Rather than guess, I tested it on my M3 (16 GB): with the sysctl at 0, Metal's `recommendedMaxWorkingSetSize` reports 12,713,115,648 bytes (the ~74% default). After `sudo sysctl iogpu.wired_limit_mb=12000` it reports 12,582,912,000 — exactly 12000 MiB. After raising it to 14000 (above the 12124 MiB default, so specifically the *raised* case my doc comment worried about): 14,680,064,000, exactly 14000 MiB. Metal tracks the override in both directions, matching @oytech's llama.cpp observation in #570. `detect_gpu_available_gb` is now just the Metal read — no subprocess, no second source to disagree with — and the doc comment records the M3 verification so future-you knows why there's no sysctl handling. The kernel-clamping concern goes with it, since we now only ever report what Metal itself says.

**4. Terminology — aligned on `GPU-available`.** The TUI status bar now prints `Apple M3 (13.7 GB GPU-available / 16.0 GB shared, Metal)` to match the CLI.

Also removed `Closes #570` from the description and reworded it to point at your follow-up issue for the placement work — happy to take a look at that one too once it exists.

CI-wise: `cargo test` (542 tests), fmt and clippy all clean locally on macOS.

AlexsJones (changes requested):
Re-reviewed at c67a4eb. Points 1, 3 and 4 are done. Point 2 isn't — and your own answer to 3 is the reason it still matters.

**1. Overrides clear `gpu_available_gb` — addressed.** `with_gpu_memory_override` clears unconditionally after the if/else (`hardware.rs:2057`), so both branches are covered, and `with_ram_override` clears at `:2073`. Tests extended in the right place: the existing unified `--ram` test now seeds `Some(27.0)` and asserts it's cleared, plus the two new override cases. That's exactly what I asked for.

**3. Sysctl cross-check — addressed, and thanks for testing it rather than guessing.** Both directions with real numbers settles it, and the M3 verification is recorded in the doc comment (`hardware.rs:2182`) so future-me won't delete the reasoning. `detect_gpu_available_gb` being a bare Metal read with no subprocess is the right end state.

**2. Clamp — still open.** Your argument is that we now only ever report what Metal itself says. But the thing you established in 3 is precisely that Metal is a pass-through for `iogpu.wired_limit_mb`: you set 12000 and got 12000, set 14000 and got 14000. So "what Metal says" *is* "what the user put in the sysctl". Dropping the shell-out moved the untrusted value one layer down; it didn't remove it. `detect_gpu_available_gb` (`hardware.rs:2189`) returns `recommendedMaxWorkingSetSize` raw and `format_unified_memory_line` (`:2166`) prints whatever it's handed — nothing between them bounds it by the shared pool.

Both values you tested were below the 16384 MiB physical, so the case the clamp was for — a limit set *above* physical — is still untested and now unguarded. I don't want to ship on the assumption the kernel refuses it. `min(available, shared_gb)` in the formatter, where both numbers are already in scope, plus a test that an absurd input renders bounded to the pool. One line, and it can't be wrong.

If you've got the M3 to hand and `sudo sysctl iogpu.wired_limit_mb=999999` shows Metal clamping to something sane, say so and I'll drop the ask — but that observation then belongs in the doc comment next to the other one.

**4. Terminology — addressed.** TUI and CLI both say `GPU-available` now. Thanks for dropping `Closes #570` too; I'll get the follow-up issue open.

**Nit:** `with_ram_override` clears the field inside the `if self.unified_memory` guard (`:2073`) while `with_gpu_memory_override` clears unconditionally (`:2057`). Safe today, since the field is only ever populated for unified memory (`:127`), but that's an invariant living in two places. Lift it out of the guard.

**CI:** green at c67a4eb, checked just now — seven SUCCESS (Detect changes, Test Suite ubuntu/macos/windows stable, Rustfmt, Clippy, Cargo Check), Python Checks skipped by the path filter.

Clamp it and I'll merge.

aramz33:
Clamped in 7d188b5, plus the nit.

**2. Clamp — done.** `format_unified_memory_line` now bounds the figure with `available.min(shared_gb)` before rendering, and the doc comment states why: Metal passes the sysctl through verbatim, so the display has to bound it. Test added: `Some(976.56)` against a 16 GB pool renders `16.00 GB GPU-available of 16.00 GB shared`. I don't have the M3 at hand today to try `iogpu.wired_limit_mb=999999`, so I took the clamp rather than the observation — it holds either way.

**Nit — done.** The `gpu_available_gb = None` in `with_ram_override` now sits before the `unified_memory` guard, unconditional like `with_gpu_memory_override`. The invariant reads the same in both places.

Issue #621, opened 2026-06-09
[feature] Approximate usable context


MrMarble:
Hey! I have moderate hardware (3080 with 10Gb vram), I'm able to run some half decent models but the leftover memory only allows for a small context window.
I was playing with some models and it was a bit frustrating, llmfit told me it was a perfect fit, I was able to run it at usable token/s but the context went from the available 200k to a mere 8k, which is unusable for code...

I have made the changes locally, but want to ask here if this is something wanted before opening a pull request:

I added it directly in the Ctx column with  an arrow `262k→14k`, but I can change it to its own column (which could be better for sorting and filtering).

<img width="2747" height="978" alt="Image" src="https://github.com/user-attachments/assets/a3c3892f-0fec-4dc7-8abd-dfae6b677aa7" />

MrMarble:
Here's a zoomed in picture just in case that one can't be read properly 

<img width="158" height="743" alt="Image" src="https://github.com/user-attachments/assets/eb7c8381-4b70-4220-9915-9f25d3de79d3" />

AlexsJones:
Yes — this is very much wanted, please open the PR! Sorry for the slow reply.

Design preference: keep it in the **Ctx column with the arrow** as in your screenshot (`262k→14k`) rather than a new column — the table is already wide. Two asks:
- When cycling sort to Ctx (`s` key), sort by the **usable** value, since that's the number that actually constrains what you can do.
- Expose the computed usable context in the JSON output (`recommend`/`list`) and the `/api/v1/plan` response too, so the web UI and scripting users get it — even if that lands as a follow-up.

If the usable value drops below some floor (say 4k), it might also deserve a warning colour, since that's effectively unusable for most workflows. Looking forward to it 👍

PR #758, opened 2026-07-17
feat(api): expose usable_context and effective_context_length in the models envelope

aramz33:
## Problem

llmfit has two independent `fit_to_json` serializers. The one behind `llmfit fit --json` (`display.rs`) already emits `usable_context` and `effective_context_length`. The one behind the **REST API** (`/api/v1/models*`, via `serve_api.rs`) and the **MCP server** (`mcp_server.rs`) — `serve_shared::fit_to_json` — does not: it only exposes the model's native `context_length`.

That leaves the API/MCP consumers without the two numbers that matter most for acting on a recommendation:

- **`usable_context`** — the context that actually fits in this node's available memory alongside the weights (computed in core since #621). For an agent calling llmfit as an MCP tool, this is arguably the single most useful number in the envelope: it's the value you'd hand to `-c` / `--ctx-size`.
- **`effective_context_length`** — the context the row's `memory_required_gb` / `estimated_tps` were computed at (`DEFAULT_ESTIMATION_CTX = 8192`, or the `max_context` query param). Without it, an estimate served over the API isn't reproducible — the same gap the Estimate Basis section closed for `llmfit info` in #292.

## Change

- `serve_shared::fit_to_json`: add `usable_context` and `effective_context_length`. Purely additive — no existing key changes value or type. Both the REST endpoints and the MCP server pick it up.
- `API.md`: document the three context fields in the `/api/v1/models` envelope and what distinguishes them.
- Test: `fit_json_exposes_context_fields` analyzes a real catalog model with a >8192 window and asserts both fields serialize, that `effective_context_length` is the estimation cap, and that `usable_context <= context_length`.

## Verification

Run against the changed path (API server from this branch, macOS, 32GB, Radeon Pro 5500M):

```console
$ llmfit serve --port 8791 &
$ curl -s "localhost:8791/api/v1/models?limit=200" | jq '[.models[] | select(.usable_context < .context_length and .usable_context > .effective_context_length)][0] | {name, context_length, usable_context, effective_context_length}'
{
  "name": "second-state/DeepSeek-R1-Distill-Qwen-14B-GGUF",
  "context_length": 131072,
  "usable_context": 47768,
  "effective_context_length": 8192
}
```

All three numbers meaningfully differ on a real model: 131k advertised, ~47k actually fits, estimates computed at 8k.

Absence on `main` at the same seam:

```console
$ git show upstream/main:llmfit-tui/src/serve_shared.rs | grep -c usable_context
0
```

`make test` green (475 core + 54 tui). `make fmt` clean. `cargo clippy -- -D warnings` output is identical before and after this branch (the pre-existing errors on `main` come from rustc 1.97 being newer than CI's).

## Follow-up (separate PR, not this one)

`generate_llamacpp_command` currently emits `-c {effective_context_length}`, i.e. the 8k estimation cap rather than what fits. Driving `-c` from `usable_context` is the natural next step — happy to open that once this lands, if the direction looks right.

### DIFF
--- a/API.md
+++ b/API.md
@@ -130,7 +130,7 @@ Envelope shape:
       "parameter_count": "7B",
       "params_b": 7.0,
       "context_length": 32768,
-      "usable_context": 24576,
+      "usable_context": 32768,
       "effective_context_length": 8192,
       "use_case": "Coding",
       "category": "Coding",
--- a/API.md
+++ b/API.md
@@ -130,6 +130,8 @@ Envelope shape:
       "parameter_count": "7B",
       "params_b": 7.0,
       "context_length": 32768,
+      "usable_context": 24576,
+      "effective_context_length": 8192,
       "use_case": "Coding",
       "category": "Coding",
       "release_date": "2025-03-14",
@@ -158,6 +160,15 @@ Envelope shape:
   ]
 }
 ```
+
+The three context fields answer different questions:
+
+- `context_length` — the model's native window, as advertised upstream.
+- `usable_context` — how much of that window actually fits in this node's
+  available memory alongside the weights. Use this one to pick a runtime `-c`.
+- `effective_context_length` — the context the `memory_required_gb` and
+  `estimated_tps` figures on this row were computed at. Defaults to
+  `min(context_length, 8192)`; set by `max_context` when supplied.
 
 ---
 
--- a/llmfit-tui/src/serve_shared.rs
+++ b/llmfit-tui/src/serve_shared.rs
@@ -40,6 +40,8 @@ pub fn fit_to_json(fit: &ModelFit) -> serde_json::Value {
         "parameter_count": fit.model.parameter_count,
         "params_b": round2(fit.model.params_b()),
         "context_length": fit.model.context_length,
+        "usable_context": fit.usable_context,
+        "effective_context_length": fit.effective_context_length,
         "use_case": fit.model.use_case,
         "category": fit.use_case.label(),
         "release_date": fit.model.release_date,
@@ -151,5 +153,25 @@ mod tests {
         let gpu = &json["gpus"][0];
         assert!(gpu.get("memory_bandwidth_gbps").is_some());
         assert!(gpu["memory_bandwidth_gbps"].is_null());
+    }
+
+    #[test]
+    fn fit_json_exposes_context_fields() {
+        let db = llmfit_core::models::ModelDatabase::new();
+        let model = db
+            .get_all_models()
+            .iter()
+            .find(|m| m.context_length > llmfit_core::fit::DEFAULT_ESTIMATION_CTX)
+            .expect("catalog has a model with a large context window");
+        let fit = ModelFit::analyze(model, &specs_with_gpu("Tesla T4"));
+
+        let json = fit_to_json(&fit);
+
+        assert_eq!(json["usable_context"], fit.usable_context);
+        assert_eq!(
+            json["effective_context_length"],
+            llmfit_core::fit::DEFAULT_ESTIMATION_CTX
+        );
+        assert!(fit.usable_context <= model.context_length);
     }
 }


AlexsJones (changes requested):
Thanks for this — the change itself is correct, additive, and lands a gap that's genuinely worth closing. I checked the two load-bearing facts against the source rather than taking the description's word for them, and both hold:

- `API.md`'s *"Defaults to `min(context_length, 8192)`"* matches the implementation exactly — `fit.rs:322`, `None => model.context_length.min(DEFAULT_ESTIMATION_CTX)`.
- Both fields are plain `u32` on `ModelFit` (`fit.rs:253` and `fit.rs:258`), and `usable_context` is documented as capped at the model's native window, so the `usable_context <= context_length` assertion is an invariant and not a coincidence of the model the test happens to pick.

CI is green on all three platforms at `77a0c04`. No existing key changes value or type.

So I have no change to ask for in the diff. What I do want to fix before this lands is the **record**, because the problem statement is wrong in a way that will mislead the next person to touch this code.

## There are two `fit_to_json`s, not one

> `serve_shared::fit_to_json` is the single serialization seam behind both `llmfit fit --json` and the `/api/v1/models*` endpoints.

It isn't. There are two independent implementations:

| consumer | path | has the fields today? |
|---|---|---|
| `llmfit fit --json` | `main.rs:1487/1489` → `display::display_json_fits*` → **`display.rs:806`** | **yes, already on `main`** |
| `/api/v1/models*` | `serve_api.rs:353/381/413` → `serve_shared::fit_to_json` | no — this PR |
| MCP server | `mcp_server.rs:155/191` → `serve_shared::fit_to_json` | no — this PR |

`display.rs:813-814` already emits `effective_context_length` and `usable_context`, and `display.rs:1184` already tests it (`fit_json_includes_effective_context_length`). So the CLI half of the problem you describe doesn't exist — scripted consumers of `llmfit fit --json` have had both numbers for a while.

That doesn't make the PR less useful; it makes it a **different, narrower** PR than the one described. The real contribution is the `/api/v1/models*` **and MCP server** gap — and the MCP half isn't mentioned anywhere in the description, which is a shame, because for an agent calling `llmfit` as a tool, `usable_context` is arguably the single most useful number in the envelope. Please rewrite the problem statement around the seam you actually changed, and mention the MCP server — it's the strongest argument for this PR.

## The Verification block can't be what it says it is

```console
$ llmfit fit --json | jq '.models[] | select(.context_length > 8192)' | head
Qwen/Qwen2.5-Coder-3B-Instruct                 native=32768  est=8192  usable=32768  Perfect
```

Two problems. `jq '.models[] | select(...)'` emits JSON objects — it will not print an aligned `native=…  est=…  usable=…` table. And `llmfit fit --json` runs through `display.rs`, which **this diff does not touch**, so that command prints those fields on `main` too and demonstrates nothing about the change.

If you hand-formatted a real run for readability, no drama — just say so. But this repo has been bitten specifically by API claims derived from assumed rather than observed behaviour ([#288](https://github.com/AlexsJones/llmfit/issues/288)), so a pasted transcript needs to be a real one. Please replace it with a run against the path you changed:

```console
$ llmfit serve &
$ curl -s localhost:8080/api/v1/models | jq '.models[0] | {context_length, usable_context, effective_context_length}'
```

That's the evidence I can't produce for you — I review the diff, I don't run contributor branches on this machine.

## Worth its own issue (not this PR)

The two envelopes have drifted well past these two fields, and one of the drifts is a live trap:

- **`fit_level` and `run_mode` mean different things under the same key.** `display.rs:817` emits `fit.fit_text()` → `"Perfect"`, `"Too Tight"`. `serve_shared.rs` emits `fit_level_code(...)` → `"perfect"`, `"too_tight"`, and puts the human string under a separate `fit_label`. A client that moves from `--json` to the API gets the same key with a different vocabulary, silently.
- `serve_shared` is missing `disk_size_gb`, `installed`, `capability_ids`, `ollama_name`, `estimate_basis`, `verify_command`, `measured_tps`.
- `display.rs` is missing `fit_label`, `run_mode_label`, `supports_tp`.

Two hand-maintained copies of one envelope will keep diverging exactly like this — your PR is the third field-by-field patch of the gap. A shared serializer with the presentation differences layered on top would end the category. Happy to see that as a follow-up issue if you want to open one.

## On your proposed follow-up

Driving `-c` from `usable_context` rather than the 8k estimation cap is the right instinct and correctly scoped as a separate PR — that one changes what users actually run, so it deserves its own review.

**Verdict: iterate** — no code changes requested. Fix the description (the seam, the MCP server, the console block) and this is ready as far as I'm concerned; the final call is @AlexsJones's.

---

aramz33:
Both catches are correct, and I've rewritten the description around them. Taking them in order:

**The seam claim was wrong.** I said `serve_shared::fit_to_json` was the single serializer behind both the CLI and the API without checking `display.rs`, which has its own private `fit_to_json` that's been emitting both fields for a while (with a test). What this PR actually changes is the REST API and MCP server envelope — the description and title now say exactly that, and the MCP angle is front and center, since you're right that it's the strongest argument here.

**The console block was not a real transcript.** I ran a real command earlier in the session, but through `llmfit fit --json` — the path this diff doesn't touch — and then hand-formatted the output into that aligned table for readability. Neither half of that is acceptable in a Verification section, and I should have caught that it proved nothing about the change. The block now in the description is a verbatim run against `llmfit serve` from this branch, plus a `git show upstream/main | grep -c` showing the field count at the same seam on `main` is zero. Point taken on #288 — evidence for an API claim comes from the API.

On the envelope drift: opening a follow-up issue for it now, with the divergence re-verified against the source rather than copied from your table. The `fit_level`/`run_mode` vocabulary split under identical keys is indeed the live trap — I hit the field lists while writing the original (wrong) description and can confirm the drift is broader than the two fields this PR adds.

AlexsJones (approved):
Delta re-review after the description rewrite — closing the loop on the previous review's three asks:

- **The seam is now stated correctly**: the description names `serve_shared::fit_to_json` behind the REST API *and* the MCP server, with the MCP angle front and center — which is indeed the strongest argument for the change.
- **The Verification block is now real evidence**: a verbatim run against `llmfit serve` from this branch, plus the field-count grep at the same seam on `main`. That's exactly what the hand-formatted `--json` block couldn't demonstrate.
- **The envelope drift is filed as #759**, re-verified against source rather than copied over — and it's confirmed there as a real bug.

No code has changed since the previous review (head still `77a0c04`, CI green on all three platforms), and the diff itself was already verified clean. Switching my review to approve — recommending to @AlexsJones for final look and merge.

PR #761, opened 2026-07-18
feat(api): unify fit envelope serializers, exposing parity fields on REST/MCP

### LINKED ISSUE #759: Two fit envelope serializers have drifted: same keys carry different vocabularies across CLI vs API/MCP
### Summary

There are two hand-maintained `fit_to_json` serializers for the same model-fit envelope, and they have drifted — in fields, and more dangerously, in the meaning of identical keys:

| consumer | serializer |
|---|---|
| `llmfit fit --json` | `display.rs` (private `fit_to_json`) |
| `/api/v1/models*` REST endpoints | `serve_shared::fit_to_json` (via `serve_api.rs`) |
| MCP server | `serve_shared::fit_to_json` (via `mcp_server.rs`) |

Spotted during review of #758, which is itself the third field-by-field patch of this gap. Everything below re-verified against current `main` + #758.

### Same key, different vocabulary (the live trap)

A client that moves between `llmfit fit --json` and the API gets the same keys with silently different value sets:

| key | CLI (`display.rs`) | API/MCP (`serve_shared.rs`) |
|---|---|---|
| `fit_level` | `fit_text()` → `"Perfect"`, `"Too Tight"` | `fit_level_code()` → `"perfect"`, `"too_tight"` |
| `run_mode` | `run_mode_text()` → `"GPU"`, `"CPU+GPU"`, `"MoE"`, `"TP"` | `run_mode_code()` → `"gpu"`, `"cpu_offload"`, `"moe_offload"`, `"tensor_parallel"` |
| `runtime` | `runtime_text()` = `label()` → `"llama.cpp"`, `"MLX"` | `runtime_code()` → `"llamacpp"`, `"mlx"` |

The API/MCP side carries the human strings under separate `fit_label` / `run_mode_label` keys; the CLI overloads the machine key with the human string.

### Field drift

- Missing from API/MCP (`serve_shared`): `disk_size_gb`, `installed`, `capability_ids`, `ollama_name`, `estimate_basis`, `verify_command`, `measured_tps`
- Missing from CLI (`display.rs`): `fit_label`, `run_mode_label`, `supports_tp`

`estimate_basis` missing from the API is worth flagging on its own: the reproducibility data added for #292 never reaches API/MCP consumers.

### Proposal

One shared base serializer (natural home: `serve_shared`, which already exists for this purpose) emitting the superset with stable machine codes + `*_label` keys for human strings, with each frontend layering its presentation differences explicitly on top. Two constraints worth honoring:

1. **No silent value changes under existing keys.** The CLI's `fit_level: "Perfect"` vocabulary is load-bearing for existing scripts; unification likely means the CLI adopting the `code + label` pair shape additively before any deprecation of the overloaded values.
2. Keep the per-consumer extras (e.g. the CLI's `llamacpp_command` injection) as explicit overlays, not forks.

Happy to take this on once #758 lands, if the direction looks right.

aramz33:
> **Stacked on #758.** This branches off `feat/usable-context-json`, so until #758 merges the diff here shows its three commits too. I'll rebase onto `main` — down to the single `feat(api):` commit — once #758 lands; the GitHub diff collapses on its own at that point.

Closes #759.

## What this does

Collapses the two hand-maintained `fit_to_json` serializers into one. `serve_shared::fit_to_json` becomes the single source of truth and emits the **superset**; `display::fit_to_json` (the `fit --json` path) is now a thin overlay over it.

Concretely, **REST and MCP gain the seven fields that were CLI-only**: `installed`, `disk_size_gb`, `capability_ids`, `ollama_name`, `estimate_basis` (the #292 reproducibility data, which never reached API/MCP before), `verify_command`, and `measured_tps`. This is purely additive — no existing API key changes value.

`generate_llamabench_command` moves into `serve_shared` alongside its only remaining consumers.

## What it deliberately does *not* do

It does not flip any existing key's value. The CLI's `fit_level: "Perfect"` vocabulary is load-bearing for existing scripts, so the CLI overlay keeps `fit_level`/`run_mode`/`runtime`/`capabilities` as their legacy human strings. The API keeps its machine codes.

So the same-key/different-vocabulary split from #759 **still exists by design after this PR** — but both sides now carry the `*_label` keys, so a *later* PR can deprecate the CLI overload additively. This PR ends the **field drift** and the two-functions-forking problem; the vocabulary unification is the follow-up that deprecation enables.

(One transient redundancy falls out of that rule: on the API, `capabilities` and `capability_ids` are equal for now — the same additive-before-deprecation window as the CLI carrying both `fit_level:"Perfect"` and `fit_label:"Perfect"`. It collapses when the CLI overload is deprecated.)

## Verification

Backward-compat is now pinned by a test — the `"Perfect"` vs `"perfect"` invariant was previously guarded by nothing.

**CLI `fit --json` (unchanged where it must be; additive gains):**
```
$ llmfit --json fit --limit 1 | jq '.models[0] | {fit_level, fit_label, run_mode, runtime, supports_tp, verify_command}'
{
  "fit_level": "Perfect",       # legacy human string, unchanged
  "fit_label": "Perfect",       # new, additive
  "run_mode": "GPU",
  "runtime": "MLX",
  "supports_tp": [1, 2, 4, 8],  # new, additive
  "verify_command": null        # MLX runtime → no llama-bench command
}
```

**REST `/api/v1/models` (drift closed — verbatim from `llmfit serve` on this branch, `force_runtime=llamacpp`):**
```
$ curl -s "127.0.0.1:8799/api/v1/models?limit=1&force_runtime=llamacpp" \
    | jq '.models[0] | {fit_level, fit_label, installed, disk_size_gb, capability_ids, verify_command, estimate_basis: (.estimate_basis|type), measured_tps}'
{
  "fit_level": "perfect",       # machine code, unchanged
  "fit_label": "Perfect",
  "installed": false,           # ┐
  "disk_size_gb": 8.6,          # │ all seven previously
  "capability_ids": ["tool_use"], # │ absent from the API
  "verify_command": "llama-bench -m <path-to-Q8_0-gguf> -ngl 99 -p 512 -n 128", # │
  "estimate_basis": "object",   # │ (#292 data now reaches API/MCP)
  "measured_tps": null          # ┘
}
```

Before/after at the seam:
```
$ git show origin/main:llmfit-tui/src/serve_shared.rs | grep -c estimate_basis   # → 0
```

`cargo test -p llmfit` green (56 unit + 8 integration), including the new backward-compat guard and drift-closed tests; `cargo fmt --check` and `cargo clippy` clean on the touched files. `API.md` updated to document the new envelope fields and the CLI-vs-API vocabulary note.

### DIFF
--- a/API.md
+++ b/API.md
@@ -130,7 +130,7 @@ Envelope shape:
       "parameter_count": "7B",
       "params_b": 7.0,
       "context_length": 32768,
-      "usable_context": 24576,
+      "usable_context": 32768,
       "effective_context_length": 8192,
       "use_case": "Coding",
       "category": "Coding",
--- a/API.md
+++ b/API.md
@@ -155,7 +155,24 @@ Envelope shape:
       "memory_available_gb": 12.0,
       "utilization_pct": 48.3,
       "notes": [],
-      "gguf_sources": []
+      "gguf_sources": [],
+      "capabilities": ["tool_use"],
+      "capability_ids": ["tool_use"],
+      "license": "apache-2.0",
+      "supports_tp": [1, 2, 4],
+      "installed": false,
+      "disk_size_gb": 5.1,
+      "ollama_name": "qwen2.5-coder:7b-instruct",
+      "estimate_basis": {
+        "method": "roofline",
+        "gpu_bandwidth_gbps": 320.0,
+        "ddr_bandwidth_gbps": null,
+        "local_calibration": null,
+        "efficiency": 0.85,
+        "assumed_context": 8192
+      },
+      "verify_command": "llama-bench -m <path-to-Q5_K_M-gguf> -ngl 99 -p 512 -n 128",
+      "measured_tps": null
     }
   ]
 }
@@ -169,6 +186,27 @@ The three context fields answer different questions:
 - `effective_context_length` — the context the `memory_required_gb` and
   `estimated_tps` figures on this row were computed at. Defaults to
   `min(context_length, 8192)`; set by `max_context` when supplied.
+
+The envelope also carries these fields, now at parity with `llmfit fit --json`
+(both frontends serialize through one shared function):
+
+- `installed` — whether the model was found in a local runtime provider.
+- `disk_size_gb` — estimated on-disk size at `best_quant`.
+- `capability_ids` — machine-readable capability ids (snake_case); mirrors
+  `capabilities` here. Note `llmfit fit --json` overloads its `capabilities`
+  key with human labels (e.g. `"Tool Use"`) — that overload is CLI-only and
+  slated for deprecation.
+- `ollama_name` — the `ollama pull` tag for this model, when derivable.
+- `estimate_basis` — how `memory_required_gb`/`estimated_tps` were derived
+  (bandwidths, efficiency, assumed context), for reproducibility.
+- `verify_command` — a `llama-bench` invocation measuring the same throughput
+  this row estimates (llama.cpp GPU / CPU-only runs; `null` otherwise).
+- `measured_tps` — a recorded benchmark result if one exists, else `null`.
+
+Note on vocabulary: `fit_level`, `run_mode`, and `runtime` here are stable
+machine codes (e.g. `"good"`, `"gpu"`, `"llamacpp"`), with the human string
+under the paired `*_label` key. `llmfit fit --json` emits the human string
+directly under those same keys — a CLI-only legacy overload.
 
 ---
 
--- a/API.md
+++ b/API.md
@@ -130,6 +130,8 @@ Envelope shape:
       "parameter_count": "7B",
       "params_b": 7.0,
       "context_length": 32768,
+      "usable_context": 24576,
+      "effective_context_length": 8192,
       "use_case": "Coding",
       "category": "Coding",
       "release_date": "2025-03-14",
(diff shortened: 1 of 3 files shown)


AlexsJones (approved):
Reviewed the full diff (post-#758 collapse: `API.md`, `display.rs`, `serve_shared.rs`). This lands the right half of #759 cleanly:

- `serve_shared::fit_to_json` becomes the single serializer, and REST/MCP gain the seven formerly CLI-only fields (`installed`, `disk_size_gb`, `capability_ids`, `ollama_name`, `estimate_basis`, `verify_command`, `measured_tps`) — purely additive, and the #292 reproducibility data finally reaches API/MCP consumers.
- The CLI overlay is exactly four keys (`fit_level`, `run_mode`, `runtime`, `capabilities`), each preserving the legacy human vocabulary existing `fit --json` scripts parse. The new `cli_json_keeps_legacy_vocab_while_shared_uses_codes` test pins the `"Perfect"` vs `"perfect"` invariant that previously nothing guarded — that was the scary regression surface in #759, and it's now tested on both sides of the seam.
- `generate_llamabench_command` moving to `serve_shared` (with its doc comment intact, visibility tightened to `pub(crate)`) removes the last duplicated logic between the frontends.
- `API.md` documents the three context fields and is honest that the remaining same-key/different-vocabulary split is a deliberate deprecation runway, not an accident.

Because no existing key's value flips, the canonical-vocabulary question raised on #759 stays open for @AlexsJones — but this PR doesn't foreclose either answer; it ends the field drift and the two-serializer fork while making the eventual CLI deprecation additive. That sequencing is right.

CI green on all four platforms + rustfmt + clippy, 56 unit + 8 integration tests passing. Approving — recommending to @AlexsJones for final look and merge. (The GitHub diff has already collapsed to the single-change view now that #758 is in `main`; squash-merge needs no rebase from you.)

PR #779, opened 2026-07-21
feat(tui): [SIMULATED] GPU indicator and / search in the community leaderboard

AlexsJones:
## What

Two UX improvements to the community leaderboard (benchmark) view:

**`[SIMULATED]` indicator when browsing another card's board.** Selecting a hardware preset via `H` now temporarily replaces the GPU string in the top system bar with a warning-colored bold `GPU: [SIMULATED] RTX 5080 (16 GB)`, and tags the leaderboard summary line with `[SIMULATED]`. The simulation resets when:
- the card is deselected via "My Hardware (auto-detect)",
- the leaderboard view is closed back to the main TUI (also drops the cached preset rows so the next open refetches for the detected hardware), or
- the app restarts (the selection was already session-only).

**`/` search over leaderboard rows.** Pressing `/` in the leaderboard types into the top Search box (same widget and semantics as the main table search: case-insensitive, space-separated AND terms) and filters rows live across model, engine, quant, and submitting user, with an `N/M match` count in the summary line. Enter applies the filter, Esc clears it; includes a no-match state and updated status-bar/help hints.

## Testing

- `cargo test -p llmfit`: 63 unit + 8 smoke tests pass (3 new tests: search filtering/AND terms, cursor bounds + clear behavior, close-resets-simulation)
- clippy clean for the touched files
- Verified live in a tmux-driven TUI session: badge on select/replace, live filtering (20→4 rows), no-match state, Esc clear, badge removal on deselect and on view close, refetch for detected hardware on reopen

## Follow-up in this PR

**Active range filters spelled out.** Advanced range filters from the `F` popup (params min/max, mem% min/max) persist across sessions and silently constrain every fit category; previously the only indicator was a cryptic `R`/`M` letter in the Fit box. The Fit box now shows the actual range (e.g. `Too Tight ≤2B`, `mem≥50%`) in warning color, and the empty-results hint names the active range explicitly.

### DIFF
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -1035,6 +1035,10 @@ pub struct App {
     /// Index into HardwarePreset::all() when the picker is open.
     pub bench_hw_picker_cursor: usize,
     pub bench_hw_picker_open: bool,
+    /// `/` filter over the leaderboard rows (model, engine, quant, user).
+    pub bench_search_query: String,
+    /// True while the `/` search box is capturing keystrokes.
+    pub bench_search_active: bool,
 
     // Live inference-bench view (llmfit bench — distinct from community benchmarks above)
     pub show_bench: bool,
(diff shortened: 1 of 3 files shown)


AlexsJones:
Added a third commit addressing a related visibility gap found while testing: a params/mem range saved from the `F` popup silently constrains every fit category, with only a cryptic single-letter `R`/`M` marker as evidence (this made fit categories appear inexplicably empty). The Fit box now spells out the active range — e.g. `Too Tight ≤2B` — in warning color, and the empty-state hint names it: "No models match current filters — active range filter: ≤2B. Press F to adjust".

Issue #732, opened 2026-07-10
[Bug]:Display corruption on Windows


benbaker76:
<img width="1366" height="1046" alt="Image" src="https://github.com/user-attachments/assets/31df1736-dad7-4f34-bd63-78a3d16d5f04" />

### Bug description

When I first run llmfit on Windows I get some corrupted row text. When I move the selector over them it will then correctly draw them.

### Expected behavior

All rows should render correctly.

### Actual behavior

It renders two rows incorrectly.

### Steps to reproduce

1. Run llmfit

### Diagnostic report (`llmfit doctor`)

```text
PS C:\Users\ben> llmfit doctor
# llmfit doctor report

Paste this whole report into a GitHub issue at https://github.com/AlexsJones/llmfit/issues — the raw tool output below is what lets detection bugs become regression tests. It contains hardware model names and driver info only.

- llmfit version: 1.0.1
- OS: windows (x86_64)

## Detected by llmfit

SystemSpecs {
    total_ram_gb: 63.69118881225586,
    available_ram_gb: 33.4454345703125,
    total_cpu_cores: 32,
    cpu_name: "Intel(R) Core(TM) i9-14900K",
    has_gpu: true,
    gpu_vram_gb: Some(
        23.98828125,
    ),
    total_gpu_vram_gb: Some(
        23.98828125,
    ),
    gpu_name: Some(
        "NVIDIA GeForce RTX 4090",
    ),
    gpu_count: 1,
    unified_memory: false,
    backend: Cuda,
    gpus: [
        GpuInfo {
            name: "NVIDIA GeForce RTX 4090",
            vram_gb: Some(
                23.98828125,
            ),
            backend: Cuda,
            count: 1,
            unified_memory: false,
        },
    ],
    cluster_mode: false,
    cluster_node_count: 0,
}


## nvidia-smi (extended query)

[N/A], 24564, NVIDIA GeForce RTX 4090


## nvidia-smi (standard query)

24564, NVIDIA GeForce RTX 4090


## rocm-smi --showmeminfo vram

(not available: program not found)


## rocm-smi --showproductname

(not available: program not found)


## PowerShell Win32_VideoController

Microsoft Remote Display Adapter|
NVIDIA GeForce RTX 4090|4293918720


## vulkaninfo --summary

==========
VULKANINFO
==========

Vulkan Instance Version: 1.4.321


Instance Extensions: count = 20
-------------------------------
VK_EXT_debug_report                    : extension revision 10
[...]

PR #786, opened 2026-07-21
Move terminal clear before main loop

benbaker76:
Current solution does not work. Tested llmfit v1.1.4 in Windows.

In reference to https://github.com/AlexsJones/llmfit/issues/732

### DIFF
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -1293,11 +1293,11 @@ fn run_tui_inner(
         app.open_bench();
     }
 
-    // Force a full clear before the first frame: ratatui only paints the cells
-    // a frame renders and trusts EnterAlternateScreen to start from a blank
-    // screen, but some terminals (notably Windows Terminal, #732) switch to the
-    // alternate screen without clearing it, leaving the previous shell content
-    // visible underneath sparse frames.
+    // Force a full clear AFTER the first drawn frame (the boot screen), not
+    // before it: on some terminals (notably Windows Terminal, #732) a clear()
+    // issued before any draw is a no-op, so clearing here — after the backend
+    // has drawn once — is what actually removes the previous shell content
+    // that EnterAlternateScreen leaves visible under sparse frames.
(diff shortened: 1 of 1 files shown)


AlexsJones (changes requested):
Thanks @benbaker76 — an empirical test on real Windows Terminal is exactly what this bug needed, and the mechanism checks out: the v1.1.4 fix called `clear()` immediately after `Terminal::new()`, before any draw, and on Windows Terminal that early clear is effectively a no-op, so #732's residual shell content survived it. Moving the clear to after the boot frame (once the backend has actually drawn) makes it take effect. CI is green and the change is minimal — good.

One change before this merges, because the relocated comment is now describing the *old* code:

> `// Force a full clear before the first frame: ...`

The clear no longer runs before the first frame — the boot screen is the first frame now. As written, the comment invites a future refactor to "restore" the clear to the top, which would silently reintroduce this bug. Could you replace it with something like:

```rust
// Force a full clear AFTER the first drawn frame (the boot screen), not
// before it: on some terminals (notably Windows Terminal, #732) a clear()
// issued before any draw is a no-op, so clearing here — after the backend
// has drawn once — is what actually removes the previous shell content
// that EnterAlternateScreen leaves visible under sparse frames.
```

Two observations for the record, neither blocking:
- On Windows Terminal the boot screen itself ("Detecting system hardware...") will still briefly render over residual shell content, since nothing effective clears before it. Given provider detection runs in background threads and the main loop clears right after, that flash should be short — acceptable.
- If your testing confirms v1.1.4 still shows the bug, #732 was closed by a fix that didn't hold on real Windows; worth the maintainer reopening or linking it here so the history is honest.

Happy to approve as soon as the comment is updated.

AlexsJones (approved):
Comment now tells the true story, thanks for the quick turnaround. The clear runs after the boot frame and the comment explains exactly why it has to sit there, so nobody will "restore" it to the top in a future refactor. Your Windows Terminal test already proved the mechanism and CI is green. Good to go.

PR #792, opened 2026-07-23
fix(hardware): detect Jetson/Tegra Orin iGPU as unified memory

lucamaf:
## Summary
- Jetson Orin's `nvidia-smi` reports `addressing_mode` as `N/A` rather than `ATS` (the legacy nvgpu/gk20a driver stack used by Orin/Xavier; `ATS` is only populated on newer Grace/Thor chips), so the existing unified-memory detection missed it and treated the on-chip GPU as a discrete CUDA device with unknown VRAM instead of unified memory shared with system RAM.
- `nvidia-smi` always suffixes Tegra/Jetson iGPU names with `(nvgpu)` (e.g. `"Orin (nvgpu)"`), so `is_nvidia_unified_memory_gpu()` now also matches on that substring, independent of `addressing_mode`.
- Added two regression tests alongside the existing GB10/ATS unified-memory tests.

Verified live on a Jetson Orin NX: before the fix, `llmfit doctor` reported `has_gpu: true, unified_memory: false, gpu_vram_gb: None` (breaking model-fit sizing); after the fix it reports `unified_memory: true, gpu_vram_gb: Some(7.1)` matching total system RAM, and `llmfit recommend` produces correctly-sized recommendations.

## Test plan
- [x] `cargo test -p llmfit-core --lib hardware::` — 85/85 pass, including new Jetson tests
- [x] `cargo test -p llmfit-core --lib` — full suite, 488/488 pass
- [x] Built release binary and ran `llmfit doctor` / `llmfit recommend` live on a Jetson Orin NX Developer Kit

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -2858,6 +2858,15 @@ fn is_nvidia_unified_memory_gpu(name: &str) -> bool {
     if lower.contains("2e12") {
         return true;
     }
+    // Jetson / older Tegra SoCs (Orin, Xavier, Nano, ...) run the legacy
+    // nvgpu/gk20a driver stack, which reports `addressing_mode` as "N/A"
+    // rather than "ATS" (that field is only populated on newer Grace/Thor
+    // chips). nvidia-smi always suffixes these iGPUs' name with "(nvgpu)",
+    // so use that as a reliable unified-memory signal independent of
+    // addressing_mode.
+    if lower.contains("nvgpu") {
+        return true;
+    }
     false
 }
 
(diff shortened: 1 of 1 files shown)


AlexsJones:
Please cargo fmt @lucamaf

AlexsJones (approved):
Nice one, verifying this live on an Orin NX before and after makes it an easy review. The `(nvgpu)` suffix is the right signal here: the legacy nvgpu/gk20a stack never populates `addressing_mode`, so name matching is the only reliable hook, and it keeps Jetson on the same unified-memory path as the GB-series chips. Tests cover both the parser and the classifier, including the discrete-GPU negative case. LGTM.

PR #793, opened 2026-07-24
fix: expose MiniMax M3 vision capability

octo-patch:
Reason: MiniMax-M3 image input is not exposed through the repository's vision capability pipeline.

Changes:
- Mark the MiniMax-M3 catalog entry as multimodal vision metadata in the scraper source and embedded catalog.
- Add a regression test that verifies the embedded MiniMax-M3 entry infers the Vision capability.

Checks:
- `git diff --check`
- `cargo test -p llmfit-core test_minimax_m3_infers_vision_capability`


AlexsJones (changes requested):
Good catch, M3 does take image input and the regression test against the embedded catalog is exactly the right shape. One thing to fix before this lands: the new `use_case` string drops the flagship facts (1M context, 128K max output) that the TUI shows users, and losing them isn't needed for the fix. Capability inference just looks for the substring `vision` or `multimodal` (`Capability::infer` in `llmfit-core/src/models.rs`, around line 352), so keeping the description and adding the keyword does both jobs. Something like:

`"Latest flagship: multimodal (text + image input), 1M context, 128K max output"`

in both `scripts/scrape_hf_models.py` and `llmfit-core/data/hf_models.json` (they need to stay in sync), and your test will still pass unchanged.

octo-patch:
Thanks for the review. I restored the flagship context and output details while keeping the multimodal text and image capability wording in both catalog sources, then pushed a new commit. I verified the scraper compiles and the two stored values match.

AlexsJones (changes requested):
Thanks for the quick follow-up, but the new strings landed on the wrong entries. In `hf_models.json` the flagship line (`Latest flagship: multimodal (text + image input), 1M context, 128K max output`) now sits on a 1k-context entry near line 19842, and in `scrape_hf_models.py` on the 131k gemma3 entry, while the actual MiniMax-M3 entries (1M context, architecture `minimax`) got the generic `Multimodal, vision and text`. The test still passes because either string triggers vision inference, which is why this slipped through. Swap them so M3 carries the flagship line and the other entries keep their original text, and this is good to go.

octo-patch:
Thanks for the review. I restored the unrelated catalog descriptions and moved the flagship multimodal wording to MiniMax-M3 in both `scripts/scrape_hf_models.py` and `llmfit-core/data/hf_models.json` as requested, then pushed a new commit. I ran the catalog source/generated assertions, `python3 -m py_compile scripts/scrape_hf_models.py`, and `cargo test -p llmfit-core test_minimax_m3_infers_vision_capability`.

AlexsJones (approved):
Thanks for the fix. The corrected `use_case` string now properly lands on the MiniMax-M3 entry and the new `test_minimax_m3_infers_vision_capability` test locks it in so this can't silently drift again. The scraper source is updated in sync too. Delta re-review LGTM.

octo-patch:
Thanks for the review. I merged the latest main to resolve the generated catalog conflict while preserving the MiniMax-M3 fallback description and vision regression test, then pushed commit ba0bb6c. I ran the targeted Rust test, Python compile, formatting check, and the full GitHub CI matrix is green.

AlexsJones (approved):
Thanks for pushing this through the catalog conflict, that file is no fun to rebase.

Re-reviewed just the merge commit. The resolution takes main's `hf_models.json`, and main already ships MiniMax-M3 with `capabilities: ["vision"]` and a multimodal `use_case` (the weekly refresh got there first), so what this PR still adds is the regression test plus the scraper string. Both are worth keeping: `test_minimax_m3_infers_vision_capability` is exactly the guard that would have caught the original miss, and it will keep catching it if a future refresh walks the entry back. CI is green across the three OSes.

Approving. One non-blocking thing for later: the hardcoded MiniMax-M3 fallback in `scripts/scrape_hf_models.py` still carries the old 230B / 117.9 GB numbers while the catalog now has 427B / 218.7 GB, so if that fallback ever fires it writes a stale entry. Fine as a follow-up.

Issue #804, opened 2026-07-27
[Feature]: CLI argument for filtering by providers


grafuls:
### Problem or motivation

If using `llmfit` programmatically by parsing the json output and without accessing the TUI, there is currently no way how to filter the results by providers.

### Proposed solution

I would like to have an argument like `--providers` that would take a list of providers to use as filter for the results.

### Feature area

CLI (new subcommand or flag)

### Would you be willing to contribute this?

Yes, I'd like to submit a PR

AlexsJones:
A `--providers` flag makes sense, filtering shouldn't require the TUI, and a PR would be very welcome. Two pointers before you start: the TUI's provider filter matches against `gguf_sources` as well as the canonical provider name, so reusing that matching (rather than a plain string compare) keeps CLI and TUI consistent. And I'd go comma-separated with case-insensitive matching, e.g. `llmfit fit --json --providers unsloth,bartowski`. There's a related open question on JSON value vocabularies in #803 if you want context on how the output format is evolving.

aramz33:
Pointer for whoever writes this, following the note above about reusing the TUI's matching rather than a plain string compare: #814 pulled that match out of the filter closure into two pure fns in `llmfit-tui/src/tui_app.rs`:

- `provider_selected(name, providers, selected) -> bool`
- `matched_gguf_provider(model, providers, selected) -> Option<&str>` — the GGUF publisher that got a row through the filter, `None` when the model's own provider is selected

They're pure over the name/selection slices, so `--providers` can call them with its own parsed list instead of the popup state. Two caveats: they live in the TUI crate, so a CLI flag would want them in `llmfit-core` first, and they compare exactly, so the case-insensitive part of `--providers unsloth,bartowski` has to happen in the caller.

SulimanAbdulrazzaq:
I'd like to take this if it's still available. I'll add the comma-separated, case-insensitive `--providers` filter to `fit`, reuse shared matching for both canonical providers and `gguf_sources`, and add focused coverage for canonical/source matches, OR semantics, and unknown providers. I'll keep the TUI behavior unchanged.

PR #814, opened 2026-07-30
feat(tui): mark list rows matched through a GGUF source

### LINKED ISSUE #569: Model listed in gguf_sources not listed in TUI
I tried to find relatively popular model `unsloth/Qwen3.6-35B-A3B-GGUF` in TUI but failed:

**Test Case 1** - Search="unsloth/Qwen3.6-35B-A3B" + Providers=ALL

Image

**Test Case 2** - Search="qwen GGUF" + Providers="unsloth"

Image

I see that the model is listed in `gguf_sources` in `llmfit-core/data/hf_models.json`:

```
 {
    "name": "Qwen/Qwen3.6-35B-A3B",
    "provider": "Alibaba",
    "parameter_count": "36.0B",
    [...]
    "gguf_sources": [
      {
        "repo": "unsloth/Qwen3.6-35B-A3B-GGUF",
        "provider": "unsloth"
      },

```

Is that working as inteded? Should `unsloth/Qwen3.6-35B-A3B-GGUF` be added separately?

Similar situation for `unsloth/gemma-4-26B-A4B-it-GGUF`.

llmfit 0.9.27
macOS 15.7.7

aramz33:
Closes #569.

Your ruling from the issue: the Provider column keeps the canonical name and takes a `*` when the row is only in the list because one of its GGUF publishers is selected. The detail pane names the actual repo.

### The three commits

1. **`refactor(tui)`** — the filter's `gguf_sources` half was inlined in the filter closure, so nothing downstream could tell which of the two names got a row through. Pulled out as `provider_selected` and `matched_gguf_provider`, both pure over the name/selection slices. No behaviour change: `matched_gguf_provider` returns `None` when the primary provider is selected, so `primary || matched.is_some()` is the same predicate as before — verified with the provider filter set to `unsloth` only, 366/4476 rows before and after.
2. **`feat(tui)`** — detail pane: the matched publisher in the GGUF Downloads list gets `▸` and the accent style. The marker replaces a leading space, so the column stays aligned.
3. **`feat(tui)`** — the list column, below.

### The marker gets its own column rather than the raw name

Appending to the provider string would have dropped the marker on exactly the rows it explains. 1382 catalog rows carry `gguf_sources`, spread over 495 distinct provider names, and around 150 of those are 12 chars or longer — `NousResearch` (12), `mlx-community` (13), `openai-community` (16), up to `optimum-intel-internal-testing` (30). Against `Constraint::Length(12)`, `NousResearch*` is clipped straight back to `NousResearch`.

The width is now a named constant so the cell formatter and the constraint can't drift apart. It replaces only the `Length(12)` that the table actually renders with: the other copy of the widths array, the one `model_col_text_width` is fed at `tui_ui.rs` L822, has drifted independently (14 constraints against the table's 15, no `disk`, three lengths disagreeing) and I would rather report that separately than hide it behind a rename here.

So a marked name is truncated to 11 chars and the marker owns column 12. Unmarked cells are returned untouched, so rows that matched on their own provider render exactly as they do today. `provider_cell` is pure and width-parameterised; four tests cover unmarked passthrough, `Alibaba*`, the 12-char-and-longer cases staying inside the column, and no marker at all when no provider filter is active (`vec![true; n]`).

### Before / after

Provider filter = `unsloth` only, 150x40 pty, debug build. `upstream/main` at afbaae1:

```
┌ Models (366/4476) ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────↑
│     Inst  Model                Provider     Params   Score ▼  tok/s*   Quant      Disk   Mode    Mem %   Ctx        Date     Fit        Use Case   █
│▶ ●   —    th/Qwen3-4B-Instruct unsloth      3.1B     88       45.0     Q8_0       3.3G   GPU     71%     4k         —        Perfect    Chat       █
│  ●   —    unsloth/Llama-3.2-3B unsloth      3.3B     87       42.7     Q8_0       3.5G   GPU     74%     4k         —        Perfect    Chat       ║
│  ●   —    unsloth/Llama-3.2-3B unsloth      3.3B     87       42.7     Q8_0       3.5G   GPU     74%     4k         —        Perfect    Chat       ║
│  ●   —    Qwen/Qwen2.5-Coder-3 Alibaba      3.1B     86       45.6     Q8_0       3.2G   GPU     73%     32k        —        Perfect    Coding     ║
│  ●   —    unsloth/Qwen2.5-Code unsloth      3.2B     86       44.4     Q8_0       3.3G   GPU     75%     32k        —        Perfect    Coding     ║
```

Same filter, this branch:

```
┌ Models (366/4476) ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────↑
│     Inst  Model                Provider     Params   Score ▼  tok/s*   Quant      Disk   Mode    Mem %   Ctx        Date     Fit        Use Case   █
│▶ ●   —    07-unsloth-bnb-4bit  unsloth      3.1B     88       45.0     Q8_0       3.3G   GPU     71%     4k         —        Perfect    Chat       █
│  ●   —    unsloth/Llama-3.2-3B unsloth      3.3B     87       42.7     Q8_0       3.5G   GPU     74%     4k         —        Perfect    Chat       ║
│  ●   —    unsloth/Llama-3.2-3B unsloth      3.3B     87       42.7     Q8_0       3.5G   GPU     74%     4k         —        Perfect    Chat       ║
│  ●   —    Qwen/Qwen2.5-Coder-3 Alibaba*     3.1B     86       45.6     Q8_0       3.2G   GPU     73%     32k        —        Perfect    Coding     ║
│  ●   —    unsloth/Qwen2.5-Code unsloth      3.2B     86       44.4     Q8_0       3.3G   GPU     75%     32k        —        Perfect    Coding     ║
```

366/4476 in both — display only. The `DeepSeek`, `Meta`, `Microsoft` and `TII` rows further down the same list are marked the same way. (The Model cell of the selected row marquees, so the two captures caught it at different scroll phases; same model.)

Truncation case, filter = `mradermacher`, search `Zion` — own provider `sicariussicariistuff`, 20 chars. Before:

```
┌ Models (3/4476) ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│     Inst  Model                Provider     Params   Score ▼  tok/s*   Quant      Disk   Mode    Mem %   Ctx        Date     Fit        Use Case   │
│▶ ●   —    riiStuff/Zion_Alpha_ sicariussica 7.2B     72       27.9     Q4_K_M     4.2G   GPU     90%     4k         —        Marginal   Chat       │
```

After:

```
┌ Models (3/4476) ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│     Inst  Model                Provider     Params   Score ▼  tok/s*   Quant      Disk   Mode    Mem %   Ctx        Date     Fit        Use Case   │
│▶ ●   —    riiStuff/Zion_Alpha_ sicariussi…* 7.2B     72       27.9     Q4_K_M     4.2G   GPU     90%     4k         —        Marginal   Chat       │
```

Local: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` (no new warnings), `cargo test` — 577 pass.

### One thing I left to you

`*` is not spelled out anywhere in the UI. Since #783 replaced the cryptic `R`/`M` markers with the range spelled out in the Fit/Filter box, you may want the same treatment here. I kept it out to hold the PR to the issue: happy to add a line naming the provider marker in that box, or leave the detail pane as the only explanation.

### DIFF
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -856,7 +856,7 @@ fn provider_selected(name: &str, providers: &[String], selected: &[bool]) -> boo
 /// by a GGUF publisher surfaces rows whose own provider was never selected.
 /// Returns `None` when the primary provider is selected (nothing to attribute)
 /// or when nothing matched at all.
-pub fn matched_gguf_provider<'a>(
+pub(crate) fn matched_gguf_provider<'a>(
     model: &'a LlmModel,
     providers: &[String],
     selected: &[bool],
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -5306,6 +5306,15 @@ mod tests {
     }
 
     #[test]
+    fn matched_gguf_provider_is_none_without_a_provider_filter() {
+        // No filter is `vec![true; n]` (see `App::new`), so every row is there
+        // on its own merit and the list marker must stay off.
+        let model = model_with_gguf_sources();
+        let selected = vec![true, true, true];
+        assert_eq!(matched_gguf_provider(&model, &providers(), &selected), None);
+    }
+
+    #[test]
     fn matched_gguf_provider_returns_first_selected_source_in_catalog_order() {
         let model = model_with_gguf_sources();
         let selected = vec![false, true, true];
--- a/llmfit-tui/src/tui_app.rs
+++ b/llmfit-tui/src/tui_app.rs
@@ -1,6 +1,6 @@
 use llmfit_core::fit::{CalcConfig, FitLevel, ModelFit, SortColumn, backend_compatible};
 use llmfit_core::hardware::SystemSpecs;
-use llmfit_core::models::{Capability, ModelDatabase, UseCase};
+use llmfit_core::models::{Capability, LlmModel, ModelDatabase, UseCase};
 use llmfit_core::plan::{PlanEstimate, PlanRequest, estimate_model_plan};
 use llmfit_core::providers::{
     self, DockerModelRunnerProvider, LlamaCppProvider, LmStudioProvider, MlxProvider,
@@ -837,6 +837,38 @@ fn fuzzy_match(query: &str, candidate: &str) -> bool {
         }
     }
     q.peek().is_none()
+}
+
+/// Whether `name` is one of the currently selected entries of the provider filter.
+fn provider_selected(name: &str, providers: &[String], selected: &[bool]) -> bool {
+    providers
+        .iter()
+        .position(|p| p == name)
+        .and_then(|idx| selected.get(idx).copied())
+        .unwrap_or(false)
+}
+
+/// The GGUF-source provider that got this model through the provider filter,
+/// when the model's own provider is not selected.
+///
+/// The catalog lists base models (e.g. `Qwen/Qwen3-4B` from Alibaba) while the
+/// GGUF quants are published by others (unsloth, bartowski, ...), so filtering
+/// by a GGUF publisher surfaces rows whose own provider was never selected.
+/// Returns `None` when the primary provider is selected (nothing to attribute)
+/// or when nothing matched at all.
+pub fn matched_gguf_provider<'a>(
+    model: &'a LlmModel,
+    providers: &[String],
+    selected: &[bool],
+) -> Option<&'a str> {
+    if provider_selected(&model.provider, providers, selected) {
+        return None;
+    }
+    model
+        .gguf_sources
+        .iter()
+        .find(|gs| provider_selected(&gs.provider, providers, selected))
+        .map(|gs| gs.provider.as_str())
 }
 
 pub struct App {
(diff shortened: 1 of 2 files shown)


aramz33:
Greptile's second-round finding is real, fixed in ea25ffd.

The detail pane compared each source against the single value `matched_gguf_provider` returns, so with two publishers selected only the one earliest in the catalog got the `▸` — a strict equality against one name can never mark two. It now asks per source: the row came in through a publisher rather than its own provider, and this source's publisher is selected.

`bartowski` + `mradermacher` selected, `huihui-ai/Qwen2.5-Coder-3B-Instruct-abliterated` (own provider not selected):

```
  ── GGUF Downloads ──

▸ 📦  bartowski
       hf.co/bartowski/Qwen2.5-Coder-3B-Instruct-abliterated-GGUF
▸ 📦  mradermacher
       hf.co/mradermacher/Qwen2.5-Coder-3B-Instruct-abliterated-GGUF
```

Before the fix only `bartowski` carried the marker. Nothing is marked when the model's own provider is selected, unchanged. `matched_gguf_provider` still returns the first match — the list column only needs to know that there was one, so the `*` behaviour and its tests are untouched. One test added for the multi-publisher predicate; `cargo test` 578 pass.

AlexsJones (approved):
Thanks for carrying this from diagnosis to fix so cleanly. The marker owning column 12 is the right call (appending to the raw name would have clipped it off exactly on the rows it explains, as your captures show), and pulling `provider_selected` / `matched_gguf_provider` out as pure fns sets up #804's `--providers` nicely.

Two nits, neither blocking: `gguf_source_attributed` only ever uses its `Option<&str>` as `.is_some()`, so a `bool` param would be more honest, and the detail pane re-runs `provider_selected` per source after `matched_gguf_provider` already scanned them (fine at current scale, just noting).

CI is green 8/8 and the truncation cases are unit-tested, including the 12-char boundary. LGTM.

PR #818, opened 2026-07-31
fix(share): give the declared memory tier a floor below 8 GB

### LINKED ISSUE #816: `nearest_mem_tier` has no tier below 8 GB, and the two copies have diverged
`nearest_mem_tier` exists twice, with two different ladders, and neither covers GPUs
below 8 GB.

Surfaced by the Greptile review on #815, which flagged that a 4 GB GTX 1050 Ti was
reporting `memTierGb: 8`:
https://github.com/AlexsJones/llmfit/pull/815#discussion_r3684606322

The value was not hand-written — it is what `llmfit 1.1.3` emits. Digging into why
turned up two separate problems.

### 1. No tier below 8 GB

Both copies start the ladder at 8, so every sub-8 GB GPU is reported as an 8 GB part:

| actual VRAM | reported `memTierGb` |
|---|---|
| 2.0 | 8 |
| 4.0 | 8 |
| 6.0 | 8 |
| 8.0 | 8 |

For a capacity field this errs in the dangerous direction — it overstates memory. A
4 GB card is grouped with 8 GB cards in any tier-based comparison, and in
`benchmarks.rs` it also queries community data recorded on 8 GB hardware, which is
precisely the data it should not be matched against.

This has stayed invisible so far because every existing entry in
`llmfit-core/data/community/` is 8 GB or more, so `nearest_mem_tier` happens to return
their real VRAM. The 1050 Ti is the first submission below the lowest rung.

### 2. The two copies have diverged

`llmfit-core/src/benchmarks.rs:305`

```rust
const TIERS: [u32; 9] = [8, 12, 16, 24, 32, 48, 80, 96, 128];
```

`llmfit-core/src/share.rs:116`

```rust
const TIERS: [u32; 12] = [8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
```

Same name, same algorithm, different constants — `share.rs` has `64`, `192` and `256`,
`benchmarks.rs` does not. The bodies differ only by a local variable name
(`best_dist` vs `best_d`), which suggests one was copied from the other and they
drifted.

The consequence is that a single machine can be classified into two different tiers
depending on the code path:

| VRAM | `benchmarks.rs` (lookup) | `share.rs` (submission) |
|---|---|---|
| 60 GB | 48 | **64** |
| 150 GB | 128 | **192** |
| 200 GB | 128 | **256** |

The existing `tesla-t4` entry carries `memTierGb: 64`, a value the `benchmarks.rs`
ladder cannot produce — so that submission can never match its own recorded tier when
the lookup path runs.

### Call sites

- `benchmarks.rs:280` and `:285` — builds the `memTier` query parameter for the
  community benchmark lookup
- `share.rs:139` and `:142` — writes `memTierGb` into the submission payload

### 3. Which capacities actually exist below 8 GB

The gap is not hypothetical. Cards that shipped under 8 GB and are still in wide
circulation:

| VRAM | cards | current tier | error |
|---|---|---|---|
| 2 GB | GT 710, GTX 1050 | 8 | +300% |
| 3 GB | GTX 1060 3 GB, GTX 780 | 8 | +167% |
| 4 GB | GTX 1050 Ti, RX 570 4 GB | 8 | +100% |
| 6 GB | GTX 1060 6 GB, RTX 2060, RTX 3050 6 GB | 8 | +33% |

The GTX 1060 3 GB in particular is one of the highest-volume discrete GPUs ever
shipped, and it is currently declared at nearly three times its real capacity.

### 4. A separate, milder gap at 10 and 11 GB

Worth flagging while you are in here, though it is a different problem and I am not
proposing to touch it:

| VRAM | cards | tier | error | direction |
|---|---|---|---|---|
| 10 GB | RTX 3080 10 GB | 8 | -20% | pessimistic |
| 11 GB | GTX 1080 Ti, RTX 2080 Ti | 12 | +9% | **optimistic** |

10 GB sits exactly between 8 and 12; the `d < best_d` comparison keeps the first
match, so it lands on 8. 11 GB is strictly closer to 12.

The directions matter because this value also drives the lookup in `benchmarks.rs`.
A 3080 owner is served results from 8 GB cards, which undersells the card but never
promises a model that will not fit. An 11 GB owner is served results from 12 GB
cards, which can promise a model that OOMs. That is the direction that actually
hurts, but the magnitude (9%) is within what a coarse bucket is presumably meant to
absorb — unlike the +100% to +300% at the bottom of the ladder.

### Suggested direction, revised

The doc comment describes these as "coarse buckets so submissions group cleanly", and
coarse grouping is clearly the intent. So the fix I would suggest is narrow: give the
bottom of the ladder a floor, and leave everything at and above 8 alone.

For the **submission path only** (`share.rs`):

```rust
const TIERS: [u32; 16] = [2, 3, 4, 6, 8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
```

Properties, deliberately:

- No tier at or above 8 moves, so **no previously submitted value changes**.
- Only inputs below about 7 GB are affected, and no shipping card sits in the
  6.5-7.0 band that shifts.
- Exact ties keep resolving downward, which understates rather than overstates — the
  safe direction for a declared capacity.
- The existing test `mem_tier_rounds_to_nearest` (7.5 -> 8) still passes.

I would **not** dedupe the two copies, despite that being the obvious reading of
section 2. `nearest` is the correct rule for a lookup key and the wrong one for a
declared capacity; merging them would freeze that conflation rather than resolve it.
If you want them unified, they probably want two names and two intents rather than
one function.

I have this written and tested locally against `main` (494 tests pass,
`cargo fmt --check` and `cargo clippy --all-targets --all-features` clean) but have
not opened a PR — happy to send it if the shape above is what you want, or to adjust
it if you would rather include 10 and 11, or handle the duplication differently.

Akciali:
Closes #816.

`nearest_mem_tier` had no rung below 8, so any GPU with less than 8 GB of VRAM was
published as an 8 GB part. A 4 GB card was declared at twice its capacity, a 2 GB one
at four times. For a coarse grouping key that is not imprecision, it is a missing
floor: small cards get bucketed upward into hardware they should not be compared
against.

### What changes

```diff
- const TIERS: [u32; 12] = [8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
+ const TIERS: [u32; 16] = [2, 3, 4, 6, 8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
```

The four added rungs are the capacities that actually shipped below 8 GB: GT 710 and
GTX 1050 at 2, GTX 1060 3 GB, GTX 1050 Ti and RX 570 at 4, GTX 1060 6 GB and RTX 2060
at 6.

Deliberate properties:

- **No tier at or above 8 moves**, so no previously submitted value changes.
- Only inputs below roughly 7 GB are affected, and no shipping card sits in the
  6.5 to 7.0 band that shifts.
- Exact ties still resolve downward, understating rather than overstating. That is
  the safe direction for a value claiming what a machine has.
- The existing rounding test still passes untouched (7.5 gives 8). It is renamed to
  `declared_mem_tier_rounds_to_nearest` to match the function.

### Your two asks

**Floor the submission ladder, leave 8 and above alone, ties keep resolving down.**
Done, as above.

**Fix the 1050 Ti entry from #815.** Already in `main`. It was corrected to
`memTierGb: 4` during the Greptile review on that PR, before it merged, so there is
nothing to do here and I have left the data file untouched to keep this a
one-concern change. Worth a quick check on your side in case we are looking at
different states.

**Make the split legible.** The two copies are renamed to carry their intent:

| file | name | role |
|---|---|---|
| `share.rs` | `declared_mem_tier` | the capacity a submission *declares* |
| `benchmarks.rs` | `lookup_mem_tier` | the bucket used as a *lookup key* |

Each carries a doc comment explaining why nearest-match is correct for one and wrong
for the other. A lookup wants the closest bucket that has data, because returning an
approximation beats returning nothing; a declared capacity wants the opposite. Each
points at the other by name, and the one in `benchmarks.rs` ends with "Please do not
unify them."

### Deliberately unchanged

`benchmarks.rs` keeps its own ladder and its exact current behaviour. This PR changes
what a submission *declares*, never which cached results a machine is matched
against. 10 GB and 11 GB are left alone as agreed.

### Verification

Software, using the commands from `CONTRIBUTING.md`:

| | |
|---|---|
| `make test` | 498 passed, 0 failed |
| `make fmt` | clean |
| `make clippy` | clean, no new warnings |
| `cargo doc` | same warning count as `main`, none introduced |

Three tests added, including an end-to-end one that runs `build_submission` with a
4 GB spec and asserts `memTierGb == 4` on the serialised payload rather than on the
helper's return value.

Hardware, on the card that surfaced this. GTX 1050 Ti back in the machine, driver
550.163.01, llama.cpp `da5b448` with `GGML_VULKAN=ON`, three models benchmarked with
the patched binary after a GPU preheat. `llmfit bench --share --dry-run` emits:

```json
"hardwareName": "NVIDIA GeForce GTX 1050 Ti",
"memTierGb": 4,
"vramGb": 4.0
```

All three entries in the local store agree. Before the patch the same payload
declared 8.

### Unrelated, spotted while validating

The `model` field written into the pending store carries the **absolute path** of the
GGUF (`/home/.../gguf/SmolLM2-135M-Instruct-Q4_K_M.gguf`) for the `llamacpp` provider,
so every submitter has to strip it by hand before opening a PR. Out of scope here.
Happy to file it separately if it is not already known.

---

@AlexsJones thanks again for this project.

### DIFF
--- a/llmfit-core/src/benchmarks.rs
+++ b/llmfit-core/src/benchmarks.rs
@@ -277,12 +277,12 @@ pub fn hw_leaderboard_params(specs: &SystemSpecs) -> Vec<(&'static str, String)>
 
     // VRAM tier
     if let Some(vram) = specs.total_gpu_vram_gb {
-        let tier = nearest_mem_tier(vram);
+        let tier = lookup_mem_tier(vram);
         if tier > 0 {
             params.push(("memTier", tier.to_string()));
         }
     } else if specs.unified_memory {
-        let tier = nearest_mem_tier(specs.total_ram_gb);
+        let tier = lookup_mem_tier(specs.total_ram_gb);
         if tier > 0 {
             params.push(("memTier", tier.to_string()));
         }
@@ -301,7 +301,18 @@ pub fn hw_leaderboard_params(specs: &SystemSpecs) -> Vec<(&'static str, String)>
     params
 }
 
-fn nearest_mem_tier(gb: f64) -> u32 {
+/// Bucket a memory size into the tier used as a *lookup key* against the
+/// cached community results.
+///
+/// Nearest-match is the right rule here and the wrong one in `share.rs`: a
+/// lookup wants the closest bucket that has data even when the fit is not
+/// exact, because returning nothing is worse than returning an approximation.
+/// A declared capacity wants the opposite.
+///
+/// That is why this is not the same function as `declared_mem_tier` in
+/// `share.rs` despite looking almost identical, and why the two ladders are
+/// allowed to differ. Please do not unify them.
+fn lookup_mem_tier(gb: f64) -> u32 {
     const TIERS: [u32; 9] = [8, 12, 16, 24, 32, 48, 80, 96, 128];
     let mut best = 0u32;
     let mut best_dist = f64::MAX;
--- a/llmfit-core/src/share.rs
+++ b/llmfit-core/src/share.rs
@@ -110,10 +110,22 @@ fn os_name() -> &'static str {
     }
 }
 
-/// Round a memory size to the nearest common tier (matches the leaderboard's
-/// coarse buckets so submissions group cleanly).
-fn nearest_mem_tier(gb: f64) -> u32 {
-    const TIERS: [u32; 12] = [8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
+/// Round a memory size to the nearest common tier, for the capacity a
+/// submission *declares* about the machine it ran on.
+///
+/// The ladder is deliberately coarse so submissions group cleanly, but it has
+/// a floor: capacities that shipped below 8 GB get their own rungs, otherwise
+/// a 4 GB card is published as an 8 GB part and compared against hardware it
+/// has nothing in common with.
+///
+/// Exact ties resolve downward, which understates rather than overstates. That
+/// is the safe direction here, because this value is a claim about what a
+/// machine actually has.
+///
+/// Not to be merged with `lookup_mem_tier` in `benchmarks.rs`, which looks
+/// almost identical and means something else. See the note there.
+fn declared_mem_tier(gb: f64) -> u32 {
+    const TIERS: [u32; 16] = [2, 3, 4, 6, 8, 12, 16, 24, 32, 48, 64, 80, 96, 128, 192, 256];
     let mut best = 0u32;
     let mut best_d = f64::MAX;
     for &t in &TIERS {
@@ -136,10 +148,10 @@ fn build_submission(results: &[BenchResult], specs: &SystemSpecs) -> Submission 
     };
 
     let mem_tier_gb = if let Some(vram) = specs.total_gpu_vram_gb {
-        let t = nearest_mem_tier(vram);
+        let t = declared_mem_tier(vram);
         (t > 0).then_some(t)
     } else if specs.unified_memory {
-        let t = nearest_mem_tier(specs.total_ram_gb);
+        let t = declared_mem_tier(specs.total_ram_gb);
         (t > 0).then_some(t)
     } else {
         None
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Fast turnaround, and thanks for putting the 1050 Ti back in the machine to verify the payload on real hardware. This lands exactly what we agreed on #816: new rungs only below 8, everything at 8 and above untouched, ties still resolving down. The rename plus the paired doc comments should keep the next reader from "helpfully" unifying the two ladders, which was the point.

The end-to-end test asserting `memTierGb` on the serialised payload is the right guard, it protects the whole `build_submission` path rather than just the helper. And you're right on the #815 entry, it was corrected to 4 during review before that PR merged, so leaving the data file untouched here was the correct call.

The absolute path leaking into the pending store's `model` field is worth its own issue, please do file it.

Akciali:
Thanks for the review. Filed the pending store path leak as #819.

PR #823, opened 2026-08-02
feat(cli): add provider filter to fit

### LINKED ISSUE #804: [Feature]: CLI argument for filtering by providers

SulimanAbdulrazzaq:
## Summary

Add a `--providers` option to `llmfit fit` for comma-separated, case-insensitive provider filtering.

The filter matches both a model's canonical provider and its GGUF source publishers, as requested. Matching is shared with the TUI while preserving the TUI's existing exact-selection behavior. CLI filtering runs before ranking and `--limit`, and the help text includes an agent-mode example.

Fixes #804.

## Validation

Using the stable `x86_64-pc-windows-gnu` toolchain:

- Core provider matcher regression: 1 passed
- CLI JSON integration regression: 1 passed
- Existing provider selector tests: 2 passed
- Existing GGUF attribution tests: 6 passed
- `cargo check -p llmfit`
- `llmfit fit --help` smoke test
- `git diff --check`

The CLI integration test exercises mixed-case, comma-separated input, an unknown provider, GGUF-source matching, JSON output, and filtering before `--limit`.

## Environment limitations

- `cargo check --workspace` reaches an existing `llmfit-desktop` Tauri configuration failure: the current schema rejects the untouched `downloadBootstrapper` value in `tauri.conf.json`.
- Package-scoped clippy with `-D warnings` reaches existing warnings in untouched core code (including `fit.rs`, `hardware.rs`, `quality.rs`, and `update.rs`); the new provider-filter code introduced no reported lint warning.
- The MSVC linker is not installed on this machine, so successful Rust validation used the GNU Windows target.

### DIFF
--- a/llmfit-core/src/models.rs
+++ b/llmfit-core/src/models.rs
@@ -686,6 +686,18 @@ pub struct GgufSource {
     pub provider: String,
 }
 
+/// Returns whether a model's canonical provider or any GGUF publisher matches.
+///
+/// The caller supplies the matching rule so interactive filters can retain
+/// exact selection semantics while CLI filters can be case-insensitive.
+pub fn matches_provider_filter(model: &LlmModel, mut matches: impl FnMut(&str) -> bool) -> bool {
+    matches(&model.provider)
+        || model
+            .gguf_sources
+            .iter()
+            .any(|source| matches(&source.provider))
+}
+
 impl LlmModel {
     /// MLX models are Apple-only — they won't run on NVIDIA/AMD/Intel hardware.
     /// We detect them by the `-MLX-` suffix that's standard on HuggingFace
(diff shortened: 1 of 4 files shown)


AlexsJones (approved):
Thanks for turning the #804 ask around this fast. Pulling `matches_provider_filter` into core and rewiring the TUI's filter through it with the exact-selection closure is the right shape: one matching rule, two selection semantics, and the smoke test proves the GGUF-source path by asserting every returned model matched on a publisher rather than its canonical provider. Filter before rank and `--limit`, trimmed case-insensitive values, and the agent-mode help example all line up with what was agreed on the issue. Approving. The build matrix still needs a maintainer click to run since this is your first contribution here, so the merge waits on that coming back green.

SulimanAbdulrazzaq:
Thank you for the detailed review and merge. I appreciate the confirmation that the shared matcher, filtering order, TUI semantics, and GGUF-source coverage match the intended behavior.

PR #833, opened 2026-08-05
fix: normalize llama.cpp GGUF model IDs

### LINKED ISSUE #819: Absolute GGUF path leaks into the pending store's `model` field for the llamacpp provider
## Symptom

When benchmarking against a llama-server instance, the `model` field stored in the
pending submission payload is the absolute filesystem path of the loaded GGUF, not
a model name:

```json
"results": [
  {
    "model": "/home/llmfit/gguf/arcee-ai_AFM-4.5B-Q4_K_M.gguf",
    "provider": "llamacpp",
    ...
  }
]
```

Every payload produced during the GTX 1050 Ti campaign (PR #815, 63 models) came
out this way and had to be cleaned by hand before opening the PR, down to
`arcee-ai_AFM-4.5B-Q4_K_M.gguf`.

## Code path

1. `llama-server` reports the value of its `-m/--model` argument, usually an
   absolute path, as the model `id` in its OpenAI-compatible `/v1/models` listing.
2. `llmfit-core/src/bench.rs`: `detect_openai_model` and `list_openai_models`
   take `data[].id` verbatim, so the path becomes the model name for the
   llamacpp target. `bench_openai_compat` then stores it as `BenchResult.model`.
3. `llmfit-core/src/share.rs`: `build_submission` copies `r.model.clone()` into
   `ResultPayload.model`, and `store_local` serialises it into
   `pending/<timestamp>-<hash>.json`.

Nothing between the `/v1/models` response and the pending store normalises the
string. Ollama, vLLM and MLX are unaffected because their model ids are tags or
HF-style names, not filesystem paths.

## Impact on contributors

- `llmfit share` would publish the raw path, so every contributor has to edit
  each pending JSON by hand before opening a community-benchmarks PR. With one
  file per model, a large campaign means dozens of manual edits.
- Privacy: absolute paths routinely embed a username or machine-specific
  directory layout, which contributors may not notice they are publishing.
- Leaderboard grouping: the same model benchmarked from two machines produces
  two different `model` strings, so results that should aggregate do not.

## Suggested direction

Normalise at ingestion, where the provider is known to be llamacpp: when the id
returned by `/v1/models` is a path (contains a separator, ends in `.gguf`), keep
only the file name, for example via `Path::file_name`. Doing it in
`detect_openai_model` / `list_openai_models` fixes the TUI display, the pending
store and the shared payload in one place. An alternative is a guard in
`build_submission`, which would also sanitise payloads produced by older
binaries still sitting in `pending/`, at the cost of leaving the TUI showing the
path. Happy to take this on and open a PR if the ingestion-side approach sounds
right.

Existing hand-cleaned community data files are already in basename form, so
normalising to the file name (keeping the `.gguf` extension, as in the merged
data) matches current expectations and requires no data migration.

AlexsJones:
## Summary
- normalize path-like GGUF IDs returned by llama-server to their filename
- apply normalization during llama.cpp auto-detection and multi-model discovery
- preserve non-path and non-GGUF IDs, and leave vLLM/MLX discovery unchanged

## Testing
- cargo fmt --check
- cargo test -p llmfit-core bench::tests
- cargo check --workspace (fails in existing llmfit-desktop Tauri config: downloadBootstrapper is not accepted by the installed schema)

Fixes #819

### DIFF
--- a/llmfit-core/src/bench.rs
+++ b/llmfit-core/src/bench.rs
@@ -400,7 +400,7 @@ pub fn auto_detect_target(model_hint: Option<&str>) -> Result<BenchTarget, Strin
     // llama.cpp answers /props, so it can be identified positively.
     let llama_url = llamacpp_url();
     if probe_llamacpp(&llama_url)
-        && let Ok(model_name) = detect_openai_model(&llama_url, model_hint)
+        && let Ok(model_name) = detect_llamacpp_model(&llama_url, model_hint)
     {
         return Ok(BenchTarget::LlamaCpp {
             url: llama_url,
@@ -460,7 +460,7 @@ pub fn discover_all_targets() -> Vec<BenchTarget> {
     // llama.cpp answers /props, so it can be identified positively.
     let llama_url = llamacpp_url();
     let llamacpp_found = probe_llamacpp(&llama_url);
-    if llamacpp_found && let Ok(models) = list_openai_models(&llama_url) {
+    if llamacpp_found && let Ok(models) = list_llamacpp_models(&llama_url) {
         for model in models {
             targets.push(BenchTarget::LlamaCpp {
                 url: llama_url.clone(),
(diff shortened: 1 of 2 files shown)


AlexsJones:
Reviewed: this is exactly the right scope for #819. Reading list_llamacpp_models and detect_llamacpp_model as thin wrappers over the OpenAI variants keeps the llama.cpp normalization in one place, and normalize_llamacpp_model_id correctly refuses non-GGUF and pathless IDs (the /models/config.json case) so discovery and model grouping stay intact. Tests cover the Unix path, the Windows separator, and the leave-alone cases. CI green.

One non-blocking nit: auto_detect_target passes the caller hint straight into detect_llamacpp_model, so a hinted full GGUF path would be matched against the already-normalized /v1/models response names and may not match. Corner case, fine as a follow-up if it ever bites.

Looks good to me. "Fixes #819" is accurate, this closes out that leak. (I can not self-approve this PR, so its Ready row reads "awaiting your approve + merge".)

PR #834, opened 2026-08-05
fix(plan): honour CalcConfig so plan and fit agree under tuned settings

marzmesas:
Follow-up to #807, split out at your request.

## Summary

`plan` reached `fit::estimate_tps` with a hardcoded `CalcConfig::default()`, so user-tuned values from the TUI Advanced Configuration panel never arrived. `tui_app.rs` calls `estimate_model_plan` with `self.calc_config` in scope and unused, so a user who raised `efficiency` to calibrate against measured benchmarks saw the fit table and the plan panel disagree — despite #807 making them share one estimator.

Pre-existing rather than introduced by #807: `default_efficiency()` is a bare `0.55`, exactly the value the old hardcoded formula used, which is why #807's dense measurements came out byte-identical. But it left a gap in that PR's "can't diverge again" claim, which this closes.

## Changes

```rust
estimate_model_plan_with_config(model, request, system, &CalcConfig)   // new
estimate_model_plan(...)  →  wrapper passing CalcConfig::default()     // signature unchanged
```

The public signature is untouched, so `serve_api.rs`, `mcp_server.rs` and `main.rs` need no edits. The config is threaded through `evaluate_current`, `build_path_estimate`, `minimum_cores_for_target` and both `estimate_tps` helpers; the TUI call site passes `self.calc_config`.

`LLMFIT_DDR_BANDWIDTH` already worked, since `ddr_bandwidth_gbps()` reads the env var regardless of the struct — only the panel value was being dropped.

As discussed, `build_path_estimate` ends up one argument over clippy's limit and is annotated rather than wrapped in a params struct, which would churn every call site for no gain.

## Tests

- `test_plan_honours_calc_config_efficiency` — raising `efficiency` 0.55 → 0.80 must change the estimate, and must still match `fit` exactly under that same config
- `test_estimate_model_plan_defaults_match_explicit_default_config` — the wrapper must stay equivalent to passing `CalcConfig::default()`, so the public entry point's behaviour is pinned

## Verification

- `cargo test -p llmfit-core` — **528 passed, 0 failed, 1 ignored**
- `cargo fmt --check` — clean
- `cargo clippy -p llmfit-core --all-targets -- -D warnings` — **23 errors on `main` and 23 with this commit**, i.e. unchanged; they are pre-existing under clippy 1.97, newer than CI's pinned version
- Default-path output is unchanged, since `CalcConfig::default().efficiency` is the same 0.55

Rebased onto `main` after #807 and #813 landed. Three conflicts, all "both sides wanted" rather than
disagreements: the `use` line in `tui_app.rs` (upstream added `LlmModel`/`matches_provider_filter`
while this commit changed the adjacent `plan` import), and the `CpuOffload`/`CpuOnly` arms of
`build_path_estimate`, where #813's `fit_level_for(..., system.available_ram_gb, rec_ram)` sits on the
line above the `estimate_tps` call this commit threads `config` into. Both changes are preserved in
each case. The six `build_path_estimate` call sites #813 added to the test module also needed the new
argument.

Tested on Ubuntu 22.04.5, RTX 3090 24 GB, Ryzen 9 3900X, 32 GB RAM.

### DIFF
--- a/llmfit-core/src/lib.rs
+++ b/llmfit-core/src/lib.rs
@@ -19,7 +19,8 @@ pub use hardware::{GpuBackend, SystemSpecs};
 pub use models::{Capability, LlmModel, ModelDatabase, ModelFormat, UseCase};
 pub use plan::{
     HardwareEstimate, PathEstimate, PlanCurrentStatus, PlanEstimate, PlanRequest, PlanRunPath,
-    UpgradeDelta, estimate_model_plan, normalize_quant, resolve_model_selector,
+    UpgradeDelta, estimate_model_plan, estimate_model_plan_with_config, normalize_quant,
+    resolve_model_selector,
 };
 pub use providers::{
     LlamaCppProvider, LmStudioProvider, MlxProvider, ModelProvider, OllamaProvider, VllmProvider,
--- a/llmfit-core/src/plan.rs
+++ b/llmfit-core/src/plan.rs
@@ -1,4 +1,4 @@
-use crate::fit::{FitLevel, RunMode};
+use crate::fit::{CalcConfig, FitLevel, RunMode};
 use crate::hardware::{GpuBackend, SystemSpecs};
 use crate::models::{KvQuant, LlmModel, quant_speed_multiplier};
 
@@ -182,8 +182,9 @@ fn estimate_tps(
     backend: GpuBackend,
     path: PlanRunPath,
     cpu_cores: usize,
+    config: &CalcConfig,
 ) -> f64 {
-    estimate_tps_with_gpu(model, quant, backend, path, cpu_cores, None)
+    estimate_tps_with_gpu(model, quant, backend, path, cpu_cores, None, config)
 }
 
 /// Run mode to estimate speed under.
@@ -212,8 +213,9 @@ fn estimate_tps_with_gpu(
     path: PlanRunPath,
     cpu_cores: usize,
     system: Option<&SystemSpecs>,
+    config: &CalcConfig,
 ) -> f64 {
-    use crate::fit::{CalcConfig, InferenceRuntime};
+    use crate::fit::InferenceRuntime;
     use crate::hardware::gpu_memory_bandwidth_gbps;
 
     let params = model.params_b().max(0.1);
@@ -240,7 +242,7 @@ fn estimate_tps_with_gpu(
             specs,
             speed_run_mode(path, model),
             InferenceRuntime::LlamaCpp,
-            &CalcConfig::default(),
+            config,
         );
     }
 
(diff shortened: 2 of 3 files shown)


marzmesas:
The review summary was right that the fallback path needed fixing too, so I've addressed it in `3aee005` rather than leave it for a follow-up — it's the same claim this PR makes, just on the branch I hadn't followed.

**What was wrong.** `estimate_tps_with_gpu` took `config` but only used it in the delegation branch. The fixed-constant fallback below it — reached for unrecognised GPUs, and *always* for `CpuOnly` since both modules gate their bandwidth path on `!= CpuOnly` — hardcoded its run-mode penalties:

```rust
PlanRunPath::CpuOffload => base *= 0.5,   // ignored config.run_mode_factors
PlanRunPath::CpuOnly    => { /* recompute with CPU K, no factor at all */ }
```

`fit.rs`'s fallback ends with `base *= config.run_mode_factors.for_run_mode(run_mode)`. So `plan` both ignored tuning *and* diverged from `fit` at defaults on the CPU-only path, by `1 / cpu_only` — a factor of 3.3.

**Fix:** drop the hardcoded penalties and apply `config.run_mode_factors.for_run_mode(path.run_mode())` at the end, mirroring `fit.rs`.

**Behaviour change, disclosed.** At default config only `cpu_only` moves, and it moves toward `fit`:

| path | before | after |
|---|---|---|
| `gpu` | 31.43 | 31.43 |
| `cpu_offload` | 15.71 | 15.71 |
| `cpu_only` | 2.46 | **0.74** |

(`Qwen3-32B`, Q4_K_M, 32k.) `gpu` and `cpu_offload` are untouched because the hardcoded `0.5` happened to equal `run_mode_factors.cpu_offload`'s default. `cpu_only` drops by exactly the 0.3 default factor it was previously skipping — a visible change to `plan`'s CPU-only numbers, but it makes them agree with what `fit` has always reported for the same machine.

New test `test_plan_cpu_only_speed_matches_fit_and_honours_config` asserts parity against `fit::estimate_tps` at both the default `cpu_only` and a tuned `0.15`, so neither the divergence nor the ignored-config half can come back.

Two related gaps I deliberately left alone, since they're pre-existing and not what this PR claims:

- The fallback has no `MoeOffload` branch, so a MoE model on an unrecognised GPU misses the DDR-bottleneck model that `fit.rs` applies there. Worth its own issue if you agree.
- `fit.rs`'s `MoeOffload` fallback early-returns *before* the mode factor, so it deliberately skips `run_mode_factors`. I mirrored the non-MoE behaviour only, rather than guess at that asymmetry.

529 tests pass, `cargo fmt --check` clean, clippy unchanged at 23 pre-existing.

AlexsJones (approved):
This closes the plan/fit gap properly. Threading `CalcConfig` through the estimator means the TUI's Advanced Config tuning now lands in `plan` the way it already did in `fit`, and the CpuOnly factor fix was a real divergence (plan was `1/cpu_only` faster than fit even at defaults). The tests asserting plan and fit agree to within 1e-9 under default and tuned configs, including the CpuOnly path, pin exactly the invariant at stake here. Keeping `estimate_model_plan` as a defaulting wrapper preserves the public API. Full matrix green. Approving.

PR #842, opened 2026-08-09
fix: resolve generic AMD/ATI Strix Halo identity

hogeheer499-commits:
## What changed

Treat the `AMD/ATI` string returned by `lspci` as a generic AMD GPU identity.

On a Ryzen AI MAX+ 395 / Radeon 8060S system, `lspci -nnD` currently reports the display controller as:

```text
Advanced Micro Devices, Inc. [AMD/ATI] Device [1002:1586]
```

`extract_model_from_lspci_line()` therefore produces `AMD/ATI`. Because that value was not considered generic, the existing Ryzen AI MAX+ fallback never replaced it with the CPU-derived Strix Halo identity.

## User impact

Before this change, the v1.1.8 binary reported:

```text
gpu_name: AMD/ATI
memory_bandwidth_gbps: null
```

After building this branch on the same retail Beelink GTR9 Pro, `llmfit --json system` reports:

```text
gpu_name: AMD RYZEN AI MAX+ 395 w/ Radeon 8060S (integrated)
memory_bandwidth_gbps: 256.0
unified_memory: true
gpu_vram_gb: 124.94
```

This reuses llmfit's existing Strix Halo fallback and bandwidth table; it only makes the fallback reachable for this real `lspci` output. The test adds the missing `AMD/ATI` case to the generic-name regression coverage.

The reproduction machine is also the primary system used for the public [AMD Strix Halo Local LLM Guide](https://github.com/hogeheer499-commits/strix-halo-guide).

## Validation

- `cargo test -p llmfit-core` (526 passed, 1 ignored; integration tests also passed)
- release build and live `./target/release/llmfit --json system` verification on Ryzen AI MAX+ 395 / Radeon 8060S

`make clippy` currently reaches unrelated pre-existing `-D warnings` failures on `main`; this change introduces no new clippy finding.

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -2432,7 +2432,7 @@ fn is_generic_amd_gpu_name(name: &str) -> bool {
     let lower = name.trim().to_lowercase();
     matches!(
         lower.as_str(),
-        "amd gpu" | "radeon graphics" | "amd radeon graphics"
+        "amd gpu" | "amd/ati" | "radeon graphics" | "amd radeon graphics"
     )
 }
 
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Right fix, and thanks for adding the test assertion. Strix Halo APUs report as `AMD/ATI` on Windows through WMI, which is a valid AMD iGPU identity — the generic-name filter just needed one more entry. The test pins all four generic names including the new one.

Build and test workflows haven't run yet since this is a first-time contributor PR. Once a maintainer approves the workflow run and CI comes back green, this is good to merge.

PR #860, opened 2026-08-10
fix(providers): resolve MLX community tags to catalog ids

### LINKED ISSUE #854: Community benchmark lookup never matches MLX submissions
#853 added the first MLX entries to the community table (apple-m4-pro, 32 results). The review thread flagged that these entries are invisible to catalog throughput lookup: `MeasuredTps::lookup()` resolves the submission `model` field through `tag_matches_model()`, which only tries the generic and llama.cpp paths. MLX basenames such as `Llama-3.2-1B-Instruct-4bit` never reduce to a catalog id, so every MLX measurement is silently skipped in favor of formula estimates.

The scope is wider than the community index, since two more consumers share the same matching path:

- `LocalBenchIndex::lookup()` (share.rs): locally stored MLX runs never annotate the catalog either
- `bench_target_matches()` (tui_app.rs): the bench-offer flow cannot match a model served by a running MLX server

Root cause is narrower than a missing feature: the MLX matching machinery already exists but is not wired into `tag_matches_model()`. `strip_trailing_quant_suffix()` strips MLX quant suffixes, `hf_name_to_mlx_candidates()` generates mlx-community candidates including `-4bit`/`-8bit` and `-mlx-` infix forms, and `is_model_installed_mlx()` matches an HF name against them. None of the three is reachable from the lookup path.

Two coverage gaps in the existing helper would also need closing for the current corpus:

- `strip_trailing_quant_suffix()` only knows `-4bit`, `-6bit`, `-8bit`. mlx-community also publishes `-2bit`, `-3bit`, DWQ variants (`-4bit-DWQ`, date-stamped `-4bit-DWQ-05082025`) and `-fp16`. A pattern like `-\d+bit(?:-[a-z0-9]+)*$` covers the bit-width family in one shot.
- `gpt-oss-20b-MXFP4-Q4` (present in the #853 data) matches neither the bit-width pattern nor the candidate list; compound schemes need explicit handling.

Proposed fix, pending your direction:

1. extend `strip_trailing_quant_suffix()` as above (its other caller, `normalize_mlx_repo_base()`, benefits on the pull/install path too)
2. add the MLX-normalized stem to the exact-match stem set in `tag_matches_model()`. This keeps the deliberate exact-stem design documented in `bench_target_matches()` (no family over-matching through the Ollama heuristics)
3. tests at the `tag_matches_model()` level using real submission tags from #853 (`Llama-3.2-1B-Instruct-4bit`, `gpt-oss-20b-MXFP4-Q4`)

One design point to settle: once MLX entries match, `CommunityBenchIndex::lookup()` will take its median across llama.cpp and MLX entries for the same model, since the index ignores the per-result `provider` field. Both are real measurements on identical hardware, so this may be exactly what you want, but it changes the number reported on machines that have submissions from both runtimes.

No behavior change on the GGUF path. Happy to submit the PR either way you prefer, this wiring or something smaller. Related: #833 for the GGUF equivalent of this normalization.

Akciali:
Fixes #854.

`tag_matches_model()` only reduced GGUF stems to catalog slugs, so MLX submission tags (`Llama-3.2-1B-Instruct-4bit`) never matched and every MLX measurement was skipped by `CommunityBenchIndex::lookup()`, `LocalBenchIndex::lookup()` and the TUI bench-offer matching, which all resolve through this function.

Following the direction given in #854: a third match arm strips the MLX quant tail and feeds the remainder through the existing exact-stem path. Purely additive, no change to GGUF behavior, no refactor of the existing helpers.

- `strip_mlx_quant_suffix()`: end-anchored strip of `-<N>bit` with optional trailing variant markers (`-4bit-dwq`, date-stamped `-4bit-dwq-05082025`) via `-\d+bit(?:-[a-z0-9]+)*$`, plus the compound schemes `-mxfp4-q4`, `-mxfp4` and `-fp16` stripped as whole units. End-anchoring keeps mid-name fragments safe; the GGUF list keeps its `rfind` semantics for multi-part stems.
- Third arm in `tag_matches_model()` adds the MLX-stripped base to the exact-match stem set. The deliberate exact-stem design is preserved: no bare stem enters the Ollama candidate heuristics, so no family over-matching.
- Tests: suffix fixtures per pattern family with negative mid-name cases, and `tag_matches_model()` cases using verbatim tags from the 32 MLX rows of #853 (`Llama-3.2-1B-Instruct-4bit`, `Qwen3-14B-4bit`, `gpt-oss-20b-MXFP4-Q4`) plus a cross-model negative.

`cargo test -p llmfit-core --lib`: 538 passed, 0 failed. `cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features` clean on the touched lines.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -2,8 +2,10 @@
 //!
 //! Each provider can list locally installed models and pull new ones.
 
+use regex::Regex;
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
 
 // ---------------------------------------------------------------------------
 // Provider trait
@@ -3136,6 +3138,34 @@ pub fn strip_gguf_quant_suffix(stem: &str) -> Option<String> {
     None
 }
 
+/// Strip an MLX quantization suffix from a lowercased model stem, so
+/// mlx-community basenames reduce to catalog slugs (#854).
+/// "llama-3.2-1b-instruct-4bit" → "llama-3.2-1b-instruct"
+///
+/// End-anchored, unlike the GGUF list above: mlx-community always places the
+/// quant scheme last, and dtype-like fragments can occur inside genuine model
+/// names. Covers `-<N>bit` with optional trailing variant markers
+/// (`-4bit-dwq`, date-stamped `-4bit-dwq-05082025`) and the compound schemes
+/// `-mxfp4-q4`, `-mxfp4` and `-fp16`, stripped as whole units.
+pub fn strip_mlx_quant_suffix(stem: &str) -> Option<String> {
+    for pat in ["-mxfp4-q4", "-mxfp4", "-fp16"] {
+        if let Some(base) = stem.strip_suffix(pat)
+            && !base.is_empty()
+        {
+            return Some(base.to_string());
+        }
+    }
+    static MLX_BIT_SUFFIX: OnceLock<Regex> = OnceLock::new();
+    let re = MLX_BIT_SUFFIX
+        .get_or_init(|| Regex::new(r"-\d+bit(?:-[a-z0-9]+)*$").expect("valid MLX suffix regex"));
+    if let Some(m) = re.find(stem)
+        && m.start() > 0
+    {
+        return Some(stem[..m.start()].to_string());
+    }
+    None
+}
+
 // ---------------------------------------------------------------------------
 // llama.cpp name-matching helpers
 // ---------------------------------------------------------------------------
@@ -3965,6 +3995,12 @@ pub fn tag_matches_model(tag: &str, hf_name: &str) -> bool {
         .to_string();
     let mut stem_set = HashSet::new();
     if let Some(base) = strip_gguf_quant_suffix(&stem) {
+        stem_set.insert(base);
+    }
+    // Third arm: MLX community tags carry mlx-community basenames
+    // (`llama-3.2-1b-instruct-4bit`); strip the quant tail so they reduce to
+    // catalog slugs through the same exact-stem path as GGUF stems (#854).
+    if let Some(base) = strip_mlx_quant_suffix(&stem) {
         stem_set.insert(base);
     }
     stem_set.insert(stem);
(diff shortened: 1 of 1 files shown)


Akciali:
Verified against the real corpus: 31 of the 32 MLX tags from #853 now resolve to their catalog ids through `tag_matches_model` (checked against the embedded `hf_models.json`). The remaining one, `Codestral-22B-v0.1-4bit`, cannot resolve because the base model is absent from the embedded catalog (only an abliterated variant and an AWQ repack exist there); the GGUF twin row from #812 has the same limitation today. Catalog gap, out of scope here.

AlexsJones (approved):
Great turnaround on this, and thanks for scoping it exactly to the third match arm rather than reworking the MLX helpers.

Read the whole diff. The strip is end-anchored, the compound schemes are tried as whole units before the `-\d+bit(?:-[a-z0-9]+)*$` regex so `-mxfp4-q4` can't be half-eaten, and the `m.start() > 0` guard keeps a bare `4bit` from collapsing to an empty stem. Because the arm lands in `tag_matches_model()`, all three consumers named in #854 (`CommunityBenchIndex::lookup()`, `LocalBenchIndex::lookup()`, the TUI bench-offer path) pick it up for free. Tests cover the verbatim tags from the #853 corpus plus the cross-model negative, which is the case I'd have asked for. All green on the 3-OS matrix.

Approving for Alex's final look.

One follow-up worth an issue, not a change here: `strip_trailing_quant_suffix()` (used by `normalize_mlx_repo_base()` on the installed-detection path) still only knows `-4bit`/`-6bit`/`-8bit`, so DWQ and MXFP4 repos normalize there but not here. Two MLX quant strippers with different coverage will drift. Collapsing that one onto `strip_mlx_quant_suffix()` is the obvious cleanup once this lands.

Akciali:
Thanks for the read, and for the update on #854. I will file the `strip_trailing_quant_suffix()` follow-up as an issue and take the PR once this one lands, so the two strippers stop drifting. They already disagree in both directions, including one case where the older function is the correct one, so it is worth doing properly rather than as a patch on top of this branch.

Before that, one thing I owe you on this PR. I re-checked the compound schemes against the live mlx-community corpus instead of the #853 corpus they were derived from, and the whole-unit list is short by two forms:

| Tag | `strip_mlx_quant_suffix` today | HF downloads |
| --- | --- | --- |
| `gpt-oss-20b-MXFP4-Q8` | `None` | 324k |
| `gpt-oss-120b-MXFP4-Q8` | `None` | 4.5k |
| `GLM-5.1-MXFP4-Q8` | `None` | 2.8k |
| `gpt-oss-120b-mxfp4-bf16` | `None` | 2.0k |

`-mxfp4-q4` was the only compound form in the 32 rows of #853, so `-mxfp4-q8` and `-mxfp4-bf16` never showed up while I was writing the fixtures. `gpt-oss-20b-MXFP4-Q8` is the most downloaded MXFP4 repo in mlx-community, 324k against 2.3k for the Q4 variant I benchmarked, so a submission carrying it would still be skipped by the lookup. The GGUF list does not catch the bf16 one either: its `rfind` on `-bf16` reduces `gpt-oss-120b-mxfp4-bf16` to `gpt-oss-120b-mxfp4`, which is not the catalog slug.

The fix is two entries in the same whole-unit array plus the matching test cases, nothing structural, and it touches neither the regex nor the match arm you read. Since this one is queued for merge I am not pushing anything onto the branch: it goes into the follow-up PR along with the stripper collapse, and this merges as reviewed. Flagging it now so it is not merged on the assumption that the compound list was complete.

PR #863, opened 2026-08-11
fix(ollama): stop one sized install marking a whole model family installed

AlexsJones:
Fixes the over-detection reported in discussion #861: three models installed, 30+ shown as installed and therefore undownloadable.

## Root cause

Two halves met in the middle, both in `llmfit-core/src/providers.rs`:

1. `build_installed_set` inserted the full tag **and** the bare family stem — `qwen3:8b` → `{"qwen3:8b", "qwen3"}`.
2. The heuristic fallback in `hf_name_to_ollama_candidates` also offered the bare family name as a candidate for any model with no `OLLAMA_MAPPINGS` entry.

Bare stem in the set, bare stem in the candidates ⇒ one install marks the entire family. Measured over the embedded catalog (9,250 entries), with the installed set built exactly as `build_installed_set` builds it:

| Ollama install | matched before | matched after |
|---|---|---|
| `qwen3:8b` | **238** | 2 (`Qwen3-8B`, `Qwen3-8B-Base`) |
| `llama3.2:3b` | 8 | 2 (`Llama-3.2-3B`, `-3B-Instruct`) |
| `gemma3:4b` | 1 (a 27B fine-tune) | 0 |

Models *with* an explicit mapping resolve to a single sized tag and were already fine — the damage was concentrated in the long tail of the catalog. Since `installed` gates the download action, this fails in the worst direction: it silently removes models from reach.

## Fix — size is the discriminator

- A **sized** install contributes its tag and nothing else; the tag already says which weights are on disk.
- Only an untagged / `:latest` install contributes a family stem, plus the sized alias implied by the parameter count Ollama already reports in `/api/tags` (`qwen3:latest` → `"8.2B"` → `qwen3:8b`), so it still resolves to one model rather than a family. Truncated, not rounded — Ollama tags follow the marketing size (`qwen2.5:14b` reports `"14.8B"`).
- Candidates derived from a sized HF name no longer include the bare family.
- Size-less candidates — `OLLAMA_MAPPINGS` tags like `phi-4` → `phi4` and `qwq-32b` → `qwq`, or HF names with no size to parse — now match any tag of that family, which is what keeps `phi4:14b` detecting `microsoft/phi-4`.

The bare-family candidate arrived in #492 (fixing #481, "installed flag always false for models without explicit mapping"). This keeps that fix and removes its blast radius. #619/#650 was the same over-match seen through cloud models.

## Tests

- `sized_install_does_not_mark_the_family_installed` — the reported scenario: `qwen3:8b` detects `Qwen3-8B` and leaves `0.6B`/`4B`/`32B`/`235B-A22B` alone.
- `latest_install_resolves_to_its_parameter_size` and `latest_install_without_a_parameter_size_stays_family_level`.
- `sizeless_mapping_still_matches_a_sized_install` — `phi4:14b` → `microsoft/phi-4`.
- `every_mapped_model_is_still_detected_from_its_own_tag` — walks all 134 `OLLAMA_MAPPINGS` entries so narrowing detection can't silently cost a model.
- `parameter_size_truncates_to_the_marketing_tag`.

`cargo test --workspace` green (543 core + 80 tui + integration); `cargo fmt` clean; no new clippy warnings (20 before, 20 after on `llmfit-core --all-targets`).

## Not covered here

`google/gemma-3-4b-it` still isn't detected from `gemma3:4b` — `OLLAMA_MAPPINGS` has `gemma-3-12b-it` but not the 1B/4B/27B siblings, and the heuristic yields `gemma-3:4b` against Ollama's `gemma3:4b`. Pre-existing false negative, unchanged by this PR, worth a follow-up on the mapping table.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -290,15 +290,29 @@ impl OllamaModel {
 /// The tag Ollama resolves when a model is pulled without one.
 const OLLAMA_DEFAULT_TAG: &str = "latest";
 
-/// Ollama-style size token for the parameter count Ollama reports, e.g.
-/// "8.2B" → `8b`. Truncates rather than rounds because Ollama tags follow the
-/// marketing size, not the true count (`qwen2.5:14b` reports "14.8B"). Sizes
-/// below 1B are reported in "M" and have no reliable tag form, so they yield
-/// `None` rather than a bogus `0b`.
-fn size_token_from_parameter_size(parameter_size: &str) -> Option<String> {
+/// Ollama-style size tokens implied by the parameter count Ollama reports,
+/// e.g. "8.2B" → `["8b", "8.2b"]`.
+///
+/// Most tags carry the marketing size rather than the true count (`qwen2.5:14b`
+/// reports "14.8B"), hence the truncated form. Families tagged with a decimal
+/// (`qwen3:1.7b`, `solar:10.7b`) need the verbatim form as well. Counts below
+/// 1B are reported in "M" — `qwen3:0.6b` reports "596.05M" — and have no
+/// reliable tag form, so they yield nothing rather than a bogus `0b`.
+fn size_tokens_from_parameter_size(parameter_size: &str) -> Vec<String> {
     let raw = parameter_size.trim().to_lowercase();
-    let value: f64 = raw.strip_suffix('b')?.parse().ok()?;
-    (value >= 1.0).then(|| format!("{}b", value.trunc() as u64))
+    let Some(value) = raw
+        .strip_suffix('b')
+        .and_then(|digits| digits.parse::<f64>().ok())
+        .filter(|v| *v >= 1.0)
+    else {
+        return Vec::new();
+    };
+
+    let mut tokens = vec![format!("{}b", value.trunc() as u64)];
+    if tokens[0] != raw {
+        tokens.push(raw);
+    }
+    tokens
 }
 
 /// Build the set of installed model name stems from Ollama's tag list, plus the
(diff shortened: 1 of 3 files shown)


AlexsJones:
## Update — CI failure and review feedback

**The failing Test Suite job was not this branch.** All Rust tests passed on all three platforms (543 core + 80 tui); the job died afterwards in the Python step:

```
ValueError: Readme path must be within the project directory: ../README.md
```

`llmfit-python/pyproject.toml` declares `readme = "../README.md"`, which current hatchling refuses, so `uv sync` fails. Main is red for the same reason ([CI on 9d240c3's base](https://github.com/AlexsJones/llmfit/actions)). Fixed in `83d6f68` by supplying the README *contents* through the custom metadata hook that already sets `version` and `license-expression` — hatchling constrains readme paths but not readme text. No duplicated or symlinked README (a symlink checks out as a text stub on Windows runners), and the built metadata is unchanged in substance: `Description-Content-Type: text/markdown` plus the full README body. Verified locally with `uv sync`, `uv run pytest`, `ruff check`, `ruff format --check`, `ty check`.

Happy to split that commit into its own PR if you'd rather land it separately — it unblocks every open PR, not just this one.

**Greptile's decimal-size point was right**, fixed in `64eb027`. `qwen2.5:14b` reports `"14.8B"`, so a `:latest` install needs the truncated marketing size — but 20 `OLLAMA_MAPPINGS` entries are tagged with the decimal itself (`solar:10.7b`, `qwen3:1.7b`, `lfm2:1.2b`), and for those the truncated alias alone reports the model absent. Both forms are now emitted. They stay inside the family that is genuinely installed, so the extra alias can't resurrect the cross-family match this PR removes — one of the two is just dead weight per install. Covered by `latest_install_of_a_decimal_tagged_family_is_detected` and an extended token test.

Residual limitation, called out honestly: a `:latest` install whose reported count doesn't line up with its tag (Ollama reports `"2.0B"` for `qwen3:1.7b`) still won't be detected. That is a false negative — the model is simply not marked installed — where the behaviour it replaces was a false positive across the whole family. Pulling the exact tag (`ollama pull qwen3:1.7b`) is unaffected: the tag itself is in the set.

Issue #862, opened 2026-08-11
Questions about sharing benchmarks


datawookie:
### Bug description

I'm trying to share a benchmark.

```
$ llmfit bench --share 

  The following stored benchmark data would be contributed:

    - gemma3:4b via ollama — 7.9 tok/s
    - gemma3:4b via ollama — 8.9 tok/s
    - qwen3.5 via llamacpp — 1.4 tok/s
    - muse-glimmer via llamacpp — 1.4 tok/s

  Share all 4 local benchmark submission(s) as a PR to AlexsJones/llmfit? [y/N] y
  GitHub: authenticated as datawookie
  Share failed: GitHub API returned 404: Not Found
  Your results remain stored locally; retry with `llmfit bench --share`.
```

Two things I wanted to raise:

1. The 404 error.
2. Is is possible to just share one benchmark? For example, right now I only want to upload the `muse-glimmer` result.

### Expected behavior

1. No 404 error.
2. Ability to select from local benchmarks to upload.

### Actual behavior

1. 404 error.
2. Looks like all with be uploaded.

### Steps to reproduce

llmfit bench --share

### Diagnostic report (`llmfit doctor`)

```text
The following stored benchmark data would be contributed:

    - gemma3:4b via ollama — 7.9 tok/s
    - gemma3:4b via ollama — 8.9 tok/s
    - qwen3.5 via llamacpp — 1.4 tok/s
    - muse-glimmer via llamacpp — 1.4 tok/s

  Share all 4 local benchmark submission(s) as a PR to AlexsJones/llmfit? [y/N] y
  GitHub: authenticated as datawookie
  Share failed: GitHub API returned 404: Not Found
  Your results remain stored locally; retry with `llmfit bench --share`.
```

### llmfit version

llmfit 1.16

### Operating system

Linux

### Affected component

CLI output (subcommands, table display)

### Mode used

CLI (--cli flag)

### GPU details (if relevant)

```text

```

AlexsJones:
Thanks for the clean paste of the whole flow, that made this quick to pin down.

[...]

PR #865, opened 2026-08-11
fix(share): explain a refused write to the user's own fork

AlexsJones:
Refs #862 (the 404 half of it).

@datawookie hit `Share failed: GitHub API returned 404: Not Found` right after `--share` printed "authenticated as datawookie". The reason that reads as a contradiction: GitHub answers 404, not 403, when a token may read a repository but has no write access to it, so the failure at `POST /repos/<login>/llmfit/git/refs` looks like a missing fork.

The gap that lets it get that far is in `preflight_auth()`: `GITHUB_TOKEN` / `GH_TOKEN` short-circuits the device flow, and the only check on it is `GET /user`, which any valid token passes. A read-only classic token or a fine-grained PAT without Contents: write therefore sails through preflight and dies at the first write.

This PR does the small half: on 403/404 from fork creation, branch creation, or the contents write, say which repo could not be written and name the token as the likely cause. Every other status keeps the existing message.

```
GitHub API returned 404: Not Found
  The token is authenticated as datawookie but cannot write to datawookie/llmfit.
  If GITHUB_TOKEN or GH_TOKEN is set, llmfit uses it as-is: unset it and retry to log in
  via the device flow (which requests public_repo), or give that token write access to
  your fork.
```

Tests cover the 404 and 403 shapes and assert that a 422 is left alone, since the hint would only mislead there.

Two things deliberately left out, both worth their own issue:

- reading `X-OAuth-Scopes` from `GET /user` in preflight so a scope-poor classic token fails before the benchmarks run. That needs `api()` to surface response headers, which is a wider change than this.
- the second ask in #862, selecting which stored benchmarks to share instead of all of them.

Not closing #862, since the selection question is still open there.

### DIFF
--- a/llmfit-core/src/share.rs
+++ b/llmfit-core/src/share.rs
@@ -1018,6 +1018,28 @@ fn api_error(status: u16, body: &Value) -> String {
     format!("GitHub API returned {status}: {msg}")
 }
 
+/// Error for a write to the user's own fork that GitHub refused.
+///
+/// GitHub answers 404 (not 403) when a token may read a repository but has no
+/// write access to it, so a bare `api_error` here reads as "your fork does not
+/// exist" when the real cause is almost always token scope: `GITHUB_TOKEN` /
+/// `GH_TOKEN` short-circuits the device flow and is only checked against
+/// `GET /user`, which any valid token passes.
+fn write_denied_error(status: u16, body: &Value, login: &str) -> String {
+    let base = api_error(status, body);
+    if status != 404 && status != 403 {
+        return base;
+    }
+    format!(
+        "{base}\n  \
+         The token is authenticated as {login} but cannot write to \
+         {login}/{UPSTREAM_REPO}.\n  \
+         If GITHUB_TOKEN or GH_TOKEN is set, llmfit uses it as-is: unset it and \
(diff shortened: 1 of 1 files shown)


datawookie:
I can confirm that this works:

```
$ ./target/release/llmfit bench --share

  The following stored benchmark data would be contributed:

    - gemma3:4b via ollama — 7.9 tok/s
    - gemma3:4b via ollama — 8.9 tok/s
    - qwen3.5 via llamacpp — 1.4 tok/s
    - muse-glimmer via llamacpp — 1.4 tok/s

  Share all 4 local benchmark submission(s) as a PR to AlexsJones/llmfit? [y/N] y
  GitHub: authenticated as datawookie
  Share failed: GitHub API returned 404: Not Found
  The token is authenticated as datawookie but cannot write to datawookie/llmfit.
  If GITHUB_TOKEN or GH_TOKEN is set, llmfit uses it as-is: unset it and retry to log in via the device flow (which requests public_repo), or give that token write access to your fork.
  Your results remain stored locally; retry with `llmfit bench --share`.
```

⛔ I didn't have a token in my environment though. Still no luck.

✅ I deleted my fork of the project and the share worked perfectly.

Issue #866, opened 2026-08-11
[Bug]: gemma3:4b does not resolve to google/gemma-3-4b-it in installed detection


AlexsJones:
Follow-up split out of [discussion #861](https://github.com/AlexsJones/llmfit/discussions/861), where it was named as the one gap #863 does not close.

#863 made size the discriminator for Ollama installed-detection, so `qwen3:8b` no longer marks 238 models installed. What it does not fix is the opposite failure: `gemma3:4b` still does not resolve to `google/gemma-3-4b-it`, because there is no entry for it in the mapping table. An install that is genuinely present therefore shows as not installed, and the user gets offered a download they already have.

Concretely:

- installed via `ollama pull gemma3:4b`
- expected: `google/gemma-3-4b-it` marked installed
- actual: not matched, no catalog entry claims it

Worth checking the rest of the gemma3 sizes (`:1b`, `:12b`, `:27b`) against their `google/gemma-3-*-it` catalog ids while the table is open, since the gap is likely to be the whole family rather than the 4B alone.

Akciali:
Repro on `main` (b79f2df), with the installed set built exactly as `build_installed_set` builds it:

| Ollama install | catalog id            | detected | candidates generated              |
| -------------- | --------------------- | -------- | --------------------------------- |
| `gemma3:1b`    | `google/gemma-3-1b-it`  | no     | `gemma-3:1b`, `gemma-3-1b`, ...   |
| `gemma3:4b`    | `google/gemma-3-4b-it`  | no     | `gemma-3:4b`, `gemma-3-4b`, ...   |
| `gemma3:12b`   | `google/gemma-3-12b-it` | yes    | `gemma3:12b` (explicit mapping)   |
| `gemma3:27b`   | `google/gemma-3-27b-it` | no     | `gemma-3:27b`, `gemma-3-27b`, ... |

So the whole family except the 12B, as you suspected.

## Root cause

Ollama glues the version number to the family name, HuggingFace separates it with a hyphen: `gemma3` against `gemma-3`. `hf_name_to_ollama_candidates` keeps the HF spelling when it falls back to the heuristic, so the candidate it emits cannot equal a real Ollama tag by construction. The missing table entries are what expose it, they are not the whole of it.

This is not gemma specific. Over the embedded catalog (10,217 entries after dedupe), 1,535 entries produce a sized candidate whose stem carries a hyphenated version, and none of them can ever match.

## What a generic fix would reach

[...]

PR #873, opened 2026-08-13
fix(providers): match every K-quant and I-quant GGUF variant

AlexsJones:
### Problem

`strip_gguf_quant_suffix` (`llmfit-core/src/providers.rs`) enumerated published
quant variants one at a time:

```rust
"-q8_0", "-q6_k", "-q6_k_l", "-q5_k_m", "-q5_k_s", "-q4_k_m", "-q4_k_s", ...
```

Anything not on that list never reduces to the catalog id, so
`is_model_installed_llamacpp` never reaches the repo name and the file reads as
**neither installed nor served**. The gaps were not exotic — they were whole
publishers:

- bartowski `_L` quants: `-q3_k_l`, `-q4_k_l`, `-q5_k_l` (only `-q6_k_l` was listed)
- Unsloth Dynamic `_XL` quants: `-q4_k_xl`, `-q5_k_xl`, `-q8_k_xl`
- the IQ family beyond four hand-picked entries: `-iq4_nl`, `-iq3_xxs`, …

`Q3_K_L` in particular is a stock llama.cpp quant, not a vendor variant.

### Fix

Match on the quant *family* stem — `-qN_k` for K-quants, `-iqN_` for I-quants —
so `_S`/`_M`/`_L`/`_XL` and `_NL`/`_XS`/`_XXS` all reduce identically and the
next published variant needs no code change. `-q6_k` and `-q2_k` already worked
this way; the rest now follow. The existing `-ud` marker strip is unchanged and
still runs after a quant hit.

### Why now

Found while reviewing #872. Three of its 122 GB10 entries are silently dropped
during lookup on this bug:

| Tag | tok/s |
|---|---|
| `Step-3.7-Flash-UD-IQ4_NL.gguf` | 21.41 |
| `Qwen-AgentWorld-35B-A3B-UD-Q4_K_XL.gguf` | 64.17 |
| `LFM2.5-8B-A1B-UD-Q4_K_XL.gguf` | 151.04 |

Greptile flagged the `IQ4_NL` one and the submitter renamed the tag to add a
`.gguf` extension — but `tag_matches_model` trims `.gguf` *before* stripping the
quant, so that rename is a no-op for matching. The data was never the problem;
**#872 can land its files unmodified once this merges.**

Scanning the whole `data/community` tree, one already-merged entry
(`hf.co/unsloth/Qwen3.6-27B-MTP-GGUF:UD-Q4_K_XL`) is also recovered. The
remaining unmatched tags there are MLX `-4bit` and AWQ, handled by
`strip_mlx_quant_suffix` on a separate path.

The blast radius is wider than community data: this function also backs local
install detection, so anyone with an Unsloth UD-XL or bartowski `_L` GGUF on
disk was being told the model was not installed.

### Tests

Two new cases in `providers::tests` — one over every K/I variant (with and
without the `-ud` marker), one asserting `tag_matches_model` now resolves the
three GB10 tags above. Full suite green: 549 + 6 + 1 passing, `cargo fmt
--check` clean, no new clippy warnings.

### Not in scope

`select_best_gguf`'s `quant_order` has the same enumeration shape, so a repo
publishing only `UD-Q4_K_XL` falls through to its "largest that fits" branch
rather than preferring by quality. That still yields a usable file, so I left
it for a separate change.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -3172,10 +3172,16 @@ fn docker_mr_installed_matches(installed_name: &str, candidate: &str) -> bool {
 /// Strip quantization suffix from a GGUF file stem.
 /// "llama-3.1-8b-instruct-q4_k_m" → "llama-3.1-8b-instruct"
 pub fn strip_gguf_quant_suffix(stem: &str) -> Option<String> {
+    // Matched on the quant *family* stem rather than each published variant:
+    // K-quants as `-qN_k` and I-quants as `-iqN_`, so `_S`/`_M`/`_L`/`_XL`
+    // and `_NL`/`_XS`/`_XXS` all reduce to the same base. Enumerating the
+    // variants instead left whole publishers unmatched — bartowski `_L` files
+    // and Unsloth Dynamic `_XL`/`IQ4_NL` files never reached the catalog id,
+    // so they read as neither installed nor served.
     let quant_patterns = [
-        "-q8_0", "-q6_k", "-q6_k_l", "-q5_k_m", "-q5_k_s", "-q4_k_m", "-q4_k_s", "-q4_0",
-        "-q3_k_m", "-q3_k_s", "-q2_k", "-iq4_xs", "-iq3_m", "-iq2_m", "-iq1_m", "-f16", "-f32",
-        "-bf16", ".q8_0", ".q6_k", ".q5_k_m", ".q4_k_m", ".q4_0", ".q3_k_m", ".q2_k",
+        "-q8_0", "-q8_k", "-q6_k", "-q5_k", "-q4_k", "-q4_0", "-q3_k", "-q2_k", "-iq4_", "-iq3_",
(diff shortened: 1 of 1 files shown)


Akciali:
Confirming the fix from your side. I reproduced the resolution chain in isolation against hf_models.json before this PR, and the three GB10 tags you list are exactly the ones that failed, so the family-stem approach resolves them.

One adjacent case on the same path, in case it belongs here or in a follow-up: the -qat marker behaves like -ud did. gemma-4-31B-it-QAT-Q4_0.gguf reduces to gemma-4-31b-it-qat once the quant is stripped, and the catalog holds google/gemma-4-31B-it with no -qat stem, so it stays unmatched. Five entries in #872 are in that shape: 12B, 26B-A4B, 31B, E2B and E4B, all published QAT-only by Google in Q4_0.

Arguable either way. QAT is a quantization recipe rather than a different model, so for throughput it is the same weights class, but Google does publish it as a separate variant.

Same failure mode as #866 on the Ollama side, different table: there the fallback cannot build the tag at all, since Ollama drops a separator the HF name keeps. Measured on this tree, that fallback reproduces only 35 percent of the 133 mappings, so the table is the sole safety net for the rest.

PR #886, opened 2026-08-16
feat: recognize native ternary (1.58-bit) models

BlueVelvetSackOfGoldPotatoes:
## What

Adds a native ternary (1.58-bit) quantization tier (`I2_S`, plus ggml `TQ1_0`/`TQ2_0`) and a `bitnet.cpp` runtime. BitNet-style models (BitNet-b1.58, Falcon3-*-1.58bit, Llama3-8B-1.58, EdgeRazor / Ternary-Bonsai, …) are now modeled as native ternary CPU models instead of generic `Q4_K_M` / llama.cpp GGUFs.

## Why

These models are already in the catalog but scored as ordinary 4-bit GGUFs. `llmfit info microsoft/bitnet-b1.58-2B-4T` reports `Quantization: Q4_K_M`, `Runtime: llama.cpp`, a GPU-default run mode and a GPU-bandwidth speed estimate — there is no i2_s tier and no bitnet.cpp runtime, even though the catalog already tags these `architecture: bitnet`.

## How

- `models::is_ternary_native()` detects them via the HF `architecture` field (`bitnet`) and repo-name conventions (`1.58bit`, `bitnet`, `ternary`, `b1_58`). `-bf16` / `unpacked` masters and Apple-MLX repacks are excluded.
- New `TERNARY_QUANT_HIERARCHY = ["I2_S"]`, mirroring the existing MLX / ONNX special-casing, so ternary models resolve to i2_s sizing + a `bitnet.cpp` runtime and get a CPU-first note.
- i2_s bytes/param is derived from released GGUFs (BitNet-2B-4T ~1.2 GB, Falcon3-10B-1.58bit ~4.0 GB).

## Testing

`cargo test -p llmfit-core` (all pass, +4 new), `cargo fmt --check`, and `cargo clippy --all-targets` are clean.

## Not in scope

- Some 1.58-bit repos report *packed* parameter counts, so absolute size for those stays off until `scrape_hf_models.py` is updated — separate concern.
- Ternary models still use the standard throughput estimator; a bitnet.cpp CPU-throughput calibration is a natural follow-up.


AlexsJones (changes requested):
Nice piece of work, and thanks for keeping the detection exclusions (bf16, unpacked, MLX repacks) explicit rather than hoping a name check was enough. The quant tier and the `is_ternary_native` tests are the right shape.

Three things to fix, all the same root cause: `InferenceRuntime::BitNet` is a new value in a vocabulary that has string tables elsewhere, and those did not move with it.

1. `llmfit-tui/src/main.rs` line 1419 (the `--force-runtime` parse) and line 1465 (the `--runtime` filter) still only know `mlx|llamacpp|vllm`. So there is no way to ask for or filter to bitnet.cpp, and more importantly `runtime=llamacpp` now silently drops every ternary model it used to return. That last part is a behaviour change for existing callers of the CLI and `/api/v1/models`.
2. `API.md` line 235 and `docs/cli.md` line 79 document that vocabulary as a contract, so `bitnetcpp` needs to appear in both.
3. `estimate_tps`: the Metal arm learned about BitNet but `(GpuBackend::Cuda, _) => 220.0` and `(GpuBackend::Rocm, _) => 180.0` did not, so a CPU-oriented runtime gets a GPU constant on exactly the machines where your new note says it runs best on CPU. Is the intent that a ternary model on a CUDA box still scores as a GPU run? If yes that is fine, it just wants saying somewhere.

Point 1 is the one I would not merge without. The rest is quick.

AlexsJones (changes requested):
Thanks for carrying the runtime vocabulary through the CLI, API, and docs. One blocker before the next review: the current head's Rustfmt check is failing. Please run `cargo fmt` and push the formatting result, then I’ll re-check the full ternary path.

AlexsJones (changes requested):
Thanks for fixing the CPU-memory path and forced-runtime mismatch. One catalogue case still needs a guard: `tiiuae/Falcon3-10B-Base-1.58bit-prequantized` declares GGUF Q4_K_M with architecture `llama`, but the name-only heuristic selects BitNet and I2_S anyway. Respect an explicitly incompatible artifact quant/format before falling back to the name, and add that catalogue entry as a regression fixture alongside a genuine I2_S model.

BlueVelvetSackOfGoldPotatoes:
Fixed in b284ed4.

On the approach: the catalogue sets `quantization` to "Q4_K_M" for every entry, including the i2_s models (`microsoft/bitnet-b1.58-2B-4T`, `1bitLLM/bitnet_b1_58-3B`, and others). A guard on the quant field would drop those as well, so it cannot separate a repack from a native model in this data.

The name is the only per-entry signal here, so I extended the existing exclusion list (`bf16` / `unpacked` / `-mlx`) to also drop `-prequantized` and `mlx-community`. That covers `tiiuae/Falcon3-10B-Base-1.58bit-prequantized` and `mlx-community/Falcon3-7B-Instruct-1.58bit-4bit`; both now resolve to llama.cpp, while `microsoft/bitnet-b1.58-2B-4T` stays bitnet.cpp.

Added `tiiuae/Falcon3-10B-Base-1.58bit-prequantized` as a regression test loaded from the embedded catalogue, checked against `microsoft/bitnet-b1.58-2B-4T`, plus unit negatives for both names.

AlexsJones (approved):
Thanks for the `is_ternary_native` guard and the catalogue regression. The `prequantized`/`mlx`/`bf16`/`unpacked` exclusions mean a Q4_K_M repack no longer gets claimed as i2_s, which is exactly the case I raised. Ready for your final look.

BlueVelvetSackOfGoldPotatoes:
Looks good from my side. Ready to merge whenever you are.

PR #895, opened 2026-08-17
fix(providers): unify the two MLX quant-suffix strippers

### LINKED ISSUE #869: MLX quant suffix stripping lives in two functions with divergent coverage
Follow-up to the review of #860, filed at the maintainer's request.

Once #860 lands, MLX quant suffixes are stripped by two independent code paths in `llmfit-core/src/providers.rs`:

| Path | Entry point | Consumers |
| --- | --- | --- |
| Community/local lookup | `strip_mlx_quant_suffix()` (added by #860), called from `tag_matches_model()` | `CommunityBenchIndex::lookup()`, `LocalBenchIndex::lookup()`, `bench_target_matches()` in the TUI |
| Installed detection | `strip_trailing_quant_suffix()`, called from `normalize_mlx_repo_base()` | `hf_name_to_mlx_candidates()`, `is_model_installed_mlx()`, `mlx_pull_tag()` |

Both take the same input shape, a lowercased mlx-community basename. Their coverage differs, and the gap runs in both directions: each one already knows a pattern the other one misses.

### Coverage matrix

Measured by running the real functions, not by reading them. Baseline `main` at `4801d4a`, plus a worktree of #860 at `95df67f`.

| Suffix | Real example (HF downloads) | `strip_mlx_quant_suffix` (#860) | `normalize_mlx_repo_base` + `strip_trailing_quant_suffix` |
| --- | --- | --- | --- |
| `-4bit`, `-6bit`, `-8bit` | `Llama-3.2-1B-Instruct-4bit` | yes | yes |
| `-2bit`, `-3bit`, `-5bit` | `mlx-community/MiniMax-M2.1-3bit` (2.4k) | yes | no |
| `-4bit-dwq` | `mlx-community/Qwen3-30B-A3B-4bit-DWQ` (740) | yes | no |
| `-4bit-dwq-<date>` | `mlx-community/Qwen3-14B-4bit-DWQ-053125` (1.2k) | yes | no |
| `-mxfp4`, `-mxfp4-q4` | `mlx-community/GLM-5.2-mxfp4` (12k), `mlx-community/gpt-oss-20b-MXFP4-Q4` (2.3k) | yes | no |
| `-fp16` | `mlx-community/llama-3.3-70b-instruct-fp16` (2.6k) | yes | no |
| `-mxfp4-q8` | `mlx-community/gpt-oss-20b-MXFP4-Q8` (324k) | **no** | **no** |
| `-mxfp4-bf16` | `mlx-community/gpt-oss-120b-mxfp4-bf16` (2.0k) | **no** | **no** |
| `-mlx-` infix | `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-8bit` (124k) | **no**, leaves `...-instruct-mlx` | yes, `-mlx` stripped after the quant |

Download counts read from the HF API on 2026-08-12.

The last row is the one that argues hardest for a single implementation: the older function is the one that gets `-mlx` right. So this is not a case of one stripper being strictly ahead of the other, and picking either as-is loses coverage.

### Observable effect on the installed-detection path

With `mlx-community/Qwen3-30B-A3B-4bit-DWQ` present in the HF cache and the catalog entry `Qwen/Qwen3-30B-A3B` selected, on `main` today:

```
is_model_installed_mlx("Qwen/Qwen3-30B-A3B", {"mlx-community/qwen3-30b-a3b-4bit-dwq", "qwen3-30b-a3b-4bit-dwq"})
  = false
is_model_installed_mlx("Qwen/Qwen3-30B-A3B", {"mlx-community/qwen3-30b-a3b-4bit",     "qwen3-30b-a3b-4bit"})
  = true      // control
mlx_pull_tag("Qwen/Qwen3-30B-A3B") = "qwen3-30b-a3b-4bit"
```

So the MLX column reads empty for a model the user has on disk, and the TUI offers to pull a variant of something already downloaded.

The `-mlx` infix row cuts the other way, on the lookup path, with #860 applied:

```
is_model_installed_mlx("Qwen/Qwen3-Coder-30B-A3B-Instruct",
                       {"lmstudio-community/qwen3-coder-30b-a3b-instruct-mlx-8bit", ...})
  = true      // -mlx-8bit candidate generated by normalize_mlx_repo_base
tag_matches_model("Qwen3-Coder-30B-A3B-Instruct-MLX-8bit", "Qwen/Qwen3-Coder-30B-A3B-Instruct")
  = false     // same repo, other path
```

That repo has 124k downloads and is what LM Studio installs. Same for `mlx-community/bge-m3-mlx-fp16` (15.5k), which reduces to `bge-m3-mlx` and stops there.

The divergence also pollutes candidate generation when the model name is itself an MLX repo id, because the unstripped quant tail is carried into every generated candidate:

```
hf_name_to_mlx_candidates("mlx-community/Qwen3-8B-4bit-DWQ")
  = ["mlx-community/qwen3-8b-4bit-dwq", "qwen3-8b-4bit-dwq",
     "qwen3-8b-4bit-dwq-4bit", "qwen3-8b-4bit-dwq-8bit",
     "qwen3-8b-4bit-dwq-mlx-4bit", "qwen3-8b-4bit-dwq-mlx-8bit"]

hf_name_to_mlx_candidates("mlx-community/Qwen3-8B-8bit")     // control, quant recognized
  = ["mlx-community/qwen3-8b-8bit", "qwen3-8b-8bit", "qwen3-8b-4bit", "qwen3-8b"]
```

The last four candidates in the first list are repo names that cannot exist.

### Suggested direction

Collapse `strip_trailing_quant_suffix()` onto `strip_mlx_quant_suffix()` once #860 is merged, teach the surviving function the `-mlx` infix so nothing regresses on the installed-detection side, and close the two compound schemes neither one covers today (`-mxfp4-q8`, `-mxfp4-bf16`, flagged on #860).

`strip_mlx_quant_suffix` is already end-anchored and returns `Option<String>`, so `normalize_mlx_repo_base()` can call it with `unwrap_or_else(|| name.to_string())` and keep its current signature. Both functions already assume a lowercased input (`strip_mlx_quant_suffix` documents it, `normalize_mlx_repo_base` takes `repo_lower`), so the contract matches and no caller needs to change how it normalizes case.

Two callers, one stripper, one test table. Happy to send the PR once #860 lands, if you want it done this way.

Context: #854 originally sketched extending `strip_trailing_quant_suffix()` in place. The separate function came out of the third-arm direction, which kept #860 small and easy to review. This issue picks the consolidation back up now that both exist side by side.

Not related to #791: that one is about which repos get classified as MLX, this one is about normalizing the name once classification has happened.

### Scope note

This is about the two strippers only. Making a DWQ or MXFP4 repo count as installed for a plain catalog entry is a separate question, because that path generates candidates from the catalog name and currently emits only `-4bit` and `-8bit` variants. Collapsing the strippers does not by itself change that, and I would rather not widen this issue into it. Say the word if you want it covered here instead of in its own thread.

Akciali:
Closes #869. Follows the shape agreed there.

`strip_mlx_quant_suffix()` is now the single stripper: quant scheme first (`-<N>bit` and `-mxfp4`, both with the same permissive tail so `-4bit-dwq`, date-stamped DWQ, `-mxfp4-q8` and `-mxfp4-bf16` are covered, plus plain `-fp16`), then a trailing `-mlx` marker, then dash trimming. The `Option` return is kept so lookup callers can still tell "no suffix present" from "already reduced". `normalize_mlx_repo_base()` becomes a thin wrapper over it and `strip_trailing_quant_suffix()` is removed.

Observable changes, measured on this branch:

- `tag_matches_model("Qwen3-Coder-30B-A3B-Instruct-MLX-8bit", "Qwen/Qwen3-Coder-30B-A3B-Instruct")` now true, was false
- `-mxfp4-q8` / `-mxfp4-bf16` now strip on both paths, were `None` on both
- `hf_name_to_mlx_candidates("mlx-community/Qwen3-8B-4bit-DWQ")` no longer emits impossible repo names (`qwen3-8b-4bit-dwq-4bit`, ...) and now yields `qwen3-8b-4bit` / `qwen3-8b-8bit` / `qwen3-8b`

Per the scope note in #869, whether a DWQ or MXFP4 download counts as installed for a plain catalog entry is untouched: candidate generation still emits only `-4bit`/`-8bit` variants.

Fixtures are the coverage matrix from the issue, added to `test_strip_mlx_quant_suffix_patterns`, plus end-to-end rows in `test_tag_matches_model_mlx_community_tags` and a candidate-generation guard for DWQ repo ids. `cargo test -p llmfit-core -p llmfit`: 652 passed, 0 failed. fmt and clippy clean against baseline.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -3269,31 +3269,40 @@ pub fn strip_gguf_quant_suffix(stem: &str) -> Option<String> {
 }
 
 /// Strip an MLX quantization suffix from a lowercased model stem, so
-/// mlx-community basenames reduce to catalog slugs (#854).
+/// mlx-community basenames reduce to catalog slugs (#854, #869).
 /// "llama-3.2-1b-instruct-4bit" → "llama-3.2-1b-instruct"
 ///
 /// End-anchored, unlike the GGUF list above: mlx-community always places the
 /// quant scheme last, and dtype-like fragments can occur inside genuine model
-/// names. Covers `-<N>bit` with optional trailing variant markers
-/// (`-4bit-dwq`, date-stamped `-4bit-dwq-05082025`) and the compound schemes
-/// `-mxfp4-q4`, `-mxfp4` and `-fp16`, stripped as whole units.
+/// names. Strips, in order: the quant scheme (`-<N>bit` or `-mxfp4`, each
+/// with optional trailing variant markers such as `-4bit-dwq`, date-stamped
+/// `-4bit-dwq-05082025`, `-mxfp4-q8`, `-mxfp4-bf16`, or a plain `-fp16`),
+/// then a trailing `-mlx` marker (`...-instruct-mlx-8bit` → `...-instruct`).
+///
+/// Returns `None` when nothing was stripped, so lookup callers can tell "no
+/// MLX suffix present" apart from "already reduced".
 pub fn strip_mlx_quant_suffix(stem: &str) -> Option<String> {
-    for pat in ["-mxfp4-q4", "-mxfp4", "-fp16"] {
-        if let Some(base) = stem.strip_suffix(pat)
-            && !base.is_empty()
-        {
-            return Some(base.to_string());
-        }
+    static MLX_QUANT_SUFFIX: OnceLock<Regex> = OnceLock::new();
+    let re = MLX_QUANT_SUFFIX.get_or_init(|| {
+        Regex::new(r"-(?:\d+bit(?:-[a-z0-9]+)*|mxfp4(?:-[a-z0-9]+)*|fp16)$")
+            .expect("valid MLX suffix regex")
+    });
+    let without_quant = match re.find(stem) {
+        Some(m) if m.start() > 0 => &stem[..m.start()],
+        _ => stem,
+    };
+    let base = match without_quant.strip_suffix("-mlx") {
+        Some(b) if !b.is_empty() => b,
+        _ => without_quant,
+    };
+    if base.len() == stem.len() {
+        return None;
     }
-    static MLX_BIT_SUFFIX: OnceLock<Regex> = OnceLock::new();
-    let re = MLX_BIT_SUFFIX
-        .get_or_init(|| Regex::new(r"-\d+bit(?:-[a-z0-9]+)*$").expect("valid MLX suffix regex"));
-    if let Some(m) = re.find(stem)
-        && m.start() > 0
-    {
-        return Some(stem[..m.start()].to_string());
+    let base = base.trim_matches('-');
+    if base.is_empty() {
+        return None;
     }
-    None
+    Some(base.to_string())
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
This is the consolidation #869 asked for, and collapsing both strippers into one end-anchored regex is the right call.

I walked the leftmost-match behaviour against the old two-pass version and it holds: `-mxfp4-q4` still wins over the bit branch, `x-mlx-4bit` strips the quant before the marker, and the `m.start() > 0` guard keeps a bare `-4bit` returning None. The Option return staying meaningful for lookup callers is the detail that makes `normalize_mlx_repo_base` safe as a thin wrapper.

Test coverage matches the matrix in the issue, full CI is green across all three platforms, and the scope note about DWQ installed-detection staying untouched is the right boundary. Recommending merge.

PR #896, opened 2026-08-17
bench: community results for apple-m1-max

maxsagt:
Community benchmark from llmfit 1.1.10.

Hardware: **Apple M1 Max** (32 GB unified memory, 10 cores, macOS).

| --- | --- | --- | --- |
| qwen3.6:27b | ollama | 9.7 | 489.0 |


AlexsJones (approved):
Thanks for sharing the M1 Max result. I checked the community-data schema, reported unified-memory hardware, sample count, and timestamp; the data-only change is consistent, and the available check is green. Approving and recommending to @AlexsJones for final look and merge.

PR #902, opened 2026-08-18
fix(providers): find LM Studio models on disk, not just loaded ones

### LINKED ISSUE #900: [Bug]: LM Studio installed-detection only sees loaded models (1 of 13)
### Bug description

llmfit's installed-model detection for LM Studio only ever sees models that are currently loaded. Everything else in the library reads as not installed.

I have 13 models in LM Studio, 113 GB on disk. llmfit finds one of them, and only because it happens to be loaded at the time.

`lms ls`:

```
You have 13 models, taking up 113.14 GB of disk space.

LLM                                   PARAMS     ARCH        SIZE        DEVICE
andy-4.2                              9.0B       qwen35      10.45 GB    Local
google/gemma-4-12b (1 variant)        12B        gemma4      12.84 GB    Local
google/gemma-4-26b-a4b (1 variant)    26B-A4B    gemma4      23.83 GB    Local
meta-llama-3.1-8b-instruct            8B         Llama        4.92 GB    Local
qwen/qwen3.6-27b (1 variant)          27B        qwen35      17.48 GB    Local
qwen/qwen3.8-27b (1 variant)          27B        qwen35      17.74 GB    Local
qwen3-coder-30b-a3b-instruct          30B-A3B    qwen3moe    21.73 GB    Local

EMBEDDING                                        PARAMS    ARCH          SIZE         DEVICE
text-embedding-bge-m3                            567M      BERT          634.55 MB    Local     ✓ LOADED
text-embedding-nomic-embed-text-v1.5-embedding             Nomic BERT    146.15 MB    Local
text-embedding-nomic-embed-text-v1.5@q4_k_m                Nomic BERT     84.11 MB    Local
text-embedding-nomic-embed-text-v1.5@q8_0                  Nomic BERT    146.15 MB    Local
text-embedding-qwen3-embedding-0.6b              0.6B      qwen3         639.15 MB    Local
text-embedding-qwen3-embedding-4b                4B        qwen3           2.50 GB    Local
```

Of those, `bge-m3` is the only one llmfit marks installed, and it is the only one marked `✓ LOADED`.

### Expected behavior

Models downloaded in LM Studio are marked installed whether or not they are loaded, the same way an Ollama pull counts without the model being resident.

### Actual behavior

Only the loaded model is detected. The other twelve are offered as downloads I already have.

### Steps to reproduce

1. Download several models in LM Studio, load none of them (or one).
2. Start the LM Studio local server.
3. Run `llmfit fit --json` and look at the `installed` flags.

Both endpoints report only the loaded model on my setup:

```sh
curl http://127.0.0.1:1234/v1/models       # 1 entry
curl http://127.0.0.1:1234/api/v0/models   # 1 entry, state=loaded
```

### Cause

`LmStudioProvider::detect_with_installed` in `llmfit-core/src/providers.rs` builds the installed set purely from the HTTP API response, so llmfit can only know what the server chooses to advertise.

There is a disk scanner, but it cannot cover this either. `LlamaCppProvider::list_gguf_files` reads `llamacpp_models_dir()`, which is llmfit's own download cache, and it uses a flat `read_dir` rather than a recursive walk. LM Studio stores models as `publisher/repo/model.gguf`, two levels down.

I tested that: pointing `LLMFIT_MODELS_DIR` at the LM Studio models root changes nothing, while pointing it at a single folder that directly contains a `.gguf` immediately marks that model installed. So the scanner works and the nesting is what defeats it.

This looks like the same shape as #731 and #734, where probing the HTTP API alone was not enough to tell whether LM Studio was installed. #734 fixed the provider status by checking the filesystem but left the model list on the API.

Two ways to fix it, and I would rather build the one you want:

1. Shell out to `lms ls`. The data is exact and already grouped by model, and `find_binary`/`command_exists` are already there for llama.cpp. It adds a dependency on the `lms` CLI being present, and on parsing its output.
2. Walk `~/.lmstudio/models` recursively. No dependency on another vendor's CLI, and it would also fix the flat-scan limitation for anyone pointing `LLMFIT_MODELS_DIR` at a nested tree. It means recursive disk walking, which costs more on a large library.

Happy to send either, or both if you would rather have the directory walk as the general fix and `lms` as a fast path.

### llmfit version

1.1.10

### Operating system

### Affected component

Provider detection (LM Studio)

### GPU details (if relevant)

NVIDIA GeForce RTX 3090, 24 GB, CUDA

ba1lly:
Fixes #900.

Went with option 2 as you asked, including both bounds: depth cap of 3, and the extension tested before anything stats the entry.

## What changed

**The GGUF scan is now recursive.** `list_gguf_files` used a flat `read_dir`, so it could only ever see files sitting directly in the models root. It now walks up to `GGUF_SCAN_MAX_DEPTH` (3), which covers `publisher/repo/model.gguf` and stops there. As you said, this is a bug in its own right: anyone pointing `LLMFIT_MODELS_DIR` at a nested tree hit the same wall, and I confirmed that before changing anything. Pointing it at my LM Studio root detected nothing, while pointing it at a single folder that directly contained a `.gguf` detected that model immediately.

On the cost bound: the name is tested for `.gguf` straight from the directory entry, and only entries that are not models cost a `file_type` call. Unreadable directories are skipped rather than aborting the walk.

**LM Studio's models directory is now scanned.** `~/.lmstudio/models`, resolved through a pure helper so tests cover the layout without a home directory, in the same shape as `lmstudio_install_candidates`.

## One thing I did not expect, and how I handled it

My first attempt merged the directory names into the existing `lmstudio` set. Detection went from 6 models to 23, but several of the new ones were wrong: `Resilient-Coders/llama`, two `mlx-community/*-bf16` entries I do not have in that format, and the non-Instruct base variants of models where I only have Instruct.

The cause is that `is_model_installed_lmstudio` matches with a bare `contains`. That is survivable while the API reports one or two loaded models, but a directory scan produces far more names, and a catalog entry called `llama` is a substring of `meta-llama-3.1-8b-instruct`. Feeding more names into a loose matcher trades one wrong answer for several.

So directory-derived names go into their own set, `lmstudio_disk`, matched by equality via `is_model_installed_lmstudio_disk`. The API set and its substring rule are untouched, so there is no behaviour change for anyone whose models are loaded. Only canonical forms are inserted: `publisher/repo`, the repo with LM Studio's `-GGUF` suffix trimmed, and the file stem with and without its quantization suffix.

I have left the substring rule on the API path alone. Tightening it looked like a separate change with its own regression risk, and every stricter rule I tried broke a case that currently works: requiring a boundary still matches `llama` inside `meta-llama`, and requiring a prefix breaks `text-embedding-bge-m3` matching `bge-m3`. Happy to open that separately if you want it.

## Verification

Real library: 13 entries in `lms ls`, 113 GB.

| | catalog rows matched to LM Studio |
|---|---|
| before | 1 |
| after | 10 |

Twelve of the 13 entries now resolve. They land on 10 catalog rows rather than 12 because the three `nomic-embed-text-v1.5` variants share one catalog id. The single miss is `Mindcraft-CE/Andy-4.2`, which has no catalog entry, so nothing can match it.

No false positives: every one of the 10 corresponds to a file on disk. The two `DeepSeek-R1` entries visible in my testing come from the separate Ollama bug in #899 and are not from this branch.

On cost, comparing debug builds of `main` and this branch so the comparison is like for like, five runs each of `fit -n 1`:

```
main     median 4679 ms
patched  median 4742 ms
```

Inside the run-to-run spread. The walk visits 19 directories on that library. Worth noting that a naive comparison against the released binary showed a 940 ms gap, which was entirely debug versus release rather than the scan.

Seven tests added, covering the descent, the depth cap holding one level short, depth 1 reproducing the old flat behaviour, a missing root, the models directory layout, projector files not being counted as models, and the equality rule keeping `Resilient-Coders/llama` out. That last one fails if the exact match is swapped back to a substring, which I checked by doing exactly that.

No new dependency: the tests use a small self-cleaning temp directory rather than pulling in `tempfile`.

`cargo fmt --all -- --check`, `cargo clippy --all-targets` and the full suite pass (656 tests). `llmfit-desktop` does not build in my environment because `icons/icon.ico` is missing, unrelated to this change, so I tested the workspace default members.

## Follow-up

`lms ls` as a fast path is not here, per your steer that it should be an optimisation rather than the mechanism. Nothing in the measurements suggests it is needed yet.

### DIFF
--- a/llmfit-core/src/analysis.rs
+++ b/llmfit-core/src/analysis.rs
@@ -23,6 +23,11 @@ pub struct InstalledIndex {
     pub docker_mr_count: usize,
     pub lmstudio: HashSet<String>,
     pub lmstudio_count: usize,
+    /// Models found in LM Studio's models directory. Kept apart from
+    /// `lmstudio` because the API ids there are matched by substring, while
+    /// these directory-derived names are matched by equality.
+    pub lmstudio_disk: HashSet<String>,
+    pub lmstudio_disk_count: usize,
     pub vllm: HashSet<String>,
     pub vllm_count: usize,
     pub ramalama: HashSet<String>,
@@ -42,6 +47,8 @@ impl InstalledIndex {
             docker_mr_count: 0,
             lmstudio: HashSet::new(),
             lmstudio_count: 0,
+            lmstudio_disk: HashSet::new(),
+            lmstudio_disk_count: 0,
             vllm: HashSet::new(),
             vllm_count: 0,
             ramalama: HashSet::new(),
@@ -73,6 +80,7 @@ impl InstalledIndex {
                 let p = LmStudioProvider::new();
                 p.installed_models_counted()
             });
+            let lmstudio_disk = s.spawn(providers::scan_lmstudio_models_dir);
             let vllm = s.spawn(|| {
                 let p = VllmProvider::new();
                 p.installed_models_counted()
@@ -87,6 +95,10 @@ impl InstalledIndex {
             let (llamacpp, llamacpp_count) = llamacpp.join().unwrap();
             let (docker_mr, docker_mr_count) = docker_mr.join().unwrap();
             let (lmstudio, lmstudio_count) = lmstudio.join().unwrap();
+            // Enrichment rather than a load-bearing provider: if the scan
+            // thread dies, report no disk models instead of taking the whole
+            // installed-model analysis down with it.
+            let (lmstudio_disk, lmstudio_disk_count) = lmstudio_disk.join().unwrap_or_default();
             let (vllm, vllm_count) = vllm.join().unwrap();
             let (ramalama, ramalama_count) = ramalama.join().unwrap();
 
@@ -100,6 +112,8 @@ impl InstalledIndex {
                 docker_mr_count,
                 lmstudio,
                 lmstudio_count,
+                lmstudio_disk,
+                lmstudio_disk_count,
                 vllm,
                 vllm_count,
                 ramalama,
@@ -120,6 +134,7 @@ impl InstalledIndex {
             || providers::is_model_installed_llamacpp(model_name, &self.llamacpp)
             || providers::is_model_installed_docker_mr(model_name, &self.docker_mr)
             || providers::is_model_installed_lmstudio(model_name, &self.lmstudio)
+            || providers::is_model_installed_lmstudio_disk(model_name, &self.lmstudio_disk)
             || providers::is_model_installed_vllm(model_name, &self.vllm)
             || providers::is_model_installed_ramalama(model_name, &self.ramalama)
     }
@@ -141,7 +156,9 @@ impl InstalledIndex {
         if providers::is_model_installed_docker_mr(model_name, &self.docker_mr) {
             out.push("Docker");
         }
-        if providers::is_model_installed_lmstudio(model_name, &self.lmstudio) {
+        if providers::is_model_installed_lmstudio(model_name, &self.lmstudio)
+            || providers::is_model_installed_lmstudio_disk(model_name, &self.lmstudio_disk)
+        {
             out.push("LM Studio");
         }
         if providers::is_model_installed_vllm(model_name, &self.vllm) {
--- a/llmfit-core/src/analysis.rs
+++ b/llmfit-core/src/analysis.rs
@@ -23,6 +23,11 @@ pub struct InstalledIndex {
     pub docker_mr_count: usize,
     pub lmstudio: HashSet<String>,
     pub lmstudio_count: usize,
+    /// Models found in LM Studio's models directory. Kept apart from
+    /// `lmstudio` because the API ids there are matched by substring, while
+    /// these directory-derived names are matched by equality.
+    pub lmstudio_disk: HashSet<String>,
+    pub lmstudio_disk_count: usize,
     pub vllm: HashSet<String>,
     pub vllm_count: usize,
     pub ramalama: HashSet<String>,
@@ -42,6 +47,8 @@ impl InstalledIndex {
             docker_mr_count: 0,
             lmstudio: HashSet::new(),
             lmstudio_count: 0,
+            lmstudio_disk: HashSet::new(),
+            lmstudio_disk_count: 0,
             vllm: HashSet::new(),
             vllm_count: 0,
             ramalama: HashSet::new(),
@@ -73,6 +80,7 @@ impl InstalledIndex {
                 let p = LmStudioProvider::new();
                 p.installed_models_counted()
             });
+            let lmstudio_disk = s.spawn(providers::scan_lmstudio_models_dir);
             let vllm = s.spawn(|| {
                 let p = VllmProvider::new();
                 p.installed_models_counted()
@@ -87,6 +95,7 @@ impl InstalledIndex {
             let (llamacpp, llamacpp_count) = llamacpp.join().unwrap();
             let (docker_mr, docker_mr_count) = docker_mr.join().unwrap();
             let (lmstudio, lmstudio_count) = lmstudio.join().unwrap();
+            let (lmstudio_disk, lmstudio_disk_count) = lmstudio_disk.join().unwrap();
             let (vllm, vllm_count) = vllm.join().unwrap();
             let (ramalama, ramalama_count) = ramalama.join().unwrap();
 
@@ -100,6 +109,8 @@ impl InstalledIndex {
                 docker_mr_count,
                 lmstudio,
                 lmstudio_count,
+                lmstudio_disk,
+                lmstudio_disk_count,
                 vllm,
                 vllm_count,
                 ramalama,
@@ -115,6 +126,7 @@ impl InstalledIndex {
             || providers::is_model_installed_llamacpp(model_name, &self.llamacpp)
             || providers::is_model_installed_docker_mr(model_name, &self.docker_mr)
             || providers::is_model_installed_lmstudio(model_name, &self.lmstudio)
+            || providers::is_model_installed_lmstudio_disk(model_name, &self.lmstudio_disk)
             || providers::is_model_installed_vllm(model_name, &self.vllm)
             || providers::is_model_installed_ramalama(model_name, &self.ramalama)
     }
@@ -135,7 +147,9 @@ impl InstalledIndex {
         if providers::is_model_installed_docker_mr(model_name, &self.docker_mr) {
             out.push("Docker");
         }
-        if providers::is_model_installed_lmstudio(model_name, &self.lmstudio) {
+        if providers::is_model_installed_lmstudio(model_name, &self.lmstudio)
+            || providers::is_model_installed_lmstudio_disk(model_name, &self.lmstudio_disk)
+        {
             out.push("LM Studio");
         }
         if providers::is_model_installed_vllm(model_name, &self.vllm) {
--- a/llmfit-core/src/analysis.rs
+++ b/llmfit-core/src/analysis.rs
@@ -95,7 +95,10 @@ impl InstalledIndex {
             let (llamacpp, llamacpp_count) = llamacpp.join().unwrap();
             let (docker_mr, docker_mr_count) = docker_mr.join().unwrap();
             let (lmstudio, lmstudio_count) = lmstudio.join().unwrap();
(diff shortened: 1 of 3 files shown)


ba1lly:
All three fixed in 207e157.

The variant suffix one was a real bug, thanks. `hf_name_to_lmstudio_candidates` also hands back a form with `-instruct`, `-chat`, `-hf` and `-it` stripped off, which is fine for substring matching but collides under equality: `gemma-4-12b-it` collapses to `gemma-4-12b`, so an installed base model claims the IT variant. I only have the IT side locally, which is why my numbers looked clean. The disk path builds its own candidates now, full name and bare repo, no stripping.

On the non-files check, the bit I'd missed was the early `continue`. A directory called `foo.gguf` got returned as a model *and* never looked inside. It now checks the entry is actually a file, and falls through to descend if it isn't. This doesn't undo the cost bound from #900, since `file_type` reads what `read_dir` already gave us rather than stat-ing the path, so the name check is still doing the filtering.

The join degrades to no disk models now instead of panicking. It's deliberately different from the six joins around it, which do unwrap. A directory scan is enrichment rather than a provider, and adding a new panic path into installed-model analysis felt wrong. Can make them uniform if you'd rather.

Two tests added. The variant one fails if you put the stripped candidates back, which I checked. Still 10 catalog rows detected on my library, so nothing lost.

ba1lly:
Symlink one fixed in 8b002be. I think the variant-suffix comment is stale now, more on that below.

It was a regression I'd introduced a commit earlier rather than a pre-existing gap. `DirEntry::file_type` describes the link, not its target, so the `is_file` check I added to stop directories being treated as models also threw away symlinked ones that the old name-only scan picked up.

Anything whose name already looks like a model gets resolved with `metadata` now. Only those pay for the call, so the #900 cost bound holds. Symlinked directories get followed too, with the depth cap bounding any cycle.

I couldn't verify it on Windows, where creating a symlink needs elevation and the test just skips, so I ran it on Linux: fails against the non-resolving check, passes with `metadata`. The test compiles on both platforms rather than being unix-gated, so it isn't dead code on the Windows runner.

On the variant suffixes, that's anchored to `providers.rs:2116` on 207e157, which is the call to `hf_name_to_lmstudio_disk_candidates`. That helper went in with the same commit precisely to drop the stripping. `hf_name_to_lmstudio_candidates` is now only reached from `is_model_installed_lmstudio` (the API path, left alone on purpose) and `poll_lmstudio_installed_models`, neither of which feeds the disk set. `test_lmstudio_disk_base_model_does_not_claim_variants` covers the exact case, base `gemma-4-12b` on disk with `-it`, `-instruct` and `-chat` asserted uninstalled, and it fails if the stripped candidates go back in.

If it's actually pointing at the collision on the API substring path, that one is real, but it predates this PR and I left it alone for the reasons in the description. Can pick it up separately if you want.

659 tests passing.

ba1lly:
Fixed in db44619.

Following directory symlinks was something I added alongside the symlinked-file fix, not behaviour the original scan had. You're right that it lets the walk leave the models root, or alias a subtree back into itself, with callers acting on whatever comes back.

Descent is real directories only now. Symlinked model *files* still get resolved, which was the regression that actually came up. That seemed better than bolting root-containment and canonical-path dedup onto a traversal nobody asked for, though say if you'd rather have it with the guards.

Verified on Linux again, since symlinks need elevation on Windows and the tests skip there. Six scan tests pass, and the escape test fails the moment directory symlinks are followed.

Still seeing the variant-suffix finding, third time now. `providers.rs:2137` is the `.any(...)` inside `is_model_installed_lmstudio_disk`, and that takes its candidates from `hf_name_to_lmstudio_disk_candidates`, which does no stripping. Only `is_model_installed_lmstudio` and `poll_lmstudio_installed_models` still call the stripping version, and neither touches the disk set. The test asserts that exact scenario and fails if the stripped candidates are restored. I don't think there's anything left there, but shout if you read it differently.

660 tests passing.

ba1lly:
Fixed in edf2b12, and it turned up a bug in one of my own tests.

Collected models get deduplicated on their resolved path now, preferring the target over the alias so callers end up with the stable one. Resolution runs once per model found rather than per directory entry, so it scales with models rather than with files walked.

The test bug: the symlink test I added two commits back had its target inside the scanned tree. Once dedupe landed that's one model, not two, so it was asserting the wrong count and would have gone red. It only passed here because symlinks need elevation on Windows and it skips. The target sits outside the tree now, which is the case it was meant to cover anyway.

I've been re-running these on Linux through WSL rather than trusting the local run, given the symlink paths never execute on Windows. Seven scan tests pass there, and the alias test fails with dedupe removed.

Detection unchanged at 10 catalog rows. 661 tests passing.

ba1lly:
Fixed in 9b18eac.

It's worse than the tie-break simply not applying. If the root or any parent is a link, `canonicalize` resolves the whole prefix, so neither the model nor its alias matches the target and that branch can never fire at all. The survivor ends up decided by directory order, and a stem lookup then fails on a model that's sitting right there. It asks whether the entry is itself a link now, which doesn't care about the prefix.

The first version of the test drove `collect_gguf_files` and passed with the old tie-break still in, because `read_dir` happened to hand back the target first. It was effectively asserting directory order, the one thing that mustn't decide this. It calls `dedupe_by_target` directly now, with the alias placed first, and checks both orderings. Fails with the old tie-break on Linux, passes with the new one.

662 tests, and detection has stayed at 10 catalog rows throughout.

AlexsJones (approved):
This is a really careful piece of work, and the part I want to call out is that you noticed the naive version made detection *worse* and split the sets instead of shipping the bigger number.

Keeping `lmstudio_disk` separate with equality matching is the right shape: the API substring rule survives untouched, so nobody's loaded-model detection changes, and the disk names never get fed to a matcher that would let `llama` swallow `meta-llama-3.1-8b-instruct`. Same for not reusing `hf_name_to_lmstudio_candidates`, the `-it` stripping really would collide under equality.

The scan itself holds up on the edge cases I went looking for: a directory named `foo.gguf` still gets descended into, directory symlinks are not followed so the walk cannot escape the root or double-count a subtree, `symlink_metadata` (not canonical-form comparison) picks the survivor in `dedupe_by_target`, and joining the scan thread with `unwrap_or_default` keeps a panic there from taking the whole installed-model analysis down. Bounding the recursion at 3 rather than adding a walk dependency fits the repo.

Also right to leave the API-path substring rule alone, that is its own change with its own regressions. Worth a separate issue if you want it.

Full CI green across all three platforms, TempTree keeps the tests dependency-free, and the before/after on a real 13-model library is the verification I would have asked for. Recommending merge.

AlexsJones:
Please resolve conflicts and we can merge @ba1lly

ba1lly:
@AlexsJones rebased onto main, conflicts resolved. Squashed to one commit since it was a single change plus its review fixes. The conflicts were all the same shape: your size-aware family tests landed in the same test block as the scan tests, nothing semantically overlapping, both kept. Core suite passes locally at 673 tests, fmt clean, clippy matches the main baseline.

On Greptile's new P1 about bare names conflating publishers, that one is a deliberate trade-off rather than an oversight. I checked it against the catalog: 547 of 12,937 entries share a basename across publishers, about 4%, and nearly all of those are mirrors or GGUF repacks of the same weights (baai, xenova and unsloth for bge-small-en-v1.5; cohereforai and coherelabs for command-r), where reporting both installed is the right answer.

The bare candidate can't come out without undoing the fix. LM Studio's publisher directory is whoever repacked to GGUF, not the catalog's publisher, so meta-llama/Meta-Llama-3.1-8B-Instruct sits on disk under unsloth/Meta-Llama-3.1-8B-Instruct-GGUF. Publisher-qualified matching alone misses it, which is the 1-of-13 case this change exists to fix. Equality matching is what keeps the blast radius at exact basename twins instead of every substring.

AlexsJones (approved):
Nice work on the recursive scan. The depth cap, the extension-tested-before-stat, the unreadable-directory skip, and the symlink dedup are all the right calls, and keeping `lmstudio_disk` separate from `lmstudio` (equality vs substring matching) avoids the false positives you hit on the first attempt. The `unwrap_or_default` on the disk-scan thread is a good touch so a scan failure doesn't take the whole installed-model analysis down.

I've read the diff end to end; I'm not re-running the Rust suite here, but the directory-layout helper and the store fallback are covered by the tests you listed. Good to go.

Issue #638, opened 2026-06-17
Detect multiple GPUs


agileramblings:
I have a dual Radeon AI PRO R9700 setup. llmfit does not seem to detect that I have both GPUs and provide appropriate recommendations. 

<img width="1701" height="327" alt="Image" src="https://github.com/user-attachments/assets/592905e8-94bb-4119-b8cb-f9982ac88d7a" />

AlexsJones:
Thanks for reporting this. I traced the root cause.

The sysfs detection path (`detect_amd_gpu_sysfs_info`) is used when ROCm is not installed, which matches your setup (the TUI shows the Vulkan backend). The function was returning `Option<GpuInfo>` instead of `Vec<GpuInfo>`, so it returned on the first AMD card it found in `/sys/class/drm/` and never enumerated additional cards. The Vulkan fallback path could not compensate because its dedup logic saw the second GPU had the same name as the first and skipped it.

The fix changes `detect_amd_gpu_sysfs_info` to iterate all sysfs card entries, group identical models with a count (like the ROCm and NVIDIA detection paths already do), and return the full set. For your dual R9700 setup, this means both cards get detected and the TUI should show something like "Radeon AI PRO R9700 x2 (31.9 GB each = 64 GB total, Vulkan)" with recommendations scaled to the combined VRAM.

cb88:
Just built this and it only detects 1 of my MI50s.... 
[cb88@MZ31AR0 release]$ ./llmfit system

=== System Specifications ===
CPU: AMD EPYC 7352 24-Core Processor (48 cores)
Total RAM: 125.65 GB
Available RAM: 121.14 GB
Backend: ROCm
GPU: AMD Instinct MI60 / MI50 (31.98 GB VRAM, ROCm)

cb88@MZ31AR0 release]$ lspci -vv | grep Disp
lspci: Unable to load libkmod resources: error -2
03:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Vega 20 [Radeon Pro VII/Radeon Instinct MI50] (rev 02)
44:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Vega 20 [Radeon Pro VII/Radeon Instinct MI50]

[cb88@MZ31AR0 release]$ rocm-smi
<snip>
0       2     0x66a1,   16893  28.0°C  21.0W     N/A, N/A, 0         925Mhz  350Mhz  14.51%  auto  225.0W  1%     0%
[...]

PR #913, opened 2026-08-19
bench: community results for nvidia-geforce-rtx-5090-laptop-gpu

sleggett-ai:

| --- | --- | --- | --- |
| unsloth/Qwen3.6-35B-A3B-GGUF:UD-IQ4_XS | llamacpp | 149.1 | — |


AlexsJones (approved):
Thanks for this, llamacpp submissions are rarer than ollama ones so a 5-run 5090 Laptop entry is welcome.

Schema and layout check out. One thing worth flagging (not blocking): the record reports `gpuCount: 4` with `vramGb: 23.89`, which looks like our detection counting something extra on that machine. If that machine really has one GPU, it would be useful context on #638.

PR #922, opened 2026-08-20
fix(hardware): stop mobile GPUs inheriting desktop specs

### LINKED ISSUE #919: [Bug]: Laptop GPUs inherit their desktop namesake's memory bandwidth (RTX 5070 Laptop -> 672 GB/s on a 128-bit bus), inflating tok/s
### Bug description

`gpu_memory_bandwidth_gbps()` matches on the model number alone, so every mobile SKU inherits the bandwidth of its **desktop** namesake. Laptop parts have narrower memory buses than the desktop cards they share a name with, so the returned figure is not just optimistic — it is physically impossible for the detected device.

On this machine llmfit detects `NVIDIA GeForce RTX 5070 Laptop GPU` and reports **672 GB/s**. That is the desktop RTX 5070 value (192-bit × 28 Gbps). The Laptop SKU has a **128-bit** bus; reaching 672 GB/s on 128 bits would need 42 Gbps memory, and the fastest GDDR7 shipping today is 32 Gbps. Published figures for this card are 384–512 GB/s depending on the source.

Since `tok/s ≈ bandwidth × efficiency ÷ weights`, the estimate is inflated by the same ratio — ~1.3–1.75× for this card.

The mismatch is in [`llmfit-core/src/hardware.rs`](https://github.com/AlexsJones/llmfit/blob/main/llmfit-core/src/hardware.rs), in the substring chain:

```rust
if lower.contains("5070") {
    return Some(672.0);
}
```

`"nvidia geforce rtx 5070 laptop gpu".contains("5070")` is `true`, and nothing downstream looks for `laptop` / `mobile` / `max-q`. Grepping the file, those tokens appear only in the AMD iGPU helper and in `match_registry_vram` fixtures — the bandwidth path has no notion of mobile parts.

It is not specific to Blackwell. Some examples where the desktop and laptop parts differ by bus width:

| Detected name | Bus | Actual bandwidth | llmfit returns | Overestimate |
|---|---|---|---|---|
| RTX 5070 Laptop GPU | 128-bit | 384–512 GB/s | 672 (desktop, 192-bit) | 1.3–1.75× |
| RTX 5070 Ti Laptop GPU | 192-bit | 672 GB/s | 896 (desktop, 256-bit) | 1.33× |
| RTX 4090 Laptop GPU | 256-bit | 576 GB/s | 1008 (desktop, 384-bit) | 1.75× |
| RTX 4080 Laptop GPU | 192-bit | 432 GB/s | 717 (desktop, 256-bit) | 1.66× |
| RTX 3080 Ti Laptop GPU | 256-bit | 512 GB/s | 912 (desktop, 384-bit) | 1.78× |

`estimate_vram_from_name()` has the same shape (`contains("5070") → 12.0`, while the 5070 Laptop ships 8 GB, or 12 GB since May 2026), though that one only fires as a fallback when VRAM reads 0.

Related in spirit to #840, where a bad hardware reading produced unrealistic GPU fit estimates.

### Expected behavior

For a mobile SKU, either return the laptop part's own bandwidth, or return `None` so the caller falls back to the fixed constant — which the function already documents as the contract:

> Returns `None` when the GPU is not recognized; callers should fall back to the existing fixed-constant approach.

Returning `None` is the conservative fix: no table of mobile values to maintain, and the estimate degrades to the generic constant instead of a wrong-by-1.75× number presented as per-GPU precision.

### Actual behavior

The desktop value is returned for the laptop part, and it propagates into `estimate_basis.gpu_bandwidth_gbps` and every tok/s estimate derived from it:

```
$ llmfit system --json
"gpus": [
  {
    "backend": "CUDA",
    "count": 1,
    "memory_bandwidth_gbps": 672.0,
    "name": "NVIDIA GeForce RTX 5070 Laptop GPU",
    "unified_memory": false,
    "vram_gb": 7.96
  }
]
```

Nothing in the output marks it as a name-derived guess, so a consumer of the API cannot tell it apart from a measured value.

### Steps to reproduce

1. On a laptop with any RTX Laptop GPU (here: RTX 5070 Laptop GPU, 8 GB), run `llmfit system --json`.
2. Read `gpus[0].memory_bandwidth_gbps` → `672.0`.
3. Compare against the card's bus width (128-bit) — no GDDR7 speed grade yields 672 GB/s on 128 bits.
4. Any `llmfit fit` / `recommend` / `/api/v1/models` tok/s for this machine is scaled by that figure.

No laptop needed to reproduce the mismatch itself:

```rust
assert_eq!(gpu_memory_bandwidth_gbps("NVIDIA GeForce RTX 5070 Laptop GPU"), Some(672.0));
// same value as the desktop part:
assert_eq!(gpu_memory_bandwidth_gbps("NVIDIA GeForce RTX 5070"), Some(672.0));
```

### Suggested fix

Guard the substring chain before it runs:

```rust
pub fn gpu_memory_bandwidth_gbps(name: &str) -> Option<f64> {
    let lower = name.to_lowercase();

    // Laptop parts share the model number with a desktop card but not its
    // memory bus (RTX 5070: 192-bit desktop vs 128-bit mobile), so the
    // desktop figure would overestimate bandwidth by up to ~1.8x.
    if lower.contains("laptop") || lower.contains("mobile") || lower.contains("max-q") {
        return None;
    }
    ...
```

`nvidia-smi` names these parts consistently with the `Laptop GPU` suffix, so the guard is reliable. If you would rather keep per-GPU precision for mobile parts, I am happy to send a PR with a mobile table (RTX 50/40/30 Laptop) plus regression tests — say which of the two you prefer.

### Diagnostic report (`llmfit doctor`)

```
# llmfit doctor report

- llmfit version: 1.1.10
- OS: windows (x86_64)

## Detected by llmfit

SystemSpecs {
    total_ram_gb: 64.0,
    available_ram_gb: 20.425132751464844,
    total_cpu_cores: 16,
    cpu_name: "AMD Ryzen 9 270 w/ Radeon 780M Graphics",
    has_gpu: true,
    gpu_vram_gb: Some(
        7.9599609375,
    ),
    total_gpu_vram_gb: Some(
        7.9599609375,
    ),
    gpu_available_gb: None,
    gpu_name: Some(
        "NVIDIA GeForce RTX 5070 Laptop GPU",
    ),
    gpu_count: 1,
    unified_memory: false,
    backend: Cuda,
    gpus: [
        GpuInfo {
            name: "NVIDIA GeForce RTX 5070 Laptop GPU",
            vram_gb: Some(
                7.9599609375,
            ),
            backend: Cuda,
            count: 1,
            unified_memory: false,
        },
    ],
    cluster_mode: false,
    cluster_node_count: 0,
}

## nvidia-smi (extended query)

[N/A], 8151, NVIDIA GeForce RTX 5070 Laptop GPU

## nvidia-smi (standard query)

8151, NVIDIA GeForce RTX 5070 Laptop GPU

## rocm-smi --showmeminfo vram

(not available: program not found)
```

### Version

llmfit 1.1.10 (Windows x86_64, CUDA backend)

matiasagustincastro-byte:
Fixes #919.

`gpu_memory_bandwidth_gbps()` and `estimate_vram_from_name()` match on the model number alone, so every mobile SKU inherits the figures of the desktop card it shares a name with. They are different silicon: an RTX 5070 Laptop GPU has a 128-bit bus and 8 GB against the desktop card's 192-bit and 12 GB. The 672 GB/s currently returned for it would need 42 Gbps memory on 128 bits; the fastest GDDR7 shipping is 32.

`tok/s ≈ bandwidth × efficiency ÷ weights`, so the error lands directly on every estimate — up to 1.8× across the 30/40/50 mobile lines (RTX 3080 Ti Laptop: 512 GB/s actual, 912 returned).

## What this does

Per your review, the marker is detected once, before the substring chains run, so it covers every entry rather than a hand-listed few:

```rust
fn is_mobile_gpu_name(lower: &str) -> bool {
    lower.contains("laptop") || lower.contains("mobile") || lower.contains("max-q")
}
```

- **Bandwidth returns `None`.** That is the function's documented contract for an unrecognized GPU, so callers take the fixed-constant path. No mobile table to keep current as new laptop parts ship, and the estimate degrades to a generic figure instead of a confidently wrong one.
- **VRAM falls through to the generic per-family estimate** (`rtx` → 8 GB, `gtx` → 4 GB), which meant extracting that tail into `estimate_vram_generic()`. Returning `0.0` instead would be worse than the status quo: `resolve_wmi_vram()` and the sysfs paths only accept a positive estimate, so a mobile card with an unreadable VRAM reading would be dropped rather than estimated.

## Verification

On the machine from the issue (RTX 5070 Laptop GPU, 8 GB, Windows 11 / CUDA):

```json
// before
{ "name": "NVIDIA GeForce RTX 5070 Laptop GPU", "memory_bandwidth_gbps": 672.0, "vram_gb": 7.96 }
// after
{ "name": "NVIDIA GeForce RTX 5070 Laptop GPU", "memory_bandwidth_gbps": null, "vram_gb": 7.96 }
```

`cargo fmt`, `cargo clippy -p llmfit-core --all-targets` (no new warnings) and `cargo test -p llmfit-core` (572 passing) are clean. `llmfit-desktop` does not build here — its Tauri build script needs `llmfit-web/dist` — which is unrelated to this change. New regression tests cover the four laptop names from the issue table plus Max-Q and Mobile, and assert that the desktop entries and `Apple M3 Max` (not the `max-q` marker) are untouched.

## One judgment call worth flagging

The guard sits at the very top of both functions, which means it also shadows the AMD unified-memory APU entries (Strix Halo / Radeon 800M) if a detection path ever reports one of those names with a `Mobile` suffix — those parts only ship in laptops, so their entries already *are* the mobile figures. I could not find a detection path that produces such a string (lspci gives `Strix Halo [Radeon 8060S Graphics]`, the CPU-derived fallback gives `AMD Ryzen AI MAX+ 395 w/ Radeon 8060S (integrated)`, and `MAX+` does not match `max-q`), so I kept the guard simple rather than carving out an exception for a hypothetical. Happy to add the exclusion if you know of one.

The JSON output still not marking these as name-derived guesses is a real gap, but as you said, separate from this fix.

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -2661,6 +2661,18 @@ fn measure_ram_bandwidth_gbps() -> Option<f64> {
     (2.0..=4000.0).contains(&total).then_some(total)
 }
 
+/// True for a mobile (laptop) SKU.
+///
+/// NVIDIA suffixes these adapters with "Laptop GPU"; earlier generations use
+/// "Max-Q" or "Mobile". The model-number tables in this module describe
+/// desktop cards, and the mobile part that shares a number is a different
+/// piece of silicon: an RTX 5070 Laptop GPU carries a 128-bit bus and 8 GB
+/// against the desktop card's 192-bit and 12 GB. Matching on the number
+/// alone is therefore wrong for every one of them (#919).
+fn is_mobile_gpu_name(lower: &str) -> bool {
+    lower.contains("laptop") || lower.contains("mobile") || lower.contains("max-q")
+}
+
 /// Estimate GPU memory bandwidth in GB/s from the GPU model name.
 ///
 /// Token generation in LLM inference is memory-bandwidth-bound (each token
@@ -2674,10 +2686,21 @@ fn measure_ram_bandwidth_gbps() -> Option<f64> {
 ///  - Google, "Efficiently Scaling Transformer Inference" (arXiv:2211.05102)
 ///  - ggerganov, llama.cpp NVIDIA T4 benchmarks (Discussion #4225)
 ///
-/// Returns `None` when the GPU is not recognized; callers should fall back
-/// to the existing fixed-constant approach.
+/// Returns `None` when the GPU is not recognized, and for mobile SKUs, whose
+/// model number does not identify their memory subsystem; callers should fall
+/// back to the existing fixed-constant approach.
 pub fn gpu_memory_bandwidth_gbps(name: &str) -> Option<f64> {
     let lower = name.to_lowercase();
+
+    // A mobile part shares its model number with a desktop card but not the
+    // memory bus behind it: an RTX 5070 Laptop GPU is 128-bit where the
+    // desktop 5070 is 192-bit, so the chain below would hand back 672 GB/s
+    // for a card that cannot physically exceed ~512. The overestimate runs
+    // to 1.8x on the 30/40/50 mobile lines and propagates into every tok/s
+    // figure. Fall back to the fixed constant instead of guessing (#919).
+    if is_mobile_gpu_name(&lower) {
+        return None;
+    }
 
     // ── NVIDIA Consumer (GeForce) ──────────────────────────────────
     // RTX 50 series (Blackwell)
@@ -3126,6 +3149,13 @@ fn is_nvidia_unified_memory_gpu(name: &str) -> bool {
 /// Used when nvidia-smi or other tools report 0 VRAM.
 fn estimate_vram_from_name(name: &str) -> f64 {
     let lower = name.to_lowercase();
+    // Same trap as in `gpu_memory_bandwidth_gbps`: the table below is keyed
+    // on the model number, which a mobile part shares with a desktop card
+    // that has a different memory configuration (RTX 5070: 12 GB desktop,
+    // 8 GB laptop). Skip it and take the generic per-family estimate (#919).
+    if is_mobile_gpu_name(&lower) {
+        return estimate_vram_generic(&lower);
+    }
     // NVIDIA RTX 50 series
     if lower.contains("5090") {
         return 32.0;
@@ -3329,6 +3359,14 @@ fn estimate_vram_from_name(name: &str) -> f64 {
         return 8.0;
     }
 
+    estimate_vram_generic(&lower)
+}
+
+/// VRAM estimate that does not depend on the model number: the integrated-GPU
+/// carve-out and the per-family defaults. Used for GPUs missing from the table
+/// above, and for mobile SKUs, where a model-number match would describe the
+/// desktop card instead.
+fn estimate_vram_generic(lower: &str) -> f64 {
     // Integrated GPUs (APU iGPUs) — must check before generic fallbacks
     // APU names like "AMD Radeon(TM) Graphics" or "Radeon Graphics" without
     // a discrete model number (RX/HD/R5/R7/R9) have very limited dedicated VRAM.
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Exactly the shape we landed on in the issue, and pulling the marker check above the substring chain is what makes it hold for parts nobody has added to the table yet.

Checked both guards and the tests. `is_mobile_gpu_name` matches laptop / mobile / max-q on the lowercased name, `Apple M3 Max` is explicitly held out from the `max-q` case, and the desktop entries keep their exact figures on both functions. Extracting `estimate_vram_generic` was the right call rather than returning `0.0`, since `resolve_wmi_vram()` and the sysfs paths only accept a positive estimate and would otherwise drop the card entirely.

On the Strix Halo / Radeon 800M question you flagged: agreed, leave it simple. Those entries already carry mobile figures, so shadowing them costs nothing real, and no detection path produces a `Mobile` suffix for them today. If one ever does, failing open to the generic estimate is the safe direction.

One heads-up rather than a change request: PR #925 opened a few hours after yours and rewrites `gpu_memory_bandwidth_gbps` the other way, with an `nvidia_mobile_bandwidth_gbps` table. Alex already called the `None` path over a mobile table on #919, so yours is the branch I am recommending. Expect a conflict on that PR, not on this one.

Only Greptile ran on this branch (repo CI does not fire automatically for a first contribution), so the three-platform suite still needs a maintainer trigger before merge. Recommending it to Alex for the final look.

PR #924, opened 2026-08-21
fix(plan): use active MoE params in fallback estimates

### LINKED ISSUE #921: [Bug]: plan ignores MoE active params on GPUs missing from the bandwidth table (13x underestimate)
### Bug description

`llmfit plan` still reduces a sparse MoE model to its total parameter count when the GPU is absent
from the bandwidth table. This is the branch that #805 and PR #807 did not cover.

PR #807 replaced the duplicated estimator with a delegation to `fit.rs`. The delegation is
conditional (`llmfit-core/src/plan.rs:234`). The GPU path passes `Some(system)` at `plan.rs:385`,
so the only condition that can fail is the bandwidth lookup:

```rust
if path != PlanRunPath::CpuOnly
    && let Some(specs) = system
    && let Some(name) = specs.gpu_name.as_deref()
    && gpu_memory_bandwidth_gbps(name).is_some()
{
    return crate::fit::estimate_tps(/* … */);
}

// Fallback: fixed-constant approach
let k: f64 = match backend { /* … */ GpuBackend::Sycl => 100.0, /* … */ };
let mut base = (k / params) * quant_speed_multiplier(quant);
```

`params` comes from `model.params_b()` at `plan.rs:221` — the total count. The fallback holds no
`is_moe` branch and no reference to `active_parameters`.

`fit.rs` resolves the same value before it splits into branches
(`llmfit-core/src/fit.rs:1186`):

```rust
let params = model
    .active_parameters
    .filter(|_| model.is_moe)
    .map(|p| (p as f64) / 1_000_000_000.0)
    .unwrap_or_else(|| model.params_b())
    .max(0.1);
```

So `fit.rs` keeps MoE sparsity on both its branches, and `plan.rs` keeps it on one.

The condition selects exactly the users who most need the estimate. `llmfit info` tells them their
GPU is unknown and promises a wide error band:

```
Estimate Basis:
  Method: per-backend heuristic constant — GPU not in the bandwidth table,
          so expect a wide error band.
```

A 13x underestimate is not a wide band. It is a different model.

### Version and environment

- llmfit 1.1.10 (installed from PyPI with `uv tool install llmfit`). The same divergence was
  present on 1.1.9, so this is not a new regression.
- Linux x86_64
- CPU: Intel Core Ultra 7 255H, 16 cores
- GPU: Intel Arc Pro 130T/140T (integrated), Arrow Lake-P, PCI 8086:7d51, `i915` driver
- 62.32 GB unified memory
- llmfit detects the backend as SYCL. The llama.cpp build on this host runs the Vulkan backend
  through Mesa ANV, so the effective backend and the detected backend differ.
- llmfit measures RAM bandwidth at 128 to 151 GB/s across runs.

This GPU is not in `gpu_memory_bandwidth_gbps`, which is what puts `plan` on the fallback branch.
The full `llmfit doctor` report is at the end of this issue.

### Steps to reproduce

Model: `Qwen/Qwen3.6-35B-A3B`. The catalog records `is_moe: true`, `num_experts: 256`,
`active_experts: 8` and `active_parameters: 3000000000` against 36.0B total — a sparsity ratio
of 12. `llmfit info` prints the same split as 1.8 GB of active weights against 18.4 GB for the
full model.

```console
$ llmfit --json plan "Qwen/Qwen3.6-35B-A3B" --context 32768 --quant Q8_0 \
    | python3 -c "import json,sys;d=json.load(sys.stdin);print([(p['path'],p['estimated_tps']) for p in d['run_paths']])"
[('gpu', 2.2252001145716376), ('cpu_offload', None), ('cpu_only', 0.46729202406004383)]
```

`fit` and `info` report the same model, same quant, same machine at 29.3 tok/s (table row
trimmed to fit):

```console
$ llmfit --max-context 32768 fit | grep "Qwen/Qwen3.6-35B-A3B"
│ 🟢 Perfect │ Qwen/Qwen3.6-35B-A3B │ Alibaba │ 36.0B │ 83 │ 29.3 │ Q8_0 │ llama.cpp │ GPU │ 29.5% │ 262k │

$ llmfit info Qwen/Qwen3.6-35B-A3B | grep "Baseline Est"
  Baseline Est. Speed: 29.3 tok/s
```

The two commands disagree by 13.2x. The 12x sparsity ratio accounts for nearly all of it.

**Control — `plan` scales purely with total parameters.** Compare the 36.0B MoE model with a dense
32.8B model at the same quant:

```console
$ llmfit --json plan "Qwen/Qwen3-32B" --context 32768 --quant Q8_0 \
    | python3 -c "import json,sys;d=json.load(sys.stdin);print([(p['path'],p['estimated_tps']) for p in d['run_paths']])"
[('gpu', 2.441844179492066), ('cpu_offload', None), ('cpu_only', 0.5127872776933337)]
```

`2.2252 x (36.0 / 32.8) = 2.4423`, against a measured `2.4418`. The MoE model and the dense model
land on one curve, to within 0.03%. `plan` reads the 3.0B active set as 36.0B.

The #805 reproducer also still holds at Q4_K_M:

```console
$ llmfit --json plan "Qwen/Qwen3-30B-A3B-Instruct-2507" --context 32768 --quant Q4_K_M \
    | python3 -c "import json,sys;d=json.load(sys.stdin);print([p['estimated_tps'] for p in d['run_paths'] if p['path']=='gpu'])"
[3.7665248962940887]
```

### Expected behavior

`plan` and `fit` report the same throughput for one model, quant and context on one machine. The
fallback branch reads the active parameter set for a MoE model, exactly as the fallback branch in
`fit.rs` does.

### Actual behavior

`plan` reports 2.23 tok/s where `fit` reports 29.3 tok/s. The estimate tracks total parameters.

### Notes

**The memory numbers are correct.** Only throughput is wrong. `plan --context 8192` gives a
minimum of 22.0 GB VRAM, which is 21.4 GB of Q4_K_M weights plus 0.62 GB of fp16 KV cache. That
matches the on-disk size of the published GGUF quantizations. The KV cache alternatives table is
also sound. A fix should touch the tok/s path only.

**Hardware datapoint for the bandwidth table.** `llmfit info` asks for the GPU model, so here is a
measurement from this host. On 2026-08-16 with `unsloth/Qwen3.6-35B-A3B-GGUF UD-Q5_K_M`
(24.63 GiB, 34.66B params, arch `qwen35moe`), llama.cpp b10068 on the Vulkan backend
(Mesa ANV 26.1.5, `uma:1`, `fp16:1`, no matrix cores):

```
llama-bench -ngl 99 -p 512 -n 128
  pp512 = 127.06 +/- 2.78 t/s
  tg128 =  12.35 +/- 0.57 t/s
```

Two caveats on that figure. The distribution llama.cpp build prints "asserts enabled, performance
may be affected". Mesa Vulkan exposes no XMX matrix cores on this part. A SYCL or oneAPI build
should score higher, so 12.35 t/s is a floor for this GPU, not a ceiling.

Set against that measurement: `plan` at 2.23 tok/s underestimates measured decode by 5.5x, and
`fit` at 29.3 tok/s overestimates it by 2.4x. The estimates are at Q8_0 and the measurement is at
Q5_K_M, so at matched quant the `fit` gap is larger and the `plan` gap is smaller. The `fit` error is a calibration gap of the kind the
bandwidth table exists to close. The `plan` error is structural, because the estimate ignores the
architecture.

### Suggested fix

Resolve `params` in `plan.rs::estimate_tps_with_gpu` the way `fit.rs::estimate_tps` resolves it,
before the delegation check, so both branches see the active set. A regression test that pins
`plan` to `fit` on a MoE model with `system: None` would hold the fallback branch closed.
`plan.rs:1495` already asserts this property on the delegation branch
(`test_moe_gpu_speed_uses_active_params`), which is why the gap survived PR #807.

### Diagnostic report (`llmfit doctor`)

# llmfit doctor report

Paste this whole report into a GitHub issue at https://github.com/AlexsJones/llmfit/issues — the raw tool output below is what lets detection bugs become regression tests. It contains hardware model names and driver info only.

- llmfit version: 1.1.10
- OS: linux (x86_64)

## Detected by llmfit
```
SystemSpecs {
    total_ram_gb: 62.32108688354492,
    available_ram_gb: 51.91744613647461,
    total_cpu_cores: 16,
    cpu_name: "Intel(R) Core(TM) Ultra 7 255H",
    has_gpu: true,
    gpu_vram_gb: Some(
        62.32108688354492,
    ),
    total_gpu_vram_gb: Some(
        62.32108688354492,
    ),
    gpu_available_gb: None,
    gpu_name: Some(
        "Intel Arc Pro 130T/140T (integrated)",
    ),
    gpu_count: 1,
    unified_memory: true,
    backend: Sycl,
    gpus: [
        GpuInfo {
            name: "Intel Arc Pro 130T/140T (integrated)",
            vram_gb: Some(
                62.32108688354492,
            ),
            backend: Sycl,
            count: 1,
            unified_memory: true,
        },
    ],
    cluster_mode: false,
    cluster_node_count: 0,
}
```

## nvidia-smi (extended query)
```
(not available: No such file or directory (os error 2))
```

## nvidia-smi (standard query)
```
(not available: No such file or directory (os error 2))
```

## rocm-smi --showmeminfo vram
```
(not available: No such file or directory (os error 2))
```

## rocm-smi --showproductname
```
(not available: No such file or directory (os error 2))
```

## sysfs DRM cards
```
card1: vendor=0x8086 device=0x7d51 driver=i915 mem_info_vram_total=-
```

## lspci (display controllers)
```
00:02.0 VGA compatible controller [0300]: Intel Corporation Arrow Lake-P [Arc Pro 130T/140T] [8086:7d51] (rev 03)
```

## vulkaninfo --summary
```
(not available: No such file or directory (os error 2))
```

## npu-smi info
```
(not available: No such file or directory (os error 2))
```

## Provider app installs
```
LM Studio installed: false
Docker Desktop installed: false
ollama on PATH: true
```

HsiangNianian:
## Problem

`plan` used total parameters for sparse MoE models when GPU bandwidth was unknown. This made fallback throughput estimates diverge sharply from `fit` for the same model and hardware.

## Change

- Resolve fallback throughput parameters from `active_parameters` for MoE models, matching `fit`.
- Add a regression that forces `system: None` and compares the `plan` and `fit` fallback estimates.
- Leave memory estimates and GPU bandwidth calibration unchanged.

## Test plan

- `cargo test -p llmfit-core plan::tests -- --nocapture`
- `make test`
- `cargo check --all-targets --all-features`
- `cargo clippy --all-targets --all-features`
- `npm ci && npm run build && npm test` in `llmfit-web`
- `uv run pytest -vv -m rust_integration` in `llmfit-python`

Closes #921

### DIFF
--- a/llmfit-core/src/plan.rs
+++ b/llmfit-core/src/plan.rs
@@ -218,7 +218,12 @@ fn estimate_tps_with_gpu(
     use crate::fit::InferenceRuntime;
     use crate::hardware::gpu_memory_bandwidth_gbps;
 
-    let params = model.params_b().max(0.1);
+    let params = model
+        .active_parameters
+        .filter(|_| model.is_moe)
+        .map(|p| (p as f64) / 1_000_000_000.0)
+        .unwrap_or_else(|| model.params_b())
+        .max(0.1);
 
     // Delegate to the shared estimator when we can resolve real GPU bandwidth.
     //
@@ -1513,6 +1518,37 @@ mod tests {
         assert!(
             moe_tps > dense_equivalent * 2.0,
             "MoE estimate {moe_tps} should far exceed dense-equivalent {dense_equivalent}"
+        );
+    }
+
+    /// The fixed-constant fallback must preserve the same sparse-MoE parameter
+    /// accounting as the known-bandwidth path.
+    #[test]
+    fn test_moe_fallback_speed_uses_active_params() {
+        let moe = test_moe_model();
+        let specs = test_specs();
+
+        let plan_tps = estimate_tps_with_gpu(
+            &moe,
+            "Q4_K_M",
+            GpuBackend::Cuda,
+            PlanRunPath::Gpu,
+            8,
+            None,
+            &cfg(),
+        );
+        let fit_tps = crate::fit::estimate_tps(
+            &moe,
+            "Q4_K_M",
+            &specs,
+            RunMode::Gpu,
+            crate::fit::InferenceRuntime::LlamaCpp,
+            &cfg(),
+        );
+
+        assert!(
+            (plan_tps - fit_tps).abs() < 1e-9,
+            "fallback MoE estimates diverged: plan={plan_tps} fit={fit_tps}"
         );
     }
 


AlexsJones (approved):
Nice, this closes the exact branch #807 left open, and the fix matches `fit.rs` line for line.

Hoisting the resolution above the delegation check is the right place for it. The delegation calls `fit::estimate_tps`, which resolves `active_parameters` itself, so the known-bandwidth path is untouched and `test_moe_gpu_speed_uses_active_params` still guards it. The new test then pins the two estimators together on the `system: None` path, which is the property that was missing, and `test_specs()` names `Test GPU`, so the fallback is guaranteed to stay the fallback rather than quietly drifting onto the bandwidth branch later.

Scope is right too: memory estimates and the bandwidth calibration are left alone, which matches what the reporter asked for.

Only Greptile ran on this branch (repo CI does not fire automatically for a first contribution), so the full suite needs a maintainer trigger before merge. Recommending it to Alex for the final look. Closes #921.

PR #927, opened 2026-08-21
fix(providers): detect nested LM Studio model files

### LINKED ISSUE #900: [Bug]: LM Studio installed-detection only sees loaded models (1 of 13)

ketpatil77:
## Summary

- recursively discover GGUF files in LM Studio and llama.cpp model directories
- cap traversal at three directory levels and filter by `.gguf` extension before file-type checks
- merge on-disk LM Studio models with server-reported loaded models without double-counting
- retain disk discovery when the LM Studio server is offline
- add regression tests for nesting, depth limits, extension filtering, aliases, and offline detection

Closes #900

## Verification

- `cargo fmt --check`
- `git diff --check`

## Local test limitation

Rust compilation cannot link in this Windows environment because the MSVC Build Tools installer exits with code `1602`, leaving `link.exe` unavailable. The draft PR's Linux CI is the compile/test authority for the added unit tests.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -861,6 +861,37 @@ impl ModelProvider for MlxProvider {
 // llama.cpp provider (direct GGUF download from HuggingFace)
 // ---------------------------------------------------------------------------
 
+const GGUF_SCAN_MAX_DEPTH: usize = 3;
+
+fn find_gguf_files(root: &Path, max_depth: usize) -> Vec<PathBuf> {
+    let mut files = Vec::new();
+    let mut pending = vec![(root.to_path_buf(), 0usize)];
+
+    while let Some((dir, depth)) = pending.pop() {
+        let Ok(entries) = std::fs::read_dir(dir) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            let path = entry.path();
+            let is_gguf = path
+                .extension()
+                .and_then(|extension| extension.to_str())
+                .is_some_and(|extension| extension.eq_ignore_ascii_case("gguf"));
+            if is_gguf {
+                if entry.file_type().is_ok_and(|file_type| file_type.is_file()) {
+                    files.push(path);
+                }
+                continue;
+            }
+            if depth < max_depth && entry.file_type().is_ok_and(|file_type| file_type.is_dir()) {
+                pending.push((path, depth + 1));
+            }
+        }
+    }
+
+    files
+}
+
 /// A provider that downloads GGUF model files directly from HuggingFace
 /// and uses llama.cpp binaries (`llama-cli`, `llama-server`) to run them.
 ///
@@ -980,16 +1011,7 @@ impl LlamaCppProvider {
 
     /// List all `.gguf` files in the cache directory.
     pub fn list_gguf_files(&self) -> Vec<PathBuf> {
-        let mut files = Vec::new();
-        if let Ok(entries) = std::fs::read_dir(&self.models_dir) {
-            for entry in entries.flatten() {
-                let path = entry.path();
-                if path.extension().and_then(|e| e.to_str()) == Some("gguf") {
-                    files.push(path);
-                }
-            }
-        }
-        files
+        find_gguf_files(&self.models_dir, GGUF_SCAN_MAX_DEPTH)
     }
 
     /// Search HuggingFace for GGUF repositories matching a query.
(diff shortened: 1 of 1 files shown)


ketpatil77:
Closing in favor of #902. The earlier PR already implements nested LM Studio model discovery with broader cross-module integration, 662 passing tests, edge-case follow-ups, green CI, and maintainer approval. Keeping this draft would duplicate active work.

PR #931, opened 2026-08-21
feat: add llmfit self-update check

### LINKED ISSUE #930: Feature request: llmfit self-update mechanism
### Problem

`llmfit` currently has no way to check for or apply new releases of the tool itself. `llmfit update` only refreshes the local model database from HuggingFace — it does not touch the `llmfit` binary. Users must notice a new release manually (via GitHub, the README badge, etc.) and re-run their original install method to upgrade.

Given how many install paths this project supports (Homebrew tap, homebrew-core, MacPorts, `install.sh` prebuilt binary, `uv tool` / `pip`, Docker/Podman, `cargo install` / crates.io, build-from-source), users on the same "latest" install rarely know they're behind, and there's no consistent UX across install methods for closing that gap.

### Proposal

1. **Version check**: on a cadence (e.g. once per day, cached — not on every invocation) or via an explicit command, check the latest published version against the running binary's version.
   - Compare semver, ignoring release-please's `v` tag prefix.
   - Read the latest version from the GitHub Releases API (or an existing lightweight endpoint, mirroring how the project already publishes `CHANGELOG.md`/tags via release-please).
2. **Notify, don't surprise**: when behind, print a one-line notice (TUI: a dismissible banner/footer hint; CLI: a stderr line that doesn't interfere with `--json` output) — never block or auto-mutate without consent.
3. **Install-method-aware upgrade**:
   - Detect how the running binary was installed (e.g. a stamped marker at build time, or heuristics on the binary's path — brew Cellar/Caskroom path, `cargo`'s `~/.cargo/bin`, `uv tool` dir, Docker/container, etc.).
   - For methods llmfit can safely self-manage (e.g. the `install.sh` prebuilt-binary path), offer an explicit `llmfit self-update` (or similar) that downloads and swaps the binary — never automatic/unattended by default.
   - For methods managed by an external package manager (Homebrew, MacPorts, `uv tool`, Docker, cargo), print the exact command the user should run instead (`brew upgrade llmfit`, `port upgrade llmfit`, `uv tool upgrade llmfit`, `docker pull ...`, `cargo install llmfit --force`) rather than trying to self-modify a file it doesn't own.
4. **Respect opt-out**: an env var (e.g. `LLMFIT_NO_UPDATE_CHECK=1`) and/or a config flag to disable the check entirely — useful for CI, Docker, and offline/air-gapped use.
5. **`llmfit doctor` integration**: include current vs. latest version in the existing `llmfit doctor` diagnostic report.

### Non-goals

- No change to `llmfit update`'s existing behavior (model DB refresh) — this is a separate concern from binary self-update; naming should avoid confusion between the two (e.g. `llmfit self-update` vs. `llmfit update`).
- No silent/automatic mutation of the binary without explicit user action.

### Open questions

- Where should the version-check endpoint live — GitHub Releases API directly, or a small hosted endpoint (mirroring `llmfit.axjns.dev`) to keep it lightweight and cacheable?
- Should the daily check be opt-in or opt-out by default?

---
Filed via automated workflow.

take566:
Fixes #930.

## What

Adds a self-update *check* for the `llmfit` binary itself (distinct from `llmfit update`, which refreshes the local model-database cache — different concern, unfortunate name collision noted in the module docs).

- `llmfit-core::self_update` — new module:
  - `parse_version` / version comparison (`major.minor.patch`, ignoring pre-release/build suffixes)
  - `detect_install_method()` — heuristic based on the running executable's own path (Homebrew Cellar/Caskroom, MacPorts, `~/.cargo/bin`, `uv`-managed tool dirs, `/.dockerenv`, or the `install.sh` prebuilt-binary layout), each mapped to the exact upgrade command for that method
  - `fetch_latest_version()` — GitHub releases API lookup
  - `check_for_update()` — live/uncached check
  - `check_for_update_cached()` — same, but rate-limited to ~once/24h via a small cache file under the data dir, for opportunistic use on every invocation
  - `LLMFIT_NO_UPDATE_CHECK` env var to disable entirely
- `llmfit doctor` now reports current vs. latest version (live check, respects the env var)
- Non-`--json`, non-TUI CLI invocations print a one-line stderr notice when behind (gated off for `--json` so it never lands in agent/script output, and off for any path that ends up launching the TUI so it doesn't flash-then-vanish under the alternate screen)
- `docs/cli.md` and `--help` updated

## Deliberately out of scope for this PR

- **No binary self-mutation.** `llmfit` never downloads/swaps its own binary — with this many install methods, silently touching a file owned by a package manager felt like the wrong default. It hands back the exact command instead.
- **No TUI banner.** The ratatui rendering pipeline (`tui_app.rs`/`tui_ui.rs`) would need a footer/banner slot for this, which felt like a separate, more involved change to get right — happy to follow up if that's wanted, or if maintainers would rather it land together.

## Testing

`llmfit-core/src/self_update.rs` has unit tests for version parsing/ordering, install-method detection from synthetic paths, and the notice-formatting logic (including the "lookup failed" and "up to date" no-notice cases).

**I could not run `cargo build`/`test`/`fmt`/`clippy` locally** — no Rust toolchain available in the environment I wrote this in. I read the code carefully against the existing `ureq`/`serde` patterns in `update.rs` and `dirs`-based cache-file pattern already used there, but please treat CI as the first real compile of this and let me know if anything needs adjusting — happy to fix up formatting/clippy nits.

## Open questions from the issue, resolved for v1

- Version-check endpoint: GitHub releases API directly (simplest, no new infra).
- Daily check: opt-out by default (env var to disable), not opt-in — matches most CLI tools' update-nag UX.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -223,4 +223,12 @@ llmfit plan "Qwen/Qwen2.5-Coder-0.5B-Instruct" --context 8192 --json
 - per-path feasibility (`gpu`, `cpu_offload`, `cpu_only`)
 - upgrade deltas
 
+### Update check
+
+On non-`--json`, non-TUI invocations, llmfit opportunistically checks GitHub for a newer release (cached for ~24h, network failures are silent) and prints a one-line notice plus the exact command to upgrade for how it detects you installed it (`brew upgrade llmfit`, `uv tool upgrade llmfit`, etc.) — it never downloads or replaces the binary itself. `llmfit doctor` always does a live (uncached) check and includes the result in its report.
+
+This is unrelated to `llmfit update`, which refreshes the local *model database* cache, not the `llmfit` binary.
+
+Set `LLMFIT_NO_UPDATE_CHECK=1` to disable the check entirely (CI, Docker, air-gapped hosts).
+
 ---
--- a/llmfit-core/src/doctor.rs
+++ b/llmfit-core/src/doctor.rs
@@ -114,9 +114,24 @@ pub fn collect_diagnostics(version: &str) -> String {
          below is what lets detection bugs become regression tests. It \
          contains hardware model names and driver info only.\n"
     );
+    let latest_line = if crate::self_update::background_check_disabled() {
+        "skipped (LLMFIT_NO_UPDATE_CHECK set)".to_string()
+    } else {
+        let version_check = crate::self_update::check_for_update(version);
+        match &version_check.latest {
+            Some(latest) if version_check.is_behind => {
+                format!(
+                    "{latest} available — upgrade with `{}`",
+                    version_check.install_method.upgrade_hint()
+                )
+            }
+            Some(latest) => format!("{latest} (up to date)"),
+            None => "unknown (couldn't reach GitHub — offline or rate-limited)".to_string(),
+        }
+    };
     let _ = writeln!(
         report,
-        "- llmfit version: {version}\n- OS: {} ({})\n",
+        "- llmfit version: {version}\n- Latest release: {latest_line}\n- OS: {} ({})\n",
         std::env::consts::OS,
         std::env::consts::ARCH
     );
--- a/llmfit-core/src/lib.rs
+++ b/llmfit-core/src/lib.rs
@@ -9,6 +9,7 @@ pub mod models;
 pub mod plan;
 pub mod providers;
 pub mod quality;
+pub mod self_update;
 pub mod share;
 pub mod task_bench;
 pub mod update;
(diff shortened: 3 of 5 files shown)


take566:
Update: I've since set up a Rust toolchain and verified locally —

```
cargo build -p llmfit-core -p llmfit   # 0 errors
cargo test -p llmfit-core self_update   # 13 passed
cargo fmt --check                       # clean
cargo clippy -p llmfit-core -p llmfit --no-deps   # no warnings from this PR's code (fixed 2 collapsible_if lints in main.rs)
```

Pushed the fmt/clippy fixes (force-push, same branch). Should be a clean CI run now.

PR #937, opened 2026-08-23
docs: place --memory before the recommend subcommand in Termux/Android examples

### LINKED ISSUE #932: [Docs]: --memory must precede the recommend subcommand in the Termux/Android examples
### Documentation area

README.md (installation, usage) — and docs/platform-support.md

### Current state

The Termux/Android "manual memory override" example places the global `--memory` flag *after* the `recommend` subcommand, so the documented command fails immediately when run:

```sh
# docs/platform-support.md:34 · README.zh.md:910 · README.ja.md:901
llmfit recommend --json --memory=8G --limit 10
# → error: unexpected argument '--memory' found   (exit 2)
```

`--memory` (and its sibling global overrides `--ram`, `--cpu-cores`, `--max-context`) are root-level flags and must appear *before* the subcommand — as the same code block's other line already shows:

```sh
llmfit --memory=8G fit -n 20                      # flag-first → works
llmfit --memory=8G recommend --json --limit 10   # flag-first → works
```

I reproduced the failure with the current `main` binary (no GPU/hardware required). The identical incorrect example exists in three files that carry the same copy.

### Suggested improvement

Change all three occurrences to the flag-first form:

```sh
llmfit --memory=8G recommend --json --limit 10
```

### Would you be willing to submit a PR for this?

Yes

junnhwan:
## What & why

Fixes #932.

The Termux/Android "manual memory override" example places the global `--memory` flag *after* the `recommend` subcommand, so the documented command fails immediately:

```sh
llmfit recommend --json --memory=8G --limit 10
# → error: unexpected argument '--memory' found   (exit 2)
``"

`--memory` is a root-level flag and must appear *before* the subcommand — as the sibling `fit` line in the same code block already shows (`llmfit --memory=8G fit -n 20`).

## Changes

Update the identical example to the flag-first form in all three files that share the same copy:

- `docs/platform-support.md`
- `README.zh.md`
- `README.ja.md`

```diff
-llmfit recommend --json --memory=8G --limit 10
+llmfit --memory=8G recommend --json --limit 10
```

Verified the failure reproduces on `main` with the current `recommend` form and that the flag-first form succeeds. Docs-only change, no code or tests affected.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -898,7 +898,7 @@ llmfit のデータベースは HuggingFace のモデル名（例: `Qwen/Qwen2.5
 
 ```sh
 llmfit --memory=8G fit -n 20
-llmfit recommend --json --memory=8G --limit 10
+llmfit --memory=8G recommend --json --limit 10
 ```
 
 これは推奨/スコアリングのみのための回避策であり、真の Android GPU ランタイム検出を提供するものではありません。
--- a/README.zh.md
+++ b/README.zh.md
@@ -907,7 +907,7 @@ llmfit 的数据库使用 HuggingFace 模型名称（例如 `Qwen/Qwen2.5-Coder-
 
 ```sh
 llmfit --memory=8G fit -n 20
-llmfit recommend --json --memory=8G --limit 10
+llmfit --memory=8G recommend --json --limit 10
 ```
 
 这仅是推荐/评分的变通方案；不提供真正的 Android GPU 运行时检测。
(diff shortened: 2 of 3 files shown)


AlexsJones (approved):
Thanks for fixing every copy of the example. I checked the three docs changes against the root-level CLI flag contract; each now places `--memory` before `recommend`, and the diff is limited to that correction. Ready for your final look.

PR #948, opened 2026-08-24
feat(mcp): report catalog provenance to external model consumers

udiedrichsen:
## What

Adds catalog provenance to llmfit's machine-readable output: a `catalog` block that states what the model list actually is, how old it is, and — explicitly — that it is not a live registry view.

```json
"catalog": {
  "catalog_models": 11271,
  "llmfit_version": "1.1.10",
  "embedded_catalog_pinned_to_build": true,
  "update_cache_present": false,
  "update_cache_models": 0,
  "update_cache_updated_unix_s": null,
  "update_cache_age_days": null,
  "is_live_registry_view": false,
  "note": "Snapshot, not a live registry view. The catalog is the build-time embedded list plus user custom models plus models appended by `llmfit update`; the update cache never refreshes metadata for models that are already embedded. Run `llmfit update` to append newly released models."
}
```

It is emitted once per response envelope, not per model, because freshness is a property of the catalog rather than of any single entry. All three surfaces get it from one place (`serve_shared::catalog_json`): CLI `--json`, the HTTP API, and the MCP tools.

The MCP tool descriptions are also rewritten to carry a grounding contract — name only models present in the response, keep `measured_tps` and `estimated_tps` distinct, read `catalog` before calling the list current, and treat an empty search result as "llmfit has no such model" rather than "no such model exists".

## Why

llmfit already ships `skills/llmfit-advisor/SKILL.md`, so "an external LLM reads llmfit's JSON and advises the user" is an established pattern in this repo. The problem is that the JSON never says how old the catalog is, and a model asked "is this list current?" will not answer "unknown" — it will fill the gap.

That is not a hypothetical. I built upstream `6a2e660` in a throwaway worktree and ran the same prompts through the same local model at the same temperature against both builds. The upstream build produced:

> "The catalog is served by the backend that powers these tools, and it is refreshed periodically (not in real time)."

There is no backend, and nothing refreshes — the update cache is append-only and never revises metadata for models that are already embedded. Every clause is invented. The patched build reported the actual state instead: build-time embedded catalog, no update cache present, age unknown because `llmfit update` had never run.

The fix is not a better prompt. It is that the payload contained no ground truth for the question, so the only difference between a grounded answer and a fabricated one was luck.

## Relationship to #935

This is deliberately the opposite design. #935 put an inference client inside llmfit; the objection there was the trust boundary, and it was a fair one. Here llmfit stays what it is — a local measurement and catalog tool — and does nothing but describe its own evidence more honestly. No outbound network, no new runtime dependencies, no inference in-process. Whatever model consumes this runs entirely on the user's side of the boundary, via MCP or via the existing skill.

## Implementation notes

- `update::cache_status()` reports presence, model count and mtime of the update cache; `CacheStatus::age_days()` derives the age.
- Age uses `checked_sub`, so a future-dated cache file yields `None` rather than `0`. Clock skew must not be able to make a stale catalog look fresh.
- A catalog with no update cache reports `null` for age and timestamp, not `0` — "unknown" and "zero days old" are different claims and a consumer must be able to tell them apart.
- `total_models` (models matching this query) and `catalog_models` (models llmfit knows about) are kept as separate names; they genuinely differ — 8970 vs 11271 in a live run here.
- `API.md` documents the block field by field; `skills/llmfit-advisor/SKILL.md` gains a catalog-freshness section so the existing skill path benefits too.

## Tests

8 new tests: cache age in whole days, age unknown on clock skew, age unknown without a timestamp, fresh-cache reporting, build-time-only catalog reporting nulls rather than zeros, never claiming a live view, version passthrough, and MCP tools carrying the catalog block.

One pre-existing failure is unrelated to this change: `tui_app::tests::changing_search_query_resets_selection_to_top` fails on my machine on clean `6a2e660` as well (verified by stashing). It appears hardware-dependent. I have deliberately not touched it.

### DIFF
--- a/API.md
+++ b/API.md
@@ -337,6 +337,54 @@ Add to your MCP client config (e.g. `claude_desktop_config.json`):
 | `get_runtimes` | Installed inference runtimes | None |
 | `get_installed_models` | Models in local runtimes | None |
 
+`recommend_models` and `search_models` return a `catalog` block alongside their
+results (see [Catalog provenance](#catalog-provenance)). Their descriptions ask
+the calling agent to recommend only models present in the response and to keep
+`measured_tps` and `estimated_tps` distinct — llmfit cannot enforce this the way
+an in-process check could, but it can state the contract where the model reads
+it.
+
+---
+
+## Catalog provenance
+
+Every response that returns a model list — CLI `--json`, the HTTP API, and the
+MCP model tools — carries a `catalog` block describing the list it was computed
+from:
+
+```json
+{
+  "catalog": {
+    "catalog_models": 1247,
+    "llmfit_version": "1.1.10",
+    "embedded_catalog_pinned_to_build": true,
+    "update_cache_present": true,
+    "update_cache_models": 63,
+    "update_cache_updated_unix_s": 1755950000,
+    "update_cache_age_days": 3,
+    "is_live_registry_view": false,
+    "note": "Snapshot, not a live registry view. ..."
+  }
+}
+```
+
+| Field | Meaning |
+|---|---|
+| `catalog_models` | Size of the catalog the response was computed from. Named apart from the surrounding `total_models`, which counts models matching the query |
+| `llmfit_version` | Build the embedded catalog is pinned to |
+| `embedded_catalog_pinned_to_build` | Always `true`; the embedded list is fixed at compile time |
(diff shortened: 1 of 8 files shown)


udiedrichsen:
I closed my previous pr https://github.com/AlexsJones/llmfit/pull/935

PR #950, opened 2026-08-24
fix(ollama): map the gemma3 family sizes to their catalog ids

### LINKED ISSUE #866: [Bug]: gemma3:4b does not resolve to google/gemma-3-4b-it in installed detection

Akciali:
Closes #866. Layer 1 as requested there: the three missing explicit mappings, no heuristic change.

`gemma3:{1b,4b,27b}` now resolve to `google/gemma-3-{1b,4b,27b}-it` like the 12B already did, which restores both installed detection and the Ollama download gate for the family. The regression test walks all four sizes end to end: candidates, mapping presence, and detection against an installed set built from the pulled tag.

The long-tail candidate rule (hyphen-before-digit collapse, 22 entries measured in the issue) stays out, per the review guidance.

### DIFF
--- a/llmfit-core/src/providers.rs
+++ b/llmfit-core/src/providers.rs
@@ -3832,7 +3832,10 @@ const OLLAMA_MAPPINGS: &[(&str, &str)] = &[
     ("codellama-13b-instruct-hf", "codellama:13b"),
     ("codellama-7b-instruct-hf", "codellama:7b"),
     // Google Gemma
+    ("gemma-3-27b-it", "gemma3:27b"),
     ("gemma-3-12b-it", "gemma3:12b"),
+    ("gemma-3-4b-it", "gemma3:4b"),
+    ("gemma-3-1b-it", "gemma3:1b"),
     ("gemma-2-27b-it", "gemma2:27b"),
     ("gemma-2-9b-it", "gemma2:9b"),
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Thanks for keeping this to the missing family mappings. I checked that the 1B, 4B, 12B, and 27B tags are covered end to end, including candidate lookup, mapping presence, and installed detection. The long-tail heuristic stays out of scope as intended. CI is green; ready for your final look.

PR #966, opened 2026-08-29
fix(cli): return structured JSON errors for missing models

### LINKED ISSUE #901: [Bug]: --json error paths return plain text and inconsistent exit codes
### Bug description

The `--json` global flag documents a contract that the error paths do not keep, and the subcommands disagree with each other about what failure looks like.

From `llmfit --help`:

```
  --json             Output structured JSON on every subcommand (for tool/agent
                     integration). Always exits 0 on success, 1 on error.

EXIT CODES:
  0  Success
  1  Any error (hardware detection failure, model not found, network error, etc.)
```

"Model not found" is named there, so it is meant to be an error.

### Actual behavior

Asking three subcommands about a model that does not exist, all with `--json`:

| command | exit code | valid JSON |
|---|---|---|
| `info does-not-exist-xyz` | 0 | no |
| `search zzzzznomatch` | 0 | no |
| `diff does-not-exist-xyz also-not-real` | 1 | no |

`info` prints `No model found matching 'does-not-exist-xyz'` as plain text and reports success. `diff` gets the exit code right and still answers in plain text.

### Expected behavior

With `--json`, a failure exits 1 and prints a JSON object describing it, consistently across subcommands.

### Steps to reproduce

```sh
llmfit --json info "does-not-exist-xyz";    echo "exit=$?"
llmfit --json search "zzzzznomatch";        echo "exit=$?"
llmfit --json diff "does-not-exist-xyz" "also-not-real"; echo "exit=$?"
```

### Why it matters

This is the agent-facing path, and `--help` has a dedicated AGENT USAGE section for it. A caller that checks the exit code is told everything worked, and a caller that pipes to `jq` gets a parse error instead of a message it can read. Right now the only reliable way to detect failure is to try parsing the output and treat unparseable text as an error.

`search` may be arguable: it falls back to searching HuggingFace, so exit 0 could be intended even when nothing local matched. The output still is not JSON though.

There is no existing error object in the JSON output to copy, so I do not want to invent a shape you will not want. Something like:

```json
{ "error": { "kind": "model_not_found", "message": "No model found matching 'does-not-exist-xyz'" } }
```

Would you want a machine-readable `kind` alongside the message, or just the message? And should `search` finding nothing locally count as an error at all, given the HuggingFace fallback?

Related but not the same: #803 covers the vocabulary used for values in JSON output, on the success path.

### llmfit version

1.1.10

### Operating system

### Affected component

CLI (`--json` output and exit codes)

mikemikimike:
Fixes #901

## What changed

When `info` or `diff` cannot resolve a model selector, `--json` now emits a valid structured error containing a machine-readable `kind` and a human-readable `message`, and the command exits with status 1. The normal text output remains unchanged. Ambiguous selectors use `invalid_selector`, while missing models use `model_not_found`.

## Validation

- `cargo test -p llmfit`: 83 unit tests and 11 CLI integration tests passed.
- `cargo check -p llmfit --all-targets` passed.
- `cargo fmt --all -- --check` passed.
- `git diff --check` passed.
- `cargo clippy -p llmfit --all-targets` completed with pre-existing warnings; strict `-D warnings` remains blocked by unrelated existing lints in `llmfit-core`.

### DIFF
--- a/llmfit-tui/src/display.rs
+++ b/llmfit-tui/src/display.rs
@@ -512,6 +512,20 @@ pub fn display_json_fits(specs: &SystemSpecs, fits: &[ModelFit]) {
     );
 }
 
+/// Serialize a CLI error for machine consumers.
+pub fn display_json_error(kind: &str, message: &str) {
+    let output = serde_json::json!({
+        "error": {
+            "kind": kind,
+            "message": message,
+        },
+    });
+    println!(
+        "{}",
+        serde_json::to_string_pretty(&output).expect("JSON serialization failed")
+    );
+}
+
 /// Serialize system specs + model fits to JSON with llama.cpp commands and print to stdout.
 pub fn display_json_fits_with_llamacpp(specs: &SystemSpecs, fits: &[ModelFit]) {
     use llmfit_core::fit::InferenceRuntime;
--- a/llmfit-tui/src/main.rs
+++ b/llmfit-tui/src/main.rs
@@ -1207,6 +1207,14 @@ fn find_fit_index_by_selector(fits: &[ModelFit], selector: &str) -> Result<usize
     find_name_index_by_selector(fits, selector, |fit| fit.model.name.as_str())
 }
 
+fn selector_error_kind(message: &str) -> &'static str {
+    if message.starts_with("No model found") {
+        "model_not_found"
+    } else {
+        "invalid_selector"
+    }
+}
+
 fn run_diff(
     model_a: Option<String>,
     model_b: Option<String>,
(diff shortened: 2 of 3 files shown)


AlexsJones (approved):
Thanks for this, clean fix. The `display_json_error` helper keeps the human path on stderr and gives `--json` consumers a stable `{error:{kind,message}}` shape, and the `model_not_found` vs `invalid_selector` split is exactly what an agent would branch on. Tests cover both the `info` and `diff` not-found paths. Approving for your final look.

PR #975, opened 2026-08-30
fix(quality): disable Ollama thinking for benchmarks

### LINKED ISSUE #960: [Bug]: quality bench never sends think:false, so thinking models return empty responses and score 0 (37% of structured-output runs)
### Bug description

`quality_ollama_generate()` (`llmfit-core/src/quality.rs:175`) posts to `/api/generate` without setting `"think": false`:

```rust
let body = serde_json::json!({
    "model": model,
    "prompt": prompt,
    "stream": false,
    "options": {
        "num_predict": max_tokens,
        "temperature": temperature,
    }
});
```

Ollama models with thinking enabled by default (Qwen3.x, Gemma 4, Nemotron 3, GLM, Granite 4.x …) spend the whole `num_predict` budget inside the reasoning block, and the `response` field comes back **empty**. Several tests in `benchmarks.yaml` set `max_tokens: 128`, which is far below what a thinking model needs to finish reasoning *and* answer — so the model is scored 0 for an answer it was never given room to produce.

### Expected behavior

The benchmark measures whether the model can produce the requested output.

### Actual behavior

It measures whether the model happens to stop thinking within 128 tokens. In a full run against 12 local models on Ollama, **37% of all `structured-output` responses came back as the empty string** (`22/60`), and 15% of `tool-calling` (`9/60`). Those all score 0.

It also makes the benchmark non-reproducible: the same model under three different Ollama tags scored 0.8 / 1.8 / 2.4 on `structured-output`, because how much it thinks varies per prompt.

### Steps to reproduce

Same model, same prompt (`structured-output/json_extraction`, `max_tokens: 128`), only `think` differs:

```python
import json, urllib.request
p = '''Extract the following information from this text and return ONLY valid JSON, no other text:

"John Smith, age 34, works at Acme Corp as a Senior Engineer. He can be reached at john@acme.com or 555-0123."

Required JSON format: {"name": "", "age": 0, "company": "", "title": "", "email": "", "phone": ""}'''

for think in (True, False):
    b = {"model": "nemotron-3-nano:4b", "prompt": p, "stream": False,
         "options": {"num_predict": 128, "temperature": 0.0}}
    if not think:
        b["think"] = False
    r = json.load(urllib.request.urlopen(urllib.request.Request(
        "http://127.0.0.1:11434/api/generate", json.dumps(b).encode(),
        {"Content-Type": "application/json"})))
    print(think, r["eval_count"], repr(r["response"])[:120])
```

```
True   128 ''
False   54 '{"name": "John Smith", "age": 34, "company": "Acme Corp", "title": "Senior Engineer", "email": "john@acme.com", "phone": "555-0123"}'
```

The `think=False` answer matches every content rule in that test. llmfit scores the `think=True` run **0.0**.

### Suggested fix

Two independent changes, both small:

1. Send `"think": false` in `quality_ollama_generate` (and the equivalent `reasoning_effort`/`enable_thinking: false` knob for the OpenAI-compatible path), at least by default, with an opt-in flag for benchmarking thinking behaviour deliberately.
2. Raise the `max_tokens` floor for quality tests, or subtract the thinking tokens from the budget. Even with thinking on, 128 tokens is not enough for these models to reason and answer.

A useful guard would be to surface empty responses as an explicit error rather than a 0 score — right now a systematic harness failure is indistinguishable from a model that genuinely failed the task.

### Impact

Combined with #959, the `structured-output` role reports no usable signal: 12 local models, none above 3.2/10. After fixing both locally (un-escaping the patterns and sending `think: false` with a 512-token budget), the same models and the same tests give 7.2–8.4 with **zero empty responses** and identical scores across three repeat runs.

### Diagnostic report

```text
llmfit 1.1.11 — Linux (Ubuntu 26.04), AMD Ryzen 9 7950X (32 threads), 60.95 GB RAM
GPU: NVIDIA GeForce RTX 3090, 24 GB, CUDA backend
Ollama 0.32.12
```

### Version

1.1.11

mikemikimike:
Fixes #960

## Summary

Ollama models with thinking enabled by default can consume the entire `num_predict` budget in reasoning, leaving an empty `response` for quality benchmarks. This resubmits the focused fix from #968 after that pull request was closed without merging.

- Send `"think": false` for Ollama quality benchmark requests.
- Keep request construction in a small helper so the payload is directly testable.
- Add regression coverage for the request flag and empty-response scoring behavior.

## Implementation

The Ollama quality path now explicitly disables thinking while preserving the existing prompt, token, temperature, and non-streaming request settings. OpenAI-compatible reasoning controls remain outside this focused fix.

## Validation

- `cargo test -p llmfit-core --quiet` — 590 passed, 0 failed, 1 ignored; integration suites: 6 passed and 2 passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy -p llmfit-core --all-targets` — passed with existing warnings
- `git diff --check` — passed

The strict `cargo clippy ... -- -D warnings` variant remains blocked by pre-existing warnings outside this change, including the existing `manual_clamp` warning in `quality.rs`.

## Compatibility

This changes only Ollama quality benchmark requests and makes their behavior deterministic for models that enable thinking by default. It does not change the public API or OpenAI-compatible provider behavior.

### DIFF
--- a/llmfit-core/src/quality.rs
+++ b/llmfit-core/src/quality.rs
@@ -172,6 +172,24 @@ pub struct InferenceResponse {
     pub wall_time_sec: f64,
 }
 
+fn ollama_request_body(
+    model: &str,
+    prompt: &str,
+    max_tokens: u32,
+    temperature: f64,
+) -> serde_json::Value {
+    serde_json::json!({
+        "model": model,
+        "prompt": prompt,
+        "stream": false,
+        "think": false,
+        "options": {
+            "num_predict": max_tokens,
+            "temperature": temperature,
+        }
+    })
+}
+
 pub fn quality_ollama_generate(
     url: &str,
     model: &str,
@@ -179,15 +197,7 @@ pub fn quality_ollama_generate(
     max_tokens: u32,
     temperature: f64,
 ) -> Result<InferenceResponse, String> {
-    let body = serde_json::json!({
-        "model": model,
-        "prompt": prompt,
-        "stream": false,
-        "options": {
-            "num_predict": max_tokens,
-            "temperature": temperature,
-        }
-    });
+    let body = ollama_request_body(model, prompt, max_tokens, temperature);
 
     let start = Instant::now();
     let resp = ureq::post(url)
@@ -773,6 +783,26 @@ mod tests {
         }];
         // Negative score clamped to 0
         assert!((evaluate_response("bad", &rules) - 0.0).abs() < 0.01);
+    }
+
+    #[test]
+    fn ollama_quality_requests_disable_thinking() {
+        let body = ollama_request_body("test-model", "Return JSON", 128, 0.3);
+
+        assert_eq!(body["think"], false);
+        assert_eq!(body["options"]["num_predict"], 128);
+    }
+
+    #[test]
+    fn empty_quality_response_does_not_match_scoring_rules() {
+        let rules = vec![ScoringRule {
+            pattern: "expected answer".to_string(),
+            weight: 10,
+            negate: false,
+            case_insensitive: false,
+        }];
+
+        assert_eq!(evaluate_response("", &rules), 0.0);
     }
 
     #[test]


AlexsJones (approved):
Thanks for resubmitting this after #968. Sending `"think": false` in the shared request helper, with a regression test for the flag and for empty-response scoring, targets #960 directly. Approving for your final look.

mikemikimike:
Thanks for the review and approval. I’ve done a final pass over the requested behavior and validation results, and no further changes are needed from my side.

PR #991, opened 2026-09-02
Fix Score sort direction floating unrunnable models to the top

psylsph:
## Problem

In the TUI, toggling the **Score** column to ascending put models that cannot run on the system (`TooTight` fit) at the **top** of the list, and installed models at the bottom — the opposite of the documented invariants.

### Root cause

`App::re_sort()` implemented ascending sort by reversing the entire ranked vector:

```rust
if self.sort_ascending {
    sorted.reverse();
}
```

The core ranker guarantees two invariants *in addition* to the key order:

1. `TooTight` models always sort last, regardless of column
2. Installed models sort first (when `installed_first` is on)

Reversing the whole list flipped those pins along with the key. Since `sort_ascending` is persisted to `~/.config/llmfit/filters.json`, the broken order could also survive app restarts.

Reproduced with a test before the fix — toggling Score ascending produced:

```
["b-tootight", "c-mid", "a-high"]   // wrong
["c-mid", "a-high", "b-tootight"]   // expected
```

## Fix

- **`llmfit-core/src/fit.rs`**: add `rank_models_by_fit_opts_col_dir()`, which applies the direction *only to the sort key* inside the comparator (via a `dir()` helper), so installed-first and TooTight-last hold in both directions. "No release date" entries also stay last in both directions (missing value, not a key).
- The existing `rank_models_by_fit_opts_col()` delegates with the default direction, so **CLI, MCP, and Web API behavior are unchanged**.
- **`llmfit-tui/src/tui_app.rs`**: `re_sort()` now calls the direction-aware function instead of `reverse()`.
- Drive-by: `clear_persisted_filters()` didn't reset the quant / run-mode / capability toggles (which also load from `filters.json`), causing a pre-existing test failure on machines with a real filter config (`changing_search_query_resets_selection_to_top` failed on clean `main`). Now completed.

## Testing

- Added `score_sort_ascending_keeps_too_tight_last` and `score_sort_keeps_installed_first_in_both_directions` (TUI), plus `test_ascending_sort_keeps_too_tight_last` (core).
- Full suite green: 680 core + 102 TUI + integration tests, including the previously failing pre-existing test.
- `cargo fmt` / `cargo clippy` clean — no new warnings.
- CLI `fit` output re-verified as correctly score-sorted.

### DIFF
--- a/llmfit-core/src/fit.rs
+++ b/llmfit-core/src/fit.rs
@@ -1141,6 +1141,28 @@ pub fn rank_models_by_fit_opts_col(
     installed_first: bool,
     sort_column: SortColumn,
 ) -> Vec<ModelFit> {
+    rank_models_by_fit_opts_col_dir(models, installed_first, sort_column, false)
+}
+
+/// Like [`rank_models_by_fit_opts_col`] with an explicit sort direction.
+///
+/// `ascending = false` keeps each column's default direction (Score, tok/s,
+/// Params, Mem%, Ctx: best value first; Date: newest first; UseCase,
+/// Provider: A→Z). `ascending = true` reverses only the sort key — installed
+/// models stay first (when requested) and TooTight entries stay last in both
+/// directions, so toggling direction cannot float unrunnable models to the
+/// top of the list.
+pub fn rank_models_by_fit_opts_col_dir(
+    models: Vec<ModelFit>,
+    installed_first: bool,
+    sort_column: SortColumn,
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for keeping this focused. I checked that sort direction now reverses only the selected key while preserving installed-first and TooTight-last invariants, with core and TUI regressions. Ready for your final look.

Issue #964, opened 2026-08-28
Integrated GPU (unified memory) reports gpu_vram_gb equal to full system RAM, not a real allocatable figure


sean-abbott:
## Summary

On a CPU-only machine with no discrete GPU, `llmfit doctor` detects the CPU's integrated
graphics (Intel UHD Graphics 630, Coffee Lake) and reports its VRAM as identical to total
system RAM, rather than a realistic allocatable figure for a unified-memory iGPU. This
makes `has_gpu: true` and a ~47GB "GPU" figure misleading for sizing recommendations,
since this iGPU has no meaningful compute for LLM inference and nothing in practice uses
it (confirmed via the actual inference runtime reporting 100% CPU for every load).

## Environment

- llmfit version: 1.1.12
- OS: linux (x86_64)
- CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (Coffee Lake, no discrete GPU installed)

## Raw doctor output

```
SystemSpecs {
    total_ram_gb: 46.879173278808594,
    available_ram_gb: 45.26317596435547,
    total_cpu_cores: 8,
    cpu_name: "Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz",
    has_gpu: true,
    gpu_vram_gb: Some(
        46.879173278808594,
    ),
    total_gpu_vram_gb: Some(
        46.879173278808594,
    ),
    gpu_available_gb: None,
    gpu_name: Some(
        "Intel CoffeeLake-S GT2 [UHD Graphics 630] (integrated)",
    ),
    gpu_count: 1,
    unified_memory: true,
    backend: Sycl,
    gpus: [
        GpuInfo {
            name: "Intel CoffeeLake-S GT2 [UHD Graphics 630] (integrated)",
            vram_gb: Some(
                46.879173278808594,
            ),
            backend: Sycl,
            count: 1,
            unified_memory: true,
        },
    ],
    cluster_mode: false,
    cluster_node_count: 0,
}
```

## Expected behavior

For a `unified_memory: true` integrated GPU, `gpu_vram_gb`/`total_gpu_vram_gb` shouldn't
just echo `total_ram_gb` as if it were separately allocatable capacity on top of system
RAM -- at minimum this double-counts the same physical memory as both "RAM" and "VRAM"
in any downstream fit calculation. It may also be worth reconsidering whether a weak,
old integrated GPU like UHD 630 should set `has_gpu: true` at all for model-fit
[...]

Issue #810, opened 2026-07-28
VRAM detected as 4.0 GB regardless of BIOS UMA setting on Ryzen 7 8845HS / Radeon 780M


lecduchesne:
### Bug description

On a Ryzen 7 8845HS (Hawk Point) with integrated Radeon 780M Graphics, llmfit always reports 4.0 GB VRAM in the TUI header, regardless of the actual GPU UMA frame buffer size 

[ISSUE_vram_detection_780m.md](https://github.com/user-attachments/files/30477919/ISSUE_vram_detection_780m.md)

configured in the BIOS.

### Expected behavior

Should report ~8 GB VRAM, matching the BIOS-configured UMA frame buffer size.

### Actual behavior

llmfit reports 4.0 GB regardless of the BIOS UMA setting (tested at both 8 GB and 16 GB in BIOS — always shows 4.0 GB).

### Steps to reproduce

1. Set BIOS UMA frame buffer size to 8 GB (or any value > 4 GB) on a Ryzen 7 8845HS / Radeon 780M system
2. Run `llmfit` (TUI) or `llmfit doctor`
3. Observe GPU VRAM reported as 4.0 GB instead of the configured value

### Diagnostic report (`llmfit doctor`)

```text
llmfit doctor
```

### llmfit version

1.1.6

### Operating system

Windows

### Affected component

Hardware detection (RAM/CPU/GPU)

### Mode used

TUI (default)

### GPU details (if relevant)

```text
N/A (Windows/AMD — no nvidia-smi/rocm-smi). DXGI output instead:
Dedicated Memory: 8448 MB
Shared Memory: 12197 MB
Display Memory: 20645 MB

WMI (buggy, capped): Win32_VideoController.AdapterRAM = 4293918720 bytes (~4.0 GB)
```

AlexsJones:
Thanks for the thorough report, the DXGI numbers you included pinpoint it. Confirmed as a bug: on Windows llmfit reads VRAM from WMI `Win32_VideoController.AdapterRAM`, which is a 32-bit field capped at ~4 GB, so it can never report your 8 GB carveout. There is an override that substitutes the real memory pool, but it currently only triggers for "Ryzen AI MAX" unified-memory APUs (`hardware.rs`, `is_amd_unified_memory_apu`), so Hawk Point / 780M systems fall through to the capped value. That's why BIOS changes make no difference.

[...]

Issue #609, opened 2026-06-06
Intel Arc Pro B70 not recognized


nmapx:
It seems Intel Arc GPUs are not supported (yet?). I have all the drivers installed (oneAPI) - successfully running models with SYCL/vllm, so it's not a driver-kind issue.
```
GPU: none (CPU (x86))
```

AlexsJones:
Thanks @nmapx. Per your output (`GPU: none (CPU (x86))`) Intel Arc discrete GPUs aren't being detected — a real gap rather than a config problem on your side, since SYCL/vllm works. To make sure a fix covers your setup:

1. OS and driver stack — Linux with `xe`/`i915`, or Windows?
2. Output of `sycl-ls` or `clinfo` if handy, so we know which enumeration path is available.

Arc support would need its own detection path (Level Zero / SYCL rather than CUDA/ROCm). Related: #303 covers discrete-vs-integrated selection once detection exists.

🤝 _Handled by Alex's [Repo Steward](https://github.com/AlexsJones/repo-steward) — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)._

Learn more or run your own: https://github.com/AlexsJones/repo-steward

nmapx:
@AlexsJones

1. `xe` on Ubuntu 24.04 kernel 6.17.0-1009-intel
2. `sycl-ls` output below
```
[level_zero:gpu][level_zero:0] Intel(R) oneAPI Unified Runtime over Level-Zero V2, Intel(R) Arc(TM) Pro B70 Graphics 20.2.0 [1.15.38646+6]
[level_zero:gpu][level_zero:1] Intel(R) oneAPI Unified Runtime over Level-Zero V2, Intel(R) Arc(TM) Pro B70 Graphics 20.2.0 [1.15.38646+6]
[opencl:cpu][opencl:0] Intel(R) OpenCL, AMD Ryzen 9 3900X 12-Core Processor             OpenCL 3.0 (Build 0) [2026.21.6.0.17_160000]
[opencl:gpu][opencl:1] Intel(R) OpenCL Graphics, Intel(R) Arc(TM) Pro B70 Graphics OpenCL 3.0 NEO  [26.22.38646.6]
[opencl:gpu][opencl:2] Intel(R) OpenCL Graphics, Intel(R) Arc(TM) Pro B70 Graphics OpenCL 3.0 NEO  [26.22.38646.6]
```

nmapx:
[...]

PR #993, opened 2026-09-03
fix(hardware): keep Linux BIOS UMA carve-out VRAM on Ryzen AI MAX

AlexsJones:
## Summary
Fixes the Linux half of [#964](https://github.com/AlexsJones/llmfit/issues/964) using [this dump](https://github.com/AlexsJones/llmfit/issues/964#issuecomment-5522910096):

- `rocm-smi` / sysfs: `VRAM Total Memory (B): 103079215104` (96 GiB BIOS UMA)
- leftover OS RAM: ~31 GB
- doctor was reporting `gpu_vram_gb = 31` because the Ryzen AI MAX unified-memory override stomped the independent reading

The gap test is the same idea as Windows `#810` (`UMA_CARVEOUT_MIN_GAP_GB`): a firmware carve-out is VRAM that exceeds leftover RAM by ≥1 GB. Intel UHD 630 aliases 1:1 to RAM and fails that test, so this is not keyed to a product name.

Does **not** yet flip Intel iGPUs to `vram: None` / `has_gpu: false` — that would break Lunar Lake Arc iGPU unified-pool grading (#609).

## Test plan
- [x] Regression fixtures from the #964 comment (96 GiB bytes, leftover 30.98 GB, rocm-smi with low-power warning)
- [ ] `cargo test -p llmfit-core --lib strix_halo`
- [ ] `cargo test -p llmfit-core --lib ram_carveout`

### DIFF
--- a/llmfit-core/src/hardware.rs
+++ b/llmfit-core/src/hardware.rs
@@ -211,7 +211,11 @@ impl SystemSpecs {
             });
             if let Some(idx) = amd_idx {
                 gpus[idx].unified_memory = true;
-                gpus[idx].vram_gb = Some(apu_pool_gb);
+                // Linux: rocm-smi / sysfs already reported the BIOS UMA
+                // carve-out (issue #964: 96 GB on leftover ~31 GB RAM).
+                // Do not overwrite that with the OS-visible remainder.
+                gpus[idx].vram_gb =
+                    apply_apu_vram_override(gpus[idx].vram_gb, apu_pool_gb, total_ram_gb);
                 // When detection could only produce a generic name (e.g. rocm-smi
                 // reported "N/A"), use the APU model instead — it names the iGPU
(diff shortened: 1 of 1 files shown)


AlexsJones:
Closing in favour of #995, which fixes both halves of #964 at the root instead of the gap heuristic. The doctor dump shows rocm-smi actually succeeded here (the `(gfx1151)` suffix only comes from the product parser); the 96 GB reading was being overwritten by the Linux `total_ram_gb`, which excludes the BIOS carveout. #995 folds the carveout back into the RAM total (the Linux analogue of #810) rather than comparing VRAM against leftover RAM, drops the unevidenced nvidia-smi exit-status change flagged by Greptile, and also applies the legacy Intel HD/UHD rule to the Linux lspci and Vulkan paths.

PR #999, opened 2026-09-04
feat(concurrency): concurrent-session capacity estimator (#140)

### LINKED ISSUE #140: Feature Request: Estimate Concurrent Users per Model (Context-Length Aware)
It would be very useful if llmfit could estimate how many concurrent users each model can support based on the detected hardware (VRAM, RAM, CPU/GPU, etc.).

Importantly, this estimation should also take the configured context length into account, since KV cache size and memory usage scale with context size and directly impact concurrency.

For example, in addition to showing whether a model fits in memory, llmfit could provide an approximate concurrency metric (e.g., estimated simultaneous inference sessions under typical load and selected context length).

This would make llmfit much more useful for real-world server deployments.

Thanks!

Akciali:
**Summary**
Adds a concurrent-session capacity estimate: how many simultaneous inference sessions of a model fit in a memory pool at a given context. It is the server planning counterpart to `usable_context`.

**Approach**
`sessions(ctx) = floor((pool - weights_resident) / per_session(ctx))`. `weights_resident` is the weight held in the pool for the run mode: `estimate_memory_gb(quant, 0)` (weights plus one time overhead) for a resident model, or only the active experts (`moe_active_vram_gb_at`) under MoE offload, so an offloaded MoE no longer reports zero sessions. `per_session` is the GQA and layout aware KV cache, plus a per session recurrent state term for hybrids (see below). Contexts run a doubling ladder (4k to 256k) or a single `--context`, clamped to the effective ceiling (native window, further capped by `--max-context` / `OLLAMA_CONTEXT_LENGTH`); `requested_context` keeps the asked value and `clamped` marks any rung the ceiling reduced.

**Two surfaces**
- `fit --concurrency` (and the global `--concurrency` default view): a density section under the fit table, at a fixed estimation context.
- `llmfit concurrency <model>`: `--quant`, `--kv-quant`, `--context`, `--users`, `--json`.

**Validation (measured)**
The KV term was checked against llama.cpp `llama_kv_cache` self size on an RX 7900 XT: exact match at 8k (512 MiB) and 16k (1024 MiB) on a hybrid Qwen3.5 class model (16 full attention of 64 layers, head_dim 256, 4 KV heads).

**Hybrid SSM handling**
Hybrid SSM and linear attention models hold a fixed size recurrent state per sequence (measured about 150 MiB), context independent, so it is charged per session rather than once. It is modeled as a coarse `linear_layers * hidden_size * C` estimate (C fitted to the measured hybrid), deliberately approximate, and zero for pure attention models. This corrects a concurrency overcount (about 23% at 8k) for those architectures.

**Quantization**
`--quant` is validated against the labels `quant_bpp` sizes, so an unknown or mis-cased value is rejected instead of silently falling to the 0.58 default. `quant_bpp` gains `AutoRound-4bit` (0.5) and `AutoRound-8bit` (1.0) at the AWQ/GPTQ scale, which also corrects fit's sizing of those models.

**Tests**
Hermetic: KV calibration, ladder monotonicity, native window clamp, tiny pool zero, coarse fallback without arch metadata, recurrent term calibration (about 150 MiB), recurrent reduces hybrid concurrency, a synthetic MoE fixture showing active-expert residency keeps sessions where full weight yields zero, and a context cap that clamps `effective_context` while preserving `requested_context`. CLI: `--users 0` and `--context 0` rejected at the parse boundary, unrecognized `--quant` rejected, `--max-context` honored by the ladder, and per-session formatting keeps sub-10-MiB values visible. `fmt` and `clippy` clean.

**Display**
The capacity table collapses ladder rungs that clamp to the same (context, sessions) pair, and formats per-session memory in MiB below 0.01 GB so small caches do not render as `0.00 GB`. JSON output and the estimate are unchanged.

**Limitations**
The recurrent term is an estimate, not exact per tensor accounting, and it only fires when a hybrid's layout resolves linear layers and `hidden_size` is known. SSM dims are not yet in the catalog. Under MoE offload, concurrency charges only the active experts, inheriting fit's own offload accounting (`moe_memory_for_quant`); fixed weights (attention, router, embeddings, shared experts) are not charged separately.

Closes #140

### DIFF
--- a/llmfit-core/src/concurrency.rs
+++ b/llmfit-core/src/concurrency.rs
@@ -1,0 +1,496 @@
+//! Concurrent-session capacity estimation (issue #140).
+//!
+//! Answers "how many simultaneous inference sessions of this model fit in the
+//! memory pool at a given context length" — the server-planning counterpart to
+//! [`crate::fit::ModelFit::usable_context`], which solves the same memory
+//! balance for tokens instead of sessions. Model weights and the fixed runtime
+//! overhead are resident once; each concurrent session adds one KV cache at the
+//! chosen context:
+//!
+//! ```text
+//! sessions(ctx) = floor( (pool - weights_resident) / kv_cache(ctx) )
+//! ```
+//!
+//! `kv_cache(ctx)` is GQA- and layout-aware via
+//! [`crate::models::LlmModel::kv_cache_gb`]; `weights_resident` is the weight
+//! actually held in the pool for the run mode: the full model
+//! ([`crate::models::LlmModel::estimate_memory_gb`] at context 0, weights plus
+//! one-time overhead, no KV) for a resident model, or only the active experts
+//! under MoE offload. This is a memory-capacity ceiling, not a throughput
+//! figure: it says how many sessions can be *resident*, not how fast they run
+//! under concurrent load.
+
+use serde::Serialize;
+
+use crate::fit::RunMode;
+use crate::models::{KvQuant, LlmModel};
+
+/// Default context ladder reported when no explicit context is given: 4k to
+/// 256k, doubling. Makes the memory-versus-context trade-off visible at a
+/// glance (a model that hosts many sessions at 8k but only one at 128k is the
+/// common real case).
+pub const DEFAULT_CONTEXT_LADDER: [u32; 7] =
+    [4_096, 8_192, 16_384, 32_768, 65_536, 131_072, 262_144];
+
+/// Concurrency at one context length.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct ConcurrencySlot {
+    /// Context requested for this rung.
+    pub requested_context: u32,
+    /// Context actually used, clamped to the effective ceiling (the native
+    /// window, or the smaller `--max-context` cap).
+    pub effective_context: u32,
+    /// True when `requested_context` exceeded the effective ceiling (the native
+    /// window, or the smaller `--max-context` cap) and was clamped down.
+    pub clamped: bool,
+    /// KV cache size for a single session at `effective_context`, in GB.
+    pub per_session_kv_gb: f64,
+    /// Maximum concurrent sessions that fit alongside the resident weights.
+    pub max_sessions: u32,
+}
+
+/// Full concurrency estimate for one model against one memory pool.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct ConcurrencyEstimate {
+    /// Memory pool the sessions share, in GB.
+    pub pool_gb: f64,
+    /// Resident weight held in the pool, in GB: the full model plus one-time
+    /// runtime overhead, or only the active experts under MoE offload.
+    pub weights_resident_gb: f64,
+    /// Pool left for KV caches after the resident weights, in GB.
+    pub kv_budget_gb: f64,
+    /// Coarse per-session recurrent-state estimate (GB) for hybrid SSM models,
+    /// added to each session on top of its KV cache. Zero for pure attention
+    /// models. See [`crate::models::LlmModel::recurrent_state_estimate_gb`].
+    pub per_session_recurrent_gb: f64,
+    /// Weight quantization the estimate assumes.
+    pub quant: String,
+    /// KV cache element representation the estimate assumes.
+    pub kv_quant: KvQuant,
+    /// Model's native context window.
+    pub native_context: u32,
+    /// One entry per requested context length.
+    pub ladder: Vec<ConcurrencySlot>,
+}
+
+impl ConcurrencyEstimate {
+    /// Largest ladder rung (by effective context) that still fits at least
+    /// `users` concurrent sessions, if any.
+    pub fn max_context_for(&self, users: u32) -> Option<u32> {
+        self.ladder
+            .iter()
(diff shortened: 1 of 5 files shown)


AlexsJones (changes requested):
Thanks for separating resident weights from per-session KV and recurrent state. One CLI boundary needs pinning before this leaves draft: `--users 0` currently reports the largest context as fitting, and `--context 0` produces a zero-context row. Please reject zero for both arguments with a range parser and add focused CLI regressions.

Akciali:
On the reviewed commit (2fbe361) the zero cases do not reproduce for me: both `--context` and `--users` already carry clap's `value_parser!(u32).range(1..)`, so `--users 0` and `--context 0` fail at parse time with exit 2 (`0 is not in 1..=...`) rather than reporting a largest context or a zero-context row. We may have been looking at different states. Either way the guard had no coverage, so I added focused CLI regressions in `cli_smoke.rs` pinning both boundaries.

While checking that I also fixed a readability issue in the same command. When a model's native window sits below the first ladder rung, every rung clamps to it and the table repeated the same line (seven identical `clamp` rows for a 512 window), and the `KV/session` column rounded sub-10-MiB caches to `0.00 GB`. The table now prints each distinct (context, sessions) pair once and switches to MiB below 0.01 GB. JSON output and the estimate are unchanged. Amended into the draft.

- Akciali

AlexsJones (commented):
Thanks for checking the reviewed commit and adding the boundary tests. You're right: both range parsers were already present at `2fbe361`, so my earlier zero-input defect claim was wrong, and I withdraw that blocker. The new smoke tests cover rejection, and the MiB formatting and duplicate-rung suppression are sensible display changes. The current platform CI is green. This remains a draft pending your readiness signal; no additional change is requested in this delta review.

Akciali:
Thanks. Addressing the three P1s.

**Quantization not validated:** fixed. `--quant` now rejects any label the sizing layer does not recognize (wrong case like `q8_0` included) instead of silently sizing it as Q4. Symmetric with `--kv-quant`, plus a CLI regression.

**Concurrency ignores the global context cap:** fixed. `--max-context` and `OLLAMA_CONTEXT_LENGTH` now clamp the ladder, an explicit `--context`, and the fit density view, not just the preliminary fit analysis. A CLI regression asserts no rung exceeds the cap.

**MoE offload charges full weights:** real gap, `estimate_concurrency` subtracts the full-model weight regardless of run mode. I could not reproduce `MoeOffload` on my hardware to size the effect, and a correct fix means run-mode-aware resident weight, which touches the core signature. In this PR, or a follow-up with a documented limitation? Your call.

- Akciali

AlexsJones (changes requested):
Thanks for adding quant validation and applying the global context cap. Please handle MoE offload in this PR, either by charging each pool only its resident weights or by explicitly refusing concurrency estimates for unsupported offload modes. `estimate_concurrency` still subtracts full-model weights from the GPU pool supplied by the CLI, so a fit model can incorrectly report zero sessions. A synthetic MoE fixture can test this without owning that hardware; cover standalone output and the fit concurrency view.

Akciali:
MoE offload handled by charging each pool its resident weight. `estimate_concurrency` now takes the run mode and, under `MoeOffload`, subtracts only the active-expert VRAM (`moe_active_vram_gb_at`) instead of the full model, so a fit MoE no longer reports zero sessions. Other run modes keep full-weight residency. Both surfaces, the `concurrency` command and the fit density view, pass the fit's run mode through the same estimator.

Covered by a synthetic MoE fixture: at a pool between the active-expert and full weight, the full-weight charge yields zero sessions while the active charge leaves budget for real ones. I could not drive `MoeOffload` from the CLI on my hardware, it stays `Gpu`, which is why the fixture carries the test.

- Akciali

Akciali:
Two Greptile findings. One fixed at the root, one is your call. Flagging it plainly: each fix here surfaces the next gap in the shared memory accounting, so I would rather scope with you than keep amending on the bot's cadence.

AutoRound: `quant_bpp` had no AutoRound arm, so those models fell to the 0.58 default everywhere, not only in concurrency. Added `AutoRound-4bit` (0.5) and `AutoRound-8bit` (1.0) at the AWQ/GPTQ scale and kept the accepted set in sync. This also corrects fit's own sizing.

MoE fixed weights: under `MoeOffload` concurrency charges active experts × bpp × 1.1, the same `moe_memory_for_quant` fit uses to pick the offload path. Greptile is right that attention, router, embeddings and shared experts stay resident, but charging them in concurrency alone makes it stricter than fit's own check, so a model fit calls "MoE offload" could show zero sessions. Fix it in the shared MoE model (fit and concurrency together), or document that concurrency inherits fit's offload accounting? I note the revert_497 history here.

- Akciali

AlexsJones (approved):
Thanks for wiring MoE offload through `moe_active_vram_gb_at` so only the active experts charge the VRAM pool, plus the `quant_is_recognized` guard and the CLI boundary regressions. The tests now pin the offload accounting and the zero-argument rejections. Ready for your final look.

Akciali:
Final look done on f560be2: description refreshed to the final state (run-mode-aware residency, quant guard, context cap metadata, AutoRound in `quant_bpp`) and a Limitations note that concurrency inherits fit's MoE offload accounting. Ready to merge from my side.

- Akciali

PR #971, opened 2026-08-29
feat: hardware profiles, MoE Tier-2 fixes, and estimate confidence (#969)

saman-mb:
## What this PR does

Lets you score models against a **named machine** (RAM, unified vs discrete, bandwidth, optional FP16 TFLOPS) instead of only the host you are on — including hardware you do not own yet.

Also tightens MoE estimates, adds a confidence label on every fit JSON row, and keeps draft draft-model families out of ranked fits.

Closes #969

---

## How it works (spin up the CLI)

**1. List what ships in the binary**

```sh
llmfit hardware list
llmfit hardware show ryzen-ai-max-plus-395
```

**2. Score against that machine**

```sh
llmfit --profile ryzen-ai-max-plus-395 fit -n 10
llmfit --profile ryzen-ai-max-plus-395 plan openai/gpt-oss-120b
```

**3. Simulate anything else (unreleased SKU, custom box)**

Write a small JSON profile → validate → pass `--profile` by name or path. Full step-by-step:

📘 **[Walkthrough: Hardware profiles](https://github.com/saman-mb/llmfit/blob/feat/issue-969-hardware-profiles/docs/cli.md#hardware-profiles)**  
(also linked from [`llmfit-core/data/hardware/README.md`](https://github.com/saman-mb/llmfit/blob/feat/issue-969-hardware-profiles/llmfit-core/data/hardware/README.md))

Quick taste:

```sh
# one-off file
llmfit --profile ./my-box.json fit --json

# or install under the user dir and use the name forever
llmfit hardware path    # drop <name>.json here
llmfit hardware validate ~/.local/share/llmfit/hardware/my-box.json
llmfit --profile my-box fit -n 20
```

`--profile` replaces `--memory` / `--ram` / `--cpu-cores` (do not mix). Bundled seeds: `ryzen-ai-max-plus-395`, `apple-m3-max-128gb`, `nvidia-rtx-4090`.

---

## What else changed (same PR)

| Area | Behavior |
| --- | --- |
| Fit verdict | Still Perfect / Great / OK / Poor from memory headroom; Perfect still needs GPU |
| MoE | Per-arch Tier-1/Tier-2 constants (`gpt-oss`, DeepSeek-class); default path unchanged for other arches |
| Bandwidth | Single resolve path; profile BW drives `estimated_tps` and `estimate_basis.gpu_bandwidth_gbps` |
| Prefill / TTFT | Only when the profile sets FP16 TFLOPS; otherwise honest `null` |
[...]

Issue #969, opened 2026-08-29
Hardware Profiles: spec-based simulation for unreleased/undetectable hardware, corrected MoE estimates, and confidence labels


saman-mb:
## The problem

New hardware is announced with full specs months before anyone can buy it — every vendor does this. The exact moment people ask "will my models run on this thing?", llmfit can't answer, because:

**1. There is no way to feed published specs into the tool.**
Bandwidth comes from a name-string table (`hardware.rs:2707`) with no override flag. The `--memory/--ram/--cpu-cores` overrides model a discrete GPU+RAM split, which is the wrong shape for unified-memory machines, and they can't express bandwidth at all. If a chip isn't in the table yet — or is exotic, mobile (#919 deliberately returns `None`), or a VM — you're stuck estimating outside the tool. This is #193's ask, still open.

**2. MoE decode is underestimated ~3x on paths without full expert metadata.**
`#474` / `#475` / `#807` fixed the total-vs-active-params bug; this is the remaining Tier-2 (metadata-absent) path. Fresh v1.1.12 output for `gpt-oss-120b` (117B total / 5.1B active) when the bandwidth roofline uses **256 GB/s** (e.g. a GPU name that hits the table at that figure): **15.8 tok/s**. Measured on real 256 GB/s hardware: **50.2** ([Quozul, llama.cpp tg128](https://quozul.dev/posts/2025-10-27-gpt-oss-120b-on-ryzen-ai-max-395/)). Same class of gap on `DeepSeek-V4-Flash` (284B/13B active): ~14 predicted on a 614 GB/s config vs **33.8-36.1** measured ([MLX](https://huggingface.co/mlx-community/DeepSeek-V4-Flash-0731-2.4bit-mixed)). The Tier-1 decomposition in `fit.rs:1374` requires config-level expert metadata that community/MLX repos don't carry, so they fall back to Tier-2 + the global 0.55 efficiency and land far below reality.

**3. The model DB lets impossible entries through.**
A 117B model at 1.2GB (0.08 bits/param), a speculative-decoding draft listed at 3.2B for a 120B model, `best_quant` reading "Q8_0" on NVFP4 repos, DSpark/DFlash/EAGLE drafts ranking as standalone fits. Anyone consuming `fit --json` programmatically has to re-filter the output.

## What I propose — one PR

[...]

PR #1003, opened 2026-09-05
Feat/974 tui confidence prefill

### LINKED ISSUE #974: ux(tui): show estimate confidence (and prefill) in detail pane
## Context

UX review on PR #971: CLI detail shows `Confidence:` and conditional prefill under Estimate Basis, but the TUI `draw_detail` path does not yet mirror those lines. The fit table also gained a full-width Confidence column that may wrap on narrow terminals; Stage 1.5 preferred keeping the table narrow and putting the full label in detail only.

## Ask

- Add `Confidence:` (+ conditional prefill/TTFT) to TUI detail estimate-basis block
- Consider dropping or shortening the table Confidence column (codes vs full labels)

Related: #969, #971

yashchandnani07:
Closes #974

## What

Two changes, matching the scope agreed in the issue:

1. **TUI detail pane** now shows `Confidence:` (full label) plus conditional `Prefill:` / `TTFT:` rows under `Baseline Est:`, mirroring the CLI's `Estimate Basis:` block from #971.
2. **CLI fit table** Confidence column now prints a short code under a `Conf` header instead of the full label.

## Why

Per the issue and the maintainer's note: the table should stay compact on narrow terminals while the detail pane carries the full label and prefill/TTFT context.

The table column was sized by its widest cell — `measured (this machine)` at 23 columns — which wrapped rows on an 80-column terminal. It is now capped at 5.

The CLI's prefill sentence is 78 columns, and the TUI detail pane is 42 on an 80-column terminal — 27 after the label gutter. Folding both figures onto one line does not fit that budget at high-end values, so prefill and TTFT get a row each, matching the rest of the pane. Both rows are omitted for fits with no throughput estimate, which is where the CLI stops its own block.

## Narrow-width coverage

`prefill_detail_rows` returns plain `(label, value)` pairs rather than ratatui widgets, so it is asserted on directly, in the style of this file's existing `provider_cell` / `truncate_str` tests:

- `prefill_detail_rows_fit_a_narrow_detail_pane` — asserts every value fits the 27-column budget and every label fills the 15-column gutter, at values well past what real hardware produces
- `confidence_labels_fit_a_narrow_detail_pane` — same for the full labels the pane now shows
- `estimate_confidence_short_labels_stay_table_narrow` — caps the table codes at 5 ASCII columns
- `estimate_confidence_short_labels_are_distinct` — no two confidences collapse to one code

I also drove the built TUI in an 80-column terminal to confirm the detail pane renders without wrapping, and eyeballed the CLI table's new `Conf` column.

## API compatibility

`EstimateConfidence::code()` and `label()` are untouched — they ship as `estimate_confidence` and `estimate_confidence_label` in the API/MCP JSON, and their existing pinning tests still pass. `short_label()` is additive and display-only.

Both new call sites derive confidence via `effective_estimate_confidence()` rather than the stored field, per #969.

## Open questions for the maintainer

1. **Placement.** The TUI has no `Estimate Basis` section; this PR puts the two new lines at the end of `Score Breakdown`, right under `Baseline Est:` — the number they qualify. Happy to pull them into a dedicated `── Estimate Basis ──` section if you'd rather mirror the CLI's structure more literally.
2. **Code spelling.** `local` / `comm` / `calib` / `est` / `n/a`. If you'd prefer the serde codes truncated, or symbols (`✓` / `~`), it's a one-line change in `short_label()`.
3. **Header rename.** `Confidence` → `Conf` saves 6 more columns. If anything scrapes the CLI table's header text, let me know and I'll keep the header while shrinking just the cells.

### DIFF
--- a/docs/how-it-works.md
+++ b/docs/how-it-works.md
@@ -64,13 +64,15 @@ The scoring model, speed estimation, and the model database.
 
    **Confidence** -- A measured throughput and a formula guess are both "tok/s", so every fit carries an `estimate_confidence` saying which it is. First match wins:
 
-   | Confidence           | Meaning                                                        |
-   |----------------------|----------------------------------------------------------------|
-   | `measured_local`     | Benchmarked by you on this machine (`llmfit bench`)            |
-   | `measured_community` | Benchmarked by others on hardware matching this machine        |
-   | `calibrated`         | Formula, scaled by a factor derived from runs on this hardware |
-   | `estimated`          | Formula only, with no measurement behind it                    |
-   | `unsupported`        | No estimate -- the model needs a runtime llmfit can't model    |
+   | Confidence           | Table | Meaning                                                        |
+   |----------------------|-------|----------------------------------------------------------------|
+   | `measured_local`     | local | Benchmarked by you on this machine (`llmfit bench`)            |
+   | `measured_community` | comm  | Benchmarked by others on hardware matching this machine        |
+   | `calibrated`         | calib | Formula, scaled by a factor derived from runs on this hardware |
+   | `estimated`          | est   | Formula only, with no measurement behind it                    |
+   | `unsupported`        | n/a   | No estimate -- the model needs a runtime llmfit can't model    |
+
+   The `Table` column is what the CLI fit table's `Conf` column prints; the detail views (`llmfit fit <model>` and the TUI detail pane) print the full name.
 
 6. **Fit analysis** -- Each model is evaluated for memory compatibility:
 
--- a/docs/how-it-works.md
+++ b/docs/how-it-works.md
@@ -72,7 +72,7 @@ The scoring model, speed estimation, and the model database.
    | `estimated`          | est   | Formula only, with no measurement behind it                    |
    | `unsupported`        | n/a   | No estimate -- the model needs a runtime llmfit can't model    |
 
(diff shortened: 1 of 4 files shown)


AlexsJones (commented):
Thanks for keeping the table compact while preserving the full detail label. The four-file diff uses effective confidence, leaves API codes untouched, and handles missing prefill estimates explicitly. Before this is ready, please get the Rust tests, formatting and Clippy checks run at this head; the only reported check is Greptile. The added width assertions cover strings, so they should be described as width checks rather than a rendered-terminal regression.

yashchandnani07:
Validated at `b914b99` in a Linux Docker environment:

- `cargo fmt --all -- --check` ✅
- `cargo test` ✅ — 705 core tests, 3 hardware-profile tests, and 6 ONNX tests passed; 1 test ignored
- `cargo clippy --all-targets --all-features -- -D warnings` — the command reaches the workspace but reports existing warnings in untouched code (`models.rs`, `quality.rs`, `update.rs`, `claim.rs`, and older test helpers). None are in this PR's changed lines.

The PR remains a focused four-file change, has no whitespace errors, and merges cleanly with current `main`. No source changes were needed for this validation.

AlexsJones:
Thanks for the work on this one. Quick check-in: the head hasn't moved since the last review and the Rust tests, formatting, and Clippy still haven't reported here, so the earlier ask stands.

Once those run green at this head, this gets another look right away.

PR #1022, opened 2026-09-11
bench: community results for nvidia-geforce-rtx-3050-ti-laptop-gpu

spsanderson:

| --- | --- | --- | --- |
| gemma3:1b-it-qat | ollama | 80.1 | 87.0 |

### DIFF
--- a/llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789092233-8d9645ad.json
+++ b/llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789092233-8d9645ad.json
@@ -1,0 +1,33 @@
+{
+  "hardware": {
+    "cpu": "AMD Ryzen 7 4800H with Radeon Graphics",
+    "cpuCores": 16,
+    "gpuCount": 1,
+    "hardwareName": "NVIDIA GeForce RTX 3050 Ti Laptop GPU",
+    "hwClass": "DISCRETE_GPU",
+    "memTierGb": 4,
+    "os": "windows",
+    "ramGb": 31.42,
+    "unifiedMemory": false,
+    "vramGb": 4.0
+  },
+  "results": [
+    {
+      "avgOutputTokens": 194.0,
+      "avgTotalMs": 2548.94,
+      "avgTps": 80.1,
+      "avgTtftMs": 86.98,
+      "maxTps": 82.31,
+      "minTps": 78.35,
+      "model": "gemma3:1b-it-qat",
+      "numRuns": 3,
+      "provider": "ollama"
+    }
+  ],
+  "schemaVersion": 1,
+  "submittedAtUnix": 1789092233,
+  "tool": {
+    "name": "llmfit",
+    "version": "1.1.15"
+  }
+}
--- a/llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789092890-1ac3c0b4.json
+++ b/llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789092890-1ac3c0b4.json
@@ -1,0 +1,33 @@
+{
+  "hardware": {
+    "cpu": "AMD Ryzen 7 4800H with Radeon Graphics",
+    "cpuCores": 16,
+    "gpuCount": 1,
+    "hardwareName": "NVIDIA GeForce RTX 3050 Ti Laptop GPU",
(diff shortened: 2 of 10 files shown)


AlexsJones (approved):
Thanks for the RTX 3050 Ti laptop results. Matches the community format. Good to merge.

AlexsJones (changes requested):
Thanks for the extra runs, having ten sessions from the same 3050 Ti laptop makes the small-model numbers much more trustworthy.

One file needs to come out before this merges: `1789127487-3fa7521e.json` records `hf.co/defog/sqlcoder-7b-2:Q5_K_M` at `avgTps: 333336.5` with `maxTps: 1000000.0`. That's a timing glitch in the bench (a near-zero `eval_duration` from Ollama), not your GPU, and once merged it would show up as the measured speed for that model. Please drop that file (or rerun that model and replace it), and the rest is good to go.

The underlying bench bug is tracked separately in #1038, so nothing else is needed from you.

spsanderson:
I’ll re-run the model


Steven P Sanderson II, MPH
Amazon Author Profile
<https://www.amazon.com/stores/Steven-P.-Sanderson/author/B0CZ157DHK?ref_=pe_2466670_811284380&isDramIntegrated=true&shoppingPortalEnabled=true>
Book on Lulu <http://goo.gl/lmrlFI>
Personal Site <http://www.spsanderson.com>

AlexsJones (changes requested):
Thanks for the extra 3050 Ti run! Two things before this goes in:

- `1789127487-3fa7521e.json` (sqlcoder-7b-2 at 333,336 tok/s) is still in the PR. Since #1042 llmfit skips rates like that when it reads community data, so it would no longer show up as a real number, but it's still a broken measurement in the dataset. Please drop that file.
- The new `1789696327-18faca62.json` has `llama3.2:1b` at 8.85 tok/s, while your two earlier runs of the same model on this laptop came in at 27.1 and 27.0. A 3x drop on the same hardware usually means something else was using the GPU or the model spilled to CPU. Was anything else running? If you can't reproduce the 8.85, please drop that file too.

Everything else looks consistent.

spsanderson (commented):
ready for next review.

AlexsJones (changes requested):
Thanks for pulling the sqlcoder file, that was the worse of the two and it's gone: `1789127487-3fa7521e.json` now reads llama3.2:1b 27.01, qwen2.5-coder:1.5b 83.14, sqlcoder:7b 6.87, gemma3:1b-it-qat 78.85 and llama3.2:latest 12.55, all with min ≤ avg ≤ max and all plausible for a 4 GB 3050 Ti.

One file left and then this is done: `1789696327-18faca62.json`. I had flagged it as a slow outlier (8.85 tok/s for llama3.2:1b where your four other runs of the same model on the same laptop read 26.04 to 27.85). Having looked closer, the reason to drop it is stronger than "it looks slow", because the file does not reconcile with itself:

- `avgOutputTokens` 168.33 over `avgTotalMs` 43065.93 is 3.9 tok/s, but `avgTps` says 8.85, and `avgTtftMs` is only 248 ms.
- Your good runs do reconcile. In `3fa7521e`, 158.33 tokens over 6053.3 ms is 26.2 tok/s against a declared 27.01, a 3% gap from per-run averaging.

So roughly 24 of those 43 seconds are in neither prefill nor generation. That is the signature of a disturbed run (a cold load off disk, or something else holding VRAM on a 4 GB card and forcing part of the model to CPU) rather than a measurement of the hardware. Either way it is not a number the fit estimator should learn from.

Drop that one file and I will approve. Everything else here is good, and eleven submissions across five models on an under-represented laptop GPU is a genuinely useful contribution.

I approved the pending CI and Community Benchmarks runs so they report on this head.

spsanderson:
@AlexsJones the 1789696327-18faca62.json has been dropped

AlexsJones:
Thanks @spsanderson! I think the delete didn't make it up to GitHub though: the PR head is still `6b8fd6c2` (your 22 Sep commit), and `llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789696327-18faca62.json` is still in the file list.

Probably just a commit that's local but not pushed. Something like this on your branch should do it:

```
git rm llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789696327-18faca62.json
git commit -m "drop unreconciled llama3.2:1b run"
git push
```

Once that lands I'll approve straight away.

PR #1023, opened 2026-09-11
feat(storage): add disk planning for model libraries

### LINKED ISSUE #970: Feature: disk/SSD planning for multi-model libraries (keep N models that fit)
## Summary

`llmfit fit` / `info` already expose `disk_size_gb` for the chosen quant, but there is no first-class way to answer:

> Given this machine (or `--memory` / `--ram` simulation), how much **internal SSD** do I need to comfortably download and keep N models that fit, plus download scratch?

That is a real buy/config decision for local LLM desktops (Mac mini / Mac Studio, etc.): RAM decides what loads; disk decides whether you can switch between models without delete-and-rewait.

## What works today

- `disk_size_gb` on fit / recommend / info JSON rows
- `llmfit plan` for VRAM/RAM/context/quant capacity planning

## What is missing

1. **`plan` does not surface disk/weight storage** for the planned quant (no `disk_size_gb` / `weight_gb` in the plan JSON I get today — only VRAM/RAM/KV paths).
2. No command or flag to aggregate storage across **multiple models that fit** a machine, e.g.:
   - keep the top N fitting models on disk
   - plus OS/apps reserve
   - plus scratch equal to one more download
   - round up to common SSD SKUs (512GB / 1TB / 2TB / 4TB …)
3. No guidance in docs that `disk_size_gb` is the field to use for library sizing (easy to miss; buyers still guess).

Related but different: #542 (server density / concurrent instances). This issue is about **on-disk library size for sequential switching**, not concurrent serving density.

## Proposed UX (sketch)

```bash
# Single model: include disk in plan output
llmfit plan "openai/gpt-oss-120b" --context 8192 --json
# → add disk_size_gb / weight_on_disk_gb for the planned quant

# Multi-model library sizing against simulated hardware
llmfit storage --memory 128G --ram 128G --cpu-cores 18 \
  --keep 3 --os-reserve 100G --json
# → models considered = those that fit; report need_gb + suggested_ssd_gb
```

Optional filters: `--perfect`, model list / search query, exclude Too Tight.

## Why it matters

Without this, people configure Macs (and other local boxes) with enough unified memory but undersized SSDs, then discover a 120B-class weight is ~100GB+ on disk and three of them blow past a 512GB drive after the OS.

## Workaround

We currently post-process fit dumps in  
https://github.com/saman-mb/mac-studio-m5-analysis/blob/main/scripts/disk_comfort.py  
using llmfit’s `disk_size_gb`. Happy to adapt that formula if you want it upstream.

## Environment

- llmfit via pipx (`llmfit --help` shows fit JSON includes `disk_size_gb`)
- Use case: Apple Silicon desktop config planning (M6 mini / M5 Studio), 13 simulated configs via `--memory` / `--ram` / `--cpu-cores`

eyenpi:
Closes #970.

Adds estimated weight storage to `plan` and a `storage` command for sizing an SSD for a library of models used sequentially.

The change is split into the two commits requested in the issue:

1. **Plan disk estimate:** expose `disk_size_gb` in the shared plan result and CLI text output. It uses the existing weight-size formula at the resolved quantization, includes all MoE experts, and excludes KV cache and runtime buffers. CLI, HTTP API, and MCP tests cover the new field. Regression tests cover the shared estimator's AutoRound-4bit and AutoRound-8bit weight factors (0.5 and 1.0 bytes per parameter), including full stored MoE weights.
2. **Library storage sizing:** select up to `--keep N` runnable models through the shared fit analysis, including backend eligibility, hardware overrides/profiles, and context limits. Selection defaults to fit score; `--selection largest`, `--perfect`, and `--search` narrow the library.

Storage totals combine the selected weights, an adjustable OS/apps reserve, and download scratch. Defaults are three models, 100 GB reserve, scratch equal to the largest selected model, and 15% free headroom. Minimum and suggested capacities round up to common decimal SSD tiers. Empty results and requirements beyond the listed tiers return explicit warnings and null recommendations. Full catalog IDs remain distinct so repositories with the same model slug do not collide.

The command estimates one library copy; it does not measure current disk usage or download models. CLI documentation explains units, allowances, and estimate limitations.

```sh
llmfit plan "openai/gpt-oss-120b" --context 8192 --json
llmfit --memory 128G --ram 128G --cpu-cores 18 storage --keep 3 --selection largest --json
```

### Verification

Verified locally on macOS ARM64 using stable Rust 1.95.0 at `12554cd`, based on upstream `1936acc`:

- `cargo fmt --all -- --check` and `git diff --check` pass.
- `cargo test --verbose` passes: 912 tests passed, one pre-existing ignored test.
- `make check`, `make build`, `cargo check --all-targets --all-features`, and the CI command `cargo clippy --all-targets --all-features` pass.
- Web build and 14 tests pass. Local Node 26 requires `NODE_OPTIONS=--no-experimental-webstorage` for Vitest/jsdom; CI uses Node LTS.
- Python packaging and four tests pass; the packaged binary matches the freshly built feature binary.
- Ten existing CLI scenarios match the previously reviewed feature head after rebasing, accounting only for existing nondeterministic catalog ordering. An independent Python calculation verifies the storage totals and SSD tiers. CLI help was also checked for `plan`, `storage`, and the newly merged upstream `concurrency` command.

AutoRound regression tests were confirmed to fail before the factor correction and pass afterward. They cover full weight storage, MoE memory, fixed-quantization memory budgets, plan JSON, and SSD tier boundaries. The reported AutoRound-8bit model requires 268.26 GB with a 200 GB reserve and automatic scratch, selecting 512 GB on simulated CUDA and ROCm hardware. The same factor correction is now also present in upstream `1936acc`; this PR retains the regression coverage.

**Clippy baseline:** strict `make clippy` fails with the same 22 errors on untouched upstream `1936acc` and this branch. The CI-style Clippy runs report 84 diagnostics on each, with zero new diagnostics after comparing lint codes, messages, and source spans while ignoring shifted line numbers. Unrelated lint cleanup is outside this feature's scope.

**GitHub-hosted CI:** all eight jobs passed in the [fork CI run](https://github.com/eyenpi/llmfit/actions/runs/34826267687), triggered by [verification PR eyenpi/llmfit#1](https://github.com/eyenpi/llmfit/pull/1): Rustfmt, Clippy, Cargo Check, Python Checks, change detection, and the Ubuntu, macOS, and Windows test suites. Rust tests passed on each platform (905 on Ubuntu, 912 on macOS, and 910 on Windows; one existing ignored test per platform). Each platform also passed the Web and Python tests and debug build. The workflow checked out merge commit `0ab19a8`; its Git tree is identical to feature head `12554cd` (`db2550e207080d3a95828738f7d2f5dd405bd3ee`). This used the existing CI workflow without source or workflow file changes.

The [upstream CI run](https://github.com/AlexsJones/llmfit/actions/runs/34788107162) still awaits maintainer approval. The successful fork run provides native cross-platform verification; it does not approve or replace the upstream run.


AlexsJones (changes requested):
Thanks for the thorough verification notes, the clippy-baseline comparison and the fail-before/pass-after AutoRound tests are exactly the evidence that makes a big PR reviewable.

Read the full diff. The storage math is careful (decimal units with the binary-suffix distinction documented, validated inputs, deterministic ordering, explicit warnings for empty and over-tier results) and the plan field correctly uses the resolved quant with KV excluded. One thing before this is ready: no Rust CI has reported at this head yet, only the Greptile review. Please get the test, fmt, and Clippy workflows green here, then this gets the final look.

eyenpi:
@AlexsJones Thanks! I’ve rebased onto current main and rerun the checks locally. 912 tests pass, along with fmt and CI-style Clippy. The upstream workflow is awaiting maintainer approval Could you approve the workflow run? I’ll address any failures it reports.

AlexsJones (changes requested):
Thanks for the rebase.

I compared every file you own across the rebase (storage.rs, plan.rs, TUI wiring): byte-identical to the head I reviewed, the only additions are main's merged concurrency work. So the earlier read still stands.

Still one thing before the final look: no Rust workflow results at the new head yet, only the Greptile review. Once the test, fmt, and Clippy workflows report green here, this is ready.

eyenpi:
@AlexsJones I’ve now run the repository’s existing CI on my fork for the current head, `12554cd`. All eight jobs passed, including Rust tests on Linux, macOS, and Windows, Rustfmt, Clippy, Cargo Check, and the Web/Python checks:
[Successful CI run](https://github.com/eyenpi/llmfit/actions/runs/34826267687)
I verified that the source tree tested by CI exactly matches this PR’s head and added the results to the PR description.
The [upstream run](https://github.com/AlexsJones/llmfit/actions/runs/34788107162) still awaits maintainer approval. Could you approve it so the checks can report directly on this PR as well?

AlexsJones (approved):
Thanks for running the full CI on your fork, that closes the one open ask.

I checked [run 34826267687](https://github.com/eyenpi/llmfit/actions/runs/34826267687): it ran against exactly `12554cd`, all eight jobs passed (tests on Linux, macOS and Windows, Rustfmt, Clippy, Cargo Check, Python), and this PR doesn't touch `.github/`, so it ran the same workflow this repo uses. With the earlier full read and the byte-identical rebase check, this is ready for the final look.

The upstream run is still waiting on workflow approval, which is a maintainer click, so checks will report here once that's done.

PR #1042, opened 2026-09-14
fix(bench): ignore implausible tok/s from degenerate Ollama timings

### LINKED ISSUE #1038: bench: Ollama runs with a near-zero eval_duration record 1,000,000 tok/s and flow into community data
Found while reviewing community benchmark PR #1022.

One submitted file, `llmfit-core/data/community/nvidia-geforce-rtx-3050-ti-laptop-gpu/1789127487-3fa7521e.json`, records `hf.co/defog/sqlcoder-7b-2:Q5_K_M` on a 4 GB RTX 3050 Ti laptop at `avgTps: 333336.5` with `minTps: 4.75` and `maxTps: 1000000.0`. The same file has `sqlcoder:7b` at a believable 6.87 tok/s, so one of the three runs produced a physically impossible number and it went straight into the averaged result.

## Code path

- [`bench.rs` Ollama TPS](https://github.com/AlexsJones/llmfit/blob/5fec377a624d75e7c97b4af650c9d05fe27c621b/llmfit-core/src/bench.rs#L199-L205) computes `eval_count / (eval_duration / 1e9)` whenever `eval_duration > 0`. One token with a reported `eval_duration` of 1000 ns gives exactly 1,000,000 tok/s, which matches the recorded `maxTps`.
- `share` serializes the run summary as-is, so the outlier is averaged into `avgTps`.
- [`LocalBenchIndex::load`](https://github.com/AlexsJones/llmfit/blob/5fec377a624d75e7c97b4af650c9d05fe27c621b/llmfit-core/src/share.rs#L462-L465) only requires `avgTps > 0.0`, so once merged this entry would be served as the measured tok/s for that model on matching hardware.

## Suggested fix

- In the Ollama path, treat a run with too few eval tokens or an implausibly short `eval_duration` as unmeasured (fall back to wall clock or skip the run) instead of dividing.
- Reject or drop per-run TPS above a sanity ceiling before summarizing, and apply the same bound when loading community data so already-merged outliers can't surface.
- Regression test: an Ollama response with `eval_count: 1, eval_duration: 1000` must not yield 1e6 tok/s.

kaanisthatyou:
Fixes #1038

A run where Ollama reports `eval_count: 1` and `eval_duration: 1000` is computed as 1,000,000 tok/s. That value goes into `avgTps` and can later come back as a "measured" rate from shared or community results.

Changes:
- **`ollama_tps`** (extracted from `ollama_generate`): uses Ollama's native timing only when at least two tokens were evaluated and the rate is at most `MAX_PLAUSIBLE_TPS` (10,000 tok/s). Otherwise it falls back to the wall clock, which was already the fallback when the timing fields are missing. A single token has no decode interval to time. As a side effect, `eval_duration: 0` now also falls back to the wall clock; before, it gave `0.0`.
- **`BenchSummary::from_runs`**: leaves rates above the bound out of `avg/min/max_tps`. Zero rates are handled as before.
- **`LocalBenchIndex::load`** (`share.rs`) and **`community_results_for_specs`** (`benchmarks.rs`): use `bench::is_plausible_tps` instead of `> 0.0`, so results that were already saved or merged with an impossible `avgTps` are not used.

The 10,000 tok/s bound is a guess at "well above any local decode rate". I'm happy to change the number.

### Testing
With a fake Ollama `/api/generate` whose second measured run returns `eval_count: 1, eval_duration: 1000` (`llmfit bench sqlcoder:7b --runs 3`):
- on `main`: run 2 shows `1000000.0` and the summary is `TPS: 333336.5 avg (4.8 min / 1000000.0 max)`, the same numbers as the submission in #1022;
- with this change: run 2 falls back to the wall clock. It shows `270.2` because the fake server answers in a few milliseconds; a real server's wall time also includes prompt processing.

Unit tests `ollama_tps_ignores_degenerate_native_timing` and `test_summary_excludes_implausible_tps` were added. `cargo test -p llmfit-core --lib`: 744 passed. `cargo fmt` is clean, and `cargo clippy` shows no new warnings in the changed files.

This branch doesn't depend on #1041 (a different part of `bench.rs`).

https://claude.ai/code/session_01EcwjgZmx8zbtn6PpwDY9A1


AlexsJones (commented):
Thanks, extracting `ollama_tps` is the right cut: the `count >= 2` and `<= MAX_PLAUSIBLE_TPS` guards send the #1038 case to the wall clock, and the summary now leaves an implausible run out instead of averaging it in. One non-blocking thought: `BenchSummary::from_runs` filters with `<= MAX_PLAUSIBLE_TPS` while the other two call sites use `is_plausible_tps`, so a NaN rate is dropped from the average but a zero or negative one is kept. Using `is_plausible_tps` there too would keep one definition, if you agree a 0 tok/s run shouldn't count. Formal approval once CI has run on this head.

kaanisthatyou:
Good point — agreed that a 0 tok/s run shouldn't count either, so `BenchSummary::from_runs` now filters with `is_plausible_tps` too (f1486fb). One definition across the three call sites, and zero/negative rates are dropped like NaN instead of pulling the average down.

`test_summary_excludes_implausible_tps` now also covers a 0.0 run and a NaN run alongside the 1e6 one: with the old `<= MAX_PLAUSIBLE_TPS` filter the average comes out 2.375 instead of 4.75, with the new one `avg`/`min`/`max` are all 4.75 over the single real measurement. `cargo fmt --check` is clean; the clippy errors I see locally are pre-existing in `fit.rs`/`hardware.rs` and unrelated to this branch.

Issue #1024, opened 2026-09-11
Ollama gemma4:12b installed locally but not recognized by llmfit (ollama_name mapping missing for whole Gemma 4 family)


GMVrgb76:
**Environment**
- llmfit 1.1.15
- macOS aarch64
- Apple M4 / 24 GB unified memory
- Ollama installed

**Ollama**
```
$ ollama list
gemma4:12b   4eb23ef187e2   7.6 GB   6 weeks ago
```

**Problem**
```
$ llmfit info "gemma4:12b"
No model found matching 'gemma4:12b'
```

**Investigation**
llmfit's catalog contains Gemma 4 models (`google/gemma-4-E4B-it`, `google/gemma-4-E2B-it`, `google/gemma-4-31B-it`, `google/gemma-4-26B-A4B-it`, plus ~74 community fine-tunes matching `gemma-4-12*`), but grepping the full `llmfit fit --json` output shows `"ollama_name": null` on every single Gemma 4 entry. By contrast, Gemma 2 and Gemma 3 entries do carry a populated `ollama_name` (`gemma2:9b`, `gemma3:12b`, `gemma3:27b`, `gemma3n:e4b`, etc.).

So this isn't a fuzzy-search issue like #315 — hyphenated queries do resolve fine:
```
$ llmfit info "gemma-4-12b"
Multiple models match 'gemma-4-12b'. Please be more specific:
  - sakamakismile/Huihui-gemma-4-12B-it-abliterated-NVFP4A16
  - r0b0tlab/gemma-4-12B-it-nvfp4
  ...
```
The colon-tag form used by the official Ollama registry (`gemma4:12b`, and presumably `gemma4:e2b`, `gemma4:e4b`, `gemma4:26b`, `gemma4:31b`) simply has no mapping at all, so `info`/`fit`/`bench`/installed-detection can't resolve or recognize it, regardless of what's already pulled locally.

**Expected**
llmfit should detect the officially distributed Ollama tags under `gemma4:*` as installed models (when present in `ollama list`) and allow `info`/`fit`/`bench` operations on them, the same way it already does for `gemma2:*` and `gemma3:*`.

**Suggested scope**
Since the whole family (`gemma4:e2b`, `e4b`, `12b`, `26b`, `31b`, plus the MLX-tagged variants) shares the same `ollama_name: null` gap, the fix likely belongs at the mapping-table level rather than as a single 12B-only patch.

**Reference**
https://ollama.com/library/gemma4

[...]

PR #1043, opened 2026-09-14
fix(providers): map Gemma 4 instruct models to their Ollama tags

kaanisthatyou:
Refs #1024

The catalog's official Gemma 4 instruct models had no `OLLAMA_MAPPINGS` entries. The heuristic fallback generates `gemma-4:31b`, `gemma-4-31b`, and so on, and none of these match Ollama's `gemma4:*` tags. As a result, `ollama_name` was `null` for the whole family and a pulled `gemma4:*` model was not detected as installed.

This adds the four tags from https://ollama.com/library/gemma4/tags for the instruct models that are in the catalog:

| Catalog model | Ollama tag |
|---|---|
| `google/gemma-4-31B-it` | `gemma4:31b` |
| `google/gemma-4-26B-A4B-it` | `gemma4:26b` |
| `google/gemma-4-E4B-it` | `gemma4:e4b` |
| `google/gemma-4-E2B-it` | `gemma4:e2b` |

Base checkpoints (`google/gemma-4-31B`, `google/gemma-4-26B-A4B`) stay unmapped, like the Gemma 3 rows, because the Ollama tags ship the instruct weights.

What this doesn't cover from the issue:
- **`gemma4:12b`:** the catalog has no `google/gemma-4-12B-it` entry, only community fine-tunes, so there's nothing to map it to yet. That needs a catalog addition.
- **`llmfit info "gemma4:12b"`:** `info` doesn't resolve Ollama tags for any family (`llmfit info gemma3:12b` also prints `No model found matching`), so that is separate from this mapping gap.

### Testing
- New `test_ollama_mappings_gemma4`: checks the four mappings, checks that the base 31B stays unmapped, and checks that an installed set from `ollama pull gemma4:31b` marks `google/gemma-4-31B-it` installed but not `google/gemma-4-26B-A4B-it`. The test fails without the new rows.
- `llmfit --json info google/gemma-4-31B-it` now reports `"ollama_name": "gemma4:31b"`.
- `cargo test -p llmfit-core`: 743 passed. `cargo fmt` is clean.

https://claude.ai/code/session_01EcwjgZmx8zbtn6PpwDY9A1


AlexsJones (commented):
Thanks, this matches the four tags on ollama.com/library/gemma4, and the test covers both the instruct-only rule and the `gemma4:31b` installed check from #1024. Code is good to go. The formal approval follows once CI has run on this head (first-contribution run, so it needs the workflow click first).

Issue #635, opened 2026-06-15
MacBook Neo shows no models under "Good" or "Runnable" due to conservative KV cache filtering


goney3:
### Describe the Bug
On an 8GB Apple Silicon device (specifically the new MacBook Neo with a binned A18 Pro chip), the TUI correctly detects system memory (displays `5.9 GB Available / 8.0 GB total`), but the model list completely zeroes out when the "Fit" filter is set to `Good` or `Runnable`. 

Every single micro-model in the catalog (even 1.5B and 2B models like Qwen 3.5 and Gemma 3) is being pushed into the `Marginal` category. This seems to be caused by the scoring engine calculating memory requirements based on the model's maximum native context window (e.g., 32k tokens), which bloats the estimated KV cache footprint beyond the 5.9 GB threshold, even though these models run perfectly fine with smaller context sizes.

### Steps to Reproduce
1. Launch `llmfit` on an 8GB MacBook Neo.
2. Observe the top header line showing correct memory detection (~5.9 GB available).
3. Press `f` to cycle the Fit filter to `Runnable` or `Good`.
4. Observe that the TUI displays "No Models" matching the criteria.
5. Cycle the filter to `Marginal` or `All`, and notice that the ultra-compact models finally populate but with heavily penalized scores.

### Expected Behavior
Ultra-compact models (under 3B parameters with low quantization) should be categorized as `Runnable` or `Good` on an 8GB machine, perhaps with a warning or visual indicator that context length should be restricted, rather than hiding them from the runnable view entirely.

### Environment Details
- **Hardware:** MacBook Neo (A18 Pro, 5-core GPU variant)
- **RAM:** 8.0 GB total (5.9 GB reported available by `llmfit`)
- **OS:** macOS Tahoe 26.5.1

AlexsJones:
Thanks for the detailed report, I'm going to take a look at lunch

three-foxes-in-a-trenchcoat:
Okay, dug into this and I can't reproduce it. A few things don't line up.

First, the KV cache mechanism you're describing was already fixed back in #311. We don't size the cache at the model's full native window anymore. There's a hard cap at 8192 tokens for estimation (`DEFAULT_ESTIMATION_CTX` in `llmfit-core/src/fit.rs`), so a 32k or 128k native window doesn't blow up the footprint the way you'd expect. You can override it with `--max-context` if you want.

[...]

PR #1044, opened 2026-09-15
fix(hardware): recognize A-series Apple Silicon unified memory

lukatao:
On an 8 GB Apple A18 Pro Mac, `llmfit system` detects Metal but reports `VRAM unknown`, and JSON reports `unified_memory: false`. The Apple Silicon detector only matches M-series names, so the A18 Pro falls through to the non-unified Mac Metal path.

Recognize A-series and M-series chip names with a shared predicate used by both detection paths. The JSON parser directly creates Apple Silicon GPUs with unified memory backed by the detected system RAM pool. Run the text profiler only when JSON did not identify an Apple unified-memory GPU: a successful JSON result is preserved even if a text query would fail, and normal detection does not count the GPU twice. Also accept the `spdisplays_metal: spdisplays_supported` field variant alongside the existing Metal family-support field.

Related to #635 and the [maintainer's request for the chipset output](https://github.com/AlexsJones/llmfit/issues/635#issuecomment-4777370934). #649 addressed CPU-only fit scoring; this change addresses the separate hardware-detection gap. The original issue's complete filtering behavior has not been reproduced, so this PR does not automatically close it.

## Local verification

Built the unmodified `58e29bb` baseline and this branch, then ran both on the same Apple A18 Pro host (5-core GPU, 8 GB RAM, macOS 26.5.2), without hardware overrides:

```text
$ llmfit-before --no-dashboard system
Backend: Metal
GPU: Apple A18 Pro (VRAM unknown, Metal)

$ llmfit-after --no-dashboard system
Backend: Metal
GPU: Apple A18 Pro (unified memory, 5.33 GB GPU-available of 8.00 GB shared, Metal)
```

JSON confirms `gpu_vram_gb: null` → `8.0` and `unified_memory: false` → `true`. Both versions detect exactly one GPU. The 5.33 GB figure is the existing Metal working-set-limit reporting, now reached for A18 Pro.

## Validation

- Parser regression tests cover A18 unified-memory recognition and both Metal field schemas.
- Four macOS CLI integration tests use a PATH-isolated profiler stub to exercise the actual subprocess flow: successful A/M-series JSON with failing/non-UTF-8/unparseable text, skipping a redundant successful text probe, recovery from failed/malformed JSON through text, and no invented GPU when both probes fail. The JSON-preservation and redundant-probe tests failed before the review fix and pass afterward.
- Hardware tests: **131 passed**. Coverage includes A18 fixtures, existing M-series names, non-chip Apple names, both Metal schemas, unsupported Metal status, and existing Intel/AMD Mac fixtures.
- `cargo test --locked --offline -j 2`: **901 passed, 1 ignored** across the default workspace members.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Strict `cargo clippy --locked --offline -j 2 -p llmfit-core --all-targets -- -D warnings` fails on existing diagnostics in unchanged code (including dead code and documentation/style lints). None of its diagnostics point to changed lines.


AlexsJones (commented):
Thanks, this is a careful change. The shared `is_apple_silicon_gpu_name` predicate, JSON-first detection with the text probe only as a fallback (and never double-counted), and fixtures for both profiler schemas all read right, and the edge-fixture update (Apple M2 reported as one unified GPU at the system pool size) is the behaviour we want. Two notes, neither blocking: `llmfit-tui/tests/macos_gpu_detection.rs` only runs on the macOS runner, which is fine since that's where the profiler lives, and `assert_cmd` is already a dev-dependency there; and on M-series Macs the GPU name now comes from `sppci_model` instead of the CPU name, the same string in practice but worth knowing for JSON consumers. Formal approval once CI has run on this head.

lukatao:
> Thanks, this is a careful change. The shared `is_apple_silicon_gpu_name` predicate, JSON-first detection with the text probe only as a fallback (and never double-counted), and fixtures for both profiler schemas all read right, and the edge-fixture update (Apple M2 reported as one unified GPU at the system pool size) is the behaviour we want. Two notes, neither blocking: `llmfit-tui/tests/macos_gpu_detection.rs` only runs on the macOS runner, which is fine since that's where the profiler lives, and `assert_cmd` is already a dev-dependency there; and on M-series Macs the GPU name now comes from `sppci_model` instead of the CPU name, the same string in practice but worth knowing for JSON consumers. Formal approval once CI has run on this head.

Thanks for the review! Looks like CI is currently awaiting maintainer approval to run.

Issue #1039, opened 2026-09-14
Weekly model refresh fails since 2026-08-31: HF 429 rate limiting degrades the scrape and trips the #963 guard


Akciali:
@AlexsJones flagging this since it is easy to miss.

The `weekly-model-update` workflow fails at "Refresh model database" on its last three scheduled runs (#11 on 08-31, #12 on 09-07, #13 on 09-14: https://github.com/AlexsJones/llmfit/actions/runs/34798219787), each after about 4m30. Last green run: #10 on 08-24 (PR #945). No automated PR since, so it is easy to miss. The `update-benchmarks` job is fine.

Tail of run #13:

```
Merged with existing database (12937 models):
  Updated: 4716, Retained historical: 8221
  Architecture heads restored from prior catalog: 1759
ERROR: config.json fetch dropped attention-head metadata for 1759 models (limit 25). Prior values were kept, but refusing to ship this scrape so the mass drop cannot land quietly again.
```

The guard from #963 is doing its job. The cause is HuggingFace rate limiting: every discovered model in the log comes out as `0 downloads, ctx 4096`, which is what the scraper yields when `fetch_model_info` and `fetch_config_json` fail.

Reproduced outside CI (`--threads 8 --no-gguf-sources`, no token): `-n 200` on a fresh quota gives 26 HTTP 429 and exit 0; `-n 1000` right after, quota exhausted, gives 192 HTTP 429 out of 192 curated fetches, then `ERROR: config.json fetch dropped attention-head metadata for 76 models (limit 25)` and exit 1. Same signature.

HF states the policy in the 429 response: `ratelimit-policy: "fixed window";"api";q=500;w=300`, with `ratelimit: "api";r=<remaining>;t=<reset>`. The scraper makes two to three requests per model (info, `config.json`, and a second info fetch per discovered model since the listing omits `likes`), so `-n 5000 --threads 8` is on the order of 13,000 requests in 4m30, above any per-window quota even with `HF_TOKEN`. Nothing in the script handles 429 or reads `Retry-After`: `fetch_model_info` logs "skipping", `fetch_config_json` swallows every exception, so a burst silently zeroes downloads, resets context and drops architecture fields. The green runs after the 08-03 `-n 5000 --threads 8` change may have shipped degraded data the same way.

Not proposing a fix, this is your scraper and guard. Reporting it because it also gates the release-date backfill in #176. Repro logs available if useful.

AlexsJones:
[...]

Issue #1021, opened 2026-09-10
`llmfit update` fetches model lists successfully but caches 0 of 237 models


qhaz:
### Bug description

llmfit version: 1.1.6
OS: Linux (x86_64), Ubuntu 24.04
Install method: binary at ~/.local/bin/llmfit

Summary

llmfit update successfully fetches the trending/download-ranked model lists from HuggingFace (237 unique models total), but the subsequent per-model config.json processing step silently discards every single one. The cache file is written but ends up with an empty models array (34 bytes). This happens consistently across multiple runs, with a freshly-created HuggingFace token, and with --no-dashboard set.

Steps to reproduce

rm ~/.local/share/llmfit/hf_models_cache.json
llmfit update

Output

Fetching 100 trending models from HuggingFace...
Received 100 trending text-generation models
Received 25 trending image-text-to-text models
Received 25 trending any-to-any models
Received 25 trending text-to-speech models
Fetching 50 top-downloaded models...
Received 50 download-ranked text-generation models
Received 12 download-ranked image-text-to-text models
Received 12 download-ranked any-to-any models
Received 12 download-ranked text-to-speech models
Processing 237 unique models (fetching config.json for KV cache metadata)...
Saving 0 cached models (0 new)...

No new models found — cache is up to date.
Total cached: 0

What I've ruled out

Network/DNS/TLS reachability — confirmed working:
curl -sIL https://huggingface.co/Qwen/Qwen2.5-7B-Instruct/resolve/main/config.json | grep -i "^HTTP"
HTTP/2 307
HTTP/2 200
HuggingFace auth/token — tried both without a token and with a freshly-generated Read-scoped token (~/.huggingface/token). No change in behavior. A gated model (meta-llama/Llama-3.1-8B) does return 401 as expected, but this can't explain 237/237 failures since the vast majority of trending/top-downloaded models are not gated.
The --no-dashboard background server — strace -f -e trace=network during a normal update run showed a tight retry loop of connect() calls to 127.0.0.1:8787 returning ECONNREFUSED. Re-ran with --no-dashboard explicitly set: identical 0-cached result, so this loop is not the root cause (though it may be worth fixing separately as background noise).
Rust internal logging — RUST_LOG=debug llmfit update --no-dashboard produced byte-for-byte identical output to a normal run. No additional logging was emitted, so either llmfit doesn't use the standard Rust logging crate conventions, or update's fetch/cache step doesn't have any instrumented log statements.

[...]

PR #1048, opened 2026-09-16
fix(update): request every mapped field in the HuggingFace list expand[]

SulimanAbdulrazzaq:
## Problem

`llmfit update` fetches the trending and top-downloaded lists successfully and then caches none of them (#1021):

```
Processing 237 unique models (fetching config.json for KV cache metadata)...
Saving 0 cached models (0 new)...
```

The reporter ruled out network, token scope, and the dashboard server, and no per-model error is ever printed. The cause is in the list query itself, not in the per-model `config.json` step.

`hf_get_list_for_pipeline()` asked for:

```
https://huggingface.co/api/models?pipeline_tag=…&sort=…&limit=…&expand[]=gguf
```

`expand[]` switches the Hub's `/api/models` endpoint into **projection mode**: the response then carries `id`, `_id`, the sort key, and *only* the properties named in `expand[]`. Verified against the live API:

```console
$ curl -s 'https://huggingface.co/api/models?pipeline_tag=text-generation&sort=trendingScore&limit=2&expand%5B%5D=gguf' \
  | jq -c '.[0] | keys'
["_id","id","trendingScore"]

$ curl -s 'https://huggingface.co/api/models?pipeline_tag=text-generation&sort=trendingScore&limit=2' \
  | jq -c '.[0] | keys'
["_id","createdAt","downloads","id","library_name","likes","modelId","pipeline_tag","private","tags","trendingScore"]
```

So `pipeline_tag`, `tags`, `author`, `createdAt` and `safetensors` were all absent from every entry. `HfApiModel` declares them `#[serde(default)]`, so they deserialized as `None`/empty instead of failing, and the text-generation check at the top of `map_to_llm_model()` then returned `None` for all 237 entries — hence a fetch that looks healthy and a cache that stays empty. `gguf` was not returned either, so the metadata the `expand[]` was added for never arrived.

Introduced in ff28cdd ("fix: use GGUF metadata for parameter counts in GGUF-only repos", 2026-06-09), first shipped in v0.9.31.

## Change

- Add `HF_LIST_EXPAND`, naming every property the mapping reads (`pipeline_tag`, `tags`, `author`, `createdAt`, `safetensors`, `gguf`), and build the list URL from it. `license` is deliberately omitted — the API rejects it as an expand property (`✖ Invalid option`), and `map_to_llm_model()` already recovers it from the `license:` prefix in `tags`.
- Extract `hf_list_url()` and `is_accepted_pipeline()` so both are unit-testable without a network call. No behaviour change beyond the query.

No dependency, schema, or cache-format changes; `CACHE_VERSION` is untouched because the shape of a cached `LlmModel` is unchanged.

Verified the corrected query against the live API — `pipeline_tag`, `tags`, `author`, `createdAt` and `safetensors` all come back populated.

## Validation

Run in a clean Ubuntu 24.04 container on `stable` (rustc 1.98.1), on this branch:

```console
$ cargo fmt --all -- --check
(clean)

$ cargo test -p llmfit-core
test result: ok. 764 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
   hardware_profiles: ok. 3 passed; 0 failed
   onnx_models:       ok. 6 passed; 0 failed
   schema_validation: ok. 2 passed; 0 failed

$ cargo test -p llmfit-core update::tests::
test update::tests::test_hf_list_url_requests_every_expanded_field ... ok
test update::tests::test_hf_list_url_requests_the_classification_fields ... ok
test update::tests::test_expand_projection_without_pipeline_fields_is_rejected ... ok
test update::tests::test_full_list_entry_is_accepted_and_keeps_its_metadata ... ok
test update::tests::test_entry_tagged_without_a_pipeline_tag_is_accepted ... ok
test result: ok. 25 passed; 0 failed

$ cargo clippy -p llmfit-core --all-targets --all-features
(19 lib + 21 lib-test warnings, all pre-existing and none in update.rs's changed lines)
```

Scoped to `llmfit-core` rather than `--workspace`: the desktop crate needs the Tauri system dependencies, which this change does not touch. The five new tests are pure — the URL builder and the pipeline check both run without network.

`test_expand_projection_without_pipeline_fields_is_rejected` is the regression guard: it feeds the exact projection shape the Hub returns for `expand[]=gguf` alone and asserts the entry is rejected, so re-narrowing the `expand[]` list fails the suite instead of silently emptying the cache.

Not verified end to end: I could not run `llmfit update` itself against a live account here, so the "237 → N cached" number is not reproduced in this PR — only the API behaviour that causes it, and the classification path it breaks.

## Note

`scripts/scrape_hf_models.py` builds its list query the same way (`expand[]=safetensors&expand[]=config&expand[]=cardData`). I have not checked whether its downstream code depends on `pipeline_tag`/`tags`, and deliberately left it alone — it is a separate path and #1039/#1047 are already in flight there. Worth a look separately.

Refs #1021


AlexsJones (approved):
Great diagnosis, the projection-mode curl output makes the 237-fetched, 0-cached mystery obvious. `HF_LIST_EXPAND` now names every field `HfApiModel` reads (with `license` still recovered from tags), and `test_expand_projection_without_pipeline_fields_is_rejected` will catch anyone narrowing the list again. CI is green on all three platforms. Approving.

PR #1049, opened 2026-09-17
bench: community results for amd-radeon-rx-7900-xt

anshusangita08:

| --- | --- | --- | --- |
| hf.co/Fbjfhvgv/Qwen2.5-Coder-7B-Instruct-Uncensored-GGUF:Q8_0 | ollama | 81.0 | 47.4 |
| ornith-q5-80536:latest | ollama | 36.3 | 141.0 |
| GLM-4.7-Flash:latest | ollama | 35.2 | 308.3 |


AlexsJones (changes requested):
Thanks for sharing the RX 7900 XT runs! Two things before these can go in:

- The two `ornith-q5-80536:latest` files disagree by 9x on the same machine (4.04 tok/s with 877 ms TTFT in `1789629217-782c8d37.json`, 36.28 tok/s in `1789629878-85320498.json`). That looks like one run spilled to CPU or used a different quant or context. Which one reflects your normal setup? Please drop the other. Also, `ornith-q5-80536` is a local Ollama tag, so llmfit can't match it to a catalog model. If it's a build of a public model, re-running under that model's tag would make the result usable.
- Every file reports `gpuCount: 2` and `vramGb: 36.0` under the single name "AMD Radeon RX 7900 XT" (a 20 GB card). What's the second GPU, and did Ollama split these models across both? Community results are matched on CPU and GPU name only, so a two-GPU result would be read as a single 7900 XT number.

The Qwen2.5-Coder and GLM-4.7-Flash files look consistent otherwise.

anshusangita08:
The second card is RX 9060XT 16GB. It is not used here. AI Workloads only use 7900XT.

Dropped the ornith-q5-80536:latest - 4token it spilled as 2 tests started at the sam time causing issues. And it offloaded to CPU

AlexsJones:
Thanks for clearing up the second card. If the 9060 XT sits idle during these runs, they're fine as single 7900 XT results. The drop of the 4 tok/s `ornith-q5-80536` run hasn't reached the PR yet though: the branch is still at `6935061` and `1789629217-782c8d37.json` is still in it. Could you push that commit? Then this is good to go.

PR #1050, opened 2026-09-17
fix(bench): guard bogus TPS when Ollama eval_duration≈0

### LINKED ISSUE #1038: bench: Ollama runs with a near-zero eval_duration record 1,000,000 tok/s and flow into community data

Frank-zhu0404:
Fixes #1038

A run where Ollama reports `eval_count: 1` and `eval_duration: 1000` is computed as 1,000,000 tok/s. That value goes into `avgTps` and can later surface as a "measured" rate from shared or community results (as seen in #1022).

## Changes

- **`ollama_tps`** (extracted from `ollama_generate`): uses Ollama's native timing only when at least two tokens were evaluated and the rate is at most `MAX_PLAUSIBLE_TPS` (10,000 tok/s). Otherwise it falls back to the wall clock. A single token has no decode interval to time; near-zero `eval_duration` no longer divides into an impossible rate.
- **`BenchSummary::from_runs`**: leaves rates above the bound out of `avg/min/max_tps` so one degenerate run cannot dominate the average.
- **`LocalBenchIndex::load`** (`share.rs`) and **`community_results_for_specs`** (`benchmarks.rs`): use `bench::is_plausible_tps` instead of `> 0.0`, so already-saved or merged outliers are not served back.

## Testing

- `ollama_tps_ignores_degenerate_native_timing` — `eval_count: 1, eval_duration: 1000` falls back to wall clock (0.5 tok/s over 2s), not 1e6.
- `summary_excludes_implausible_tps` — a 1e6 outlier is excluded from summary stats.
- `cargo test -p llmfit-core --lib -- ollama_tps_ignores_degenerate_native_timing summary_excludes_implausible_tps test_summary_ share:: community_results community_embed` — all passed.

Note: #1042 targets the same issue with a similar approach; happy to close this if that lands first.


AlexsJones:
Thanks for digging into #1038, and for the clear write-up. #1042 landed yesterday with the same change: the `ollama_tps` helper with the two-token check, plus `is_plausible_tps` and the 10,000 tok/s bound used in `bench.rs`, `share.rs` and `benchmarks.rs`. So main already has everything this PR does. As you offered, it's fine to close this one. Appreciate you picking it up.

Frank-zhu0404:
Closing as duplicate — maintainer confirmed the same fix already landed on main via #1042. Thanks @AlexsJones.

PR #1051, opened 2026-09-17
fix(providers): find HF models in HF_HUB_CACHE and XDG_CACHE_HOME

SulimanAbdulrazzaq:
## Problem

`llmfit` pulls MLX models with `hf download`, which stores them wherever `huggingface_hub` resolves its cache. Checked against `hf env` (huggingface_hub 1.32.0), that order is `HF_HUB_CACHE`, the legacy `HUGGINGFACE_HUB_CACHE`, `$HF_HOME/hub`, `$XDG_CACHE_HOME/huggingface/hub`, then `~/.cache/huggingface/hub`.

`dirs_hf_cache_all()` only read `HF_HOME`, and it feeds both installed-model scans (`scan_hf_cache_for_mlx`, `scan_hf_cache_for_gguf`). So with a relocated hub cache, a model `llmfit` had just downloaded was never marked installed.

Planting `models--mlx-community--Qwen3-1.7B-4bit` in each cache and reading the `installed` column of `llmfit --no-dashboard --csv fit`, on `main` (Linux, isolated `HOME`):

```
default cache (~/.cache/huggingface/hub)             installed=true
HF_HOME=<dir>                                        installed=true
XDG_CACHE_HOME=<dir>                                 installed=true    (Linux: dirs::cache_dir() is already $XDG_CACHE_HOME)
HF_HUB_CACHE=<dir>                                   installed=false
HUGGINGFACE_HUB_CACHE=<dir>                          installed=false
HF_HOME=<a> and HF_HUB_CACHE=<b>                     installed=false
```

With the fix, all six report `installed=true`. The first row is a control: it shows the name matching itself works.

## Change

Resolve the scan directories in `huggingface_hub`'s order. The `~/.cache` and macOS `dirs::cache_dir()` fallbacks from #575 are unchanged, and `$XDG_CACHE_HOME` is listed once, since `dirs::cache_dir()` already is that path on Linux. Resolution moves into a pure `hf_cache_dirs_for()` so the tests can cover each variable without `set_var`, which is `unsafe` and not used in this crate.

On Linux the `XDG_CACHE_HOME` case was already covered through `dirs::cache_dir()`; the new candidate matters on macOS and Windows, where that function returns a different directory. A unit test pins each case, including the de-duplication.

## Validation

Linux, rustc 1.98.1:

- `cargo test -p llmfit-core` — 765 passed, 1 ignored
- six new `hf_cache_dirs` tests; with the two new lookups removed, three of them fail and the rest still pass
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p llmfit-core --all-targets --all-features` — unchanged, the same 25 pre-existing warnings as `main`, none in the changed code

`cargo test -p llmfit` has 2 failures on this branch **and on untouched `main`** in the same container: `plan_json_reports_disk_size_at_the_requested_quant`, `Some(126.4329543456)` vs `Some(126.43295434560001)`. CI is green on `main`, so that assertion looks environment-dependent; it is unrelated to this change.

Not run: `cargo test --workspace` (the desktop crate needs Tauri system dependencies), macOS and Windows runners, and a real model download.


AlexsJones (approved):
Nice work checking the lookup order against `hf env` and planting the model in each cache to prove it. `hf_cache_dirs_for` follows huggingface_hub's order (`HF_HUB_CACHE`, `HUGGINGFACE_HUB_CACHE`, `$HF_HOME/hub`, `$XDG_CACHE_HOME`) and keeps the #575 macOS and `~/.cache` fallbacks. The de-dup test covers Linux, where `dirs::cache_dir()` already is `$XDG_CACHE_HOME`. Keeping the resolver pure instead of reaching for `set_var` is a good call too. CI is green, approving.

